PR #2, opened 2017-03-10
Update Readme file, fix for lodash inclusion

jeff-phillips-18:


src/components/services-view/services-view.controller.ts
```diff
@@ -21,7 +21,7 @@ export class ServicesViewController implements angular.IController {
   }
 
   public $onInit() {
-    this.ctrl.origServices = this.ctrl.services.map(a => _.assign({}, a, {ctrl: this}));  // clone
+    this.ctrl.origServices = this.ctrl.services.map(a =>  _.assign(a, {ctrl: this}));
     this.ctrl.currentFilter = 'all';
```
dtaylor113: Is there a space before _.assign(...)?

jeff-phillips-18: There is. I can remove it.

PR #3, opened 2017-03-10
True 'mock' of DataServices.  Added Constants 'mock'

dtaylor113:
Beefing up web-catalog to be closer to a production app:
- moved mock data to separate file(s)
- added 'DataService' and 'Constants' to MockServices
- updated mock service data to be closer to data returned from service catalog API
- simulated async mock data loads
- added progress spinners
- refactored (moved) some files


dtaylor113:
Working on fixing unit tests....

app/config.local.js
```diff
@@ -29,3 +34,4 @@ kubernetes: "dev-mode"
 };
 
 window.MOCK_ORIGIN_SERVICES = false;
+window.MOCK_SERVICES_DATA = true;
```
jeff-phillips-18: So the idea of the mock services is to provide a service that mimics the services in origin-web-common. One flag for all to be in mock mode.
The mock data service should be the same as the DataService in origin-web-common and register itself as "DataService" in the 'mockServices' module. When the window.MOCK_ORIGIN_SERVICES flag is true, then the app uses the 'mockServices' module rather than the 'openshiftCommon' module and all the injection of services will be mock services.

Is the mockDataService ready to be used as a mock for the DataService in origin-web-common?

dtaylor113: I was thinking we'd need to mock out certain sub-sets of data.  Perhaps, 'catalog services' isn't quite ready, so we'd mock it, whereas 'brokers' can be retrieved from DataServices.  So not an all-or-nothing, but 'a la cart'.

dtaylor113: ...although, I'm wondering if variables like 'window.MOCK_SERVICES_DATA' should just be set in the DataManager class.  I don't think it needs to be user modifiable in config.local.js.

dtaylor113: Ok, understand how we want to mock things going forward, will work towards that.  Ok if we merge this and I follow up with another PR?

jeff-phillips-18:
The idea is to make it easy for UX designers to run the App without needing to run 'oc cluster up' or modifying code. This is a one stop shopping flag.

app/config.local.js
jeff-phillips-18: Do we need both flags? Are you looking to get some real data and some mock data in the same app? Is that necessary?

dtaylor113: Yes, I thought it necessary as I feel different parts of the API are going to be ready at different times.

dtaylor113:
I am currently working on mocking DataServices the way we want.  I'm ok with either killing this PR, keeping this PR open until the updates, or merging this PR then following up with another.

app/config.local.js
dtaylor113: Updated to a true 'mock' of DataServices

dtaylor113:
Added a true 'mock' of DataServices

Issue #8, opened 2017-03-14
config.local.js should not be tracked


spadgett:
In origin-web-console, we have `config.js` and `config.local.js`. `config.local.js` is added to `.gitignore` and is for changes specific to your env that you don't want to commit. `config.js` is used when `config.local.js` is not found.

https://github.com/openshift/origin-web-console/blob/master/Gruntfile.js#L525-L536

We should do the same here (or at least remove `config.local.js` and use `config.js`).

cc @jeff-phillips-18

dtaylor113:
Hi, this has been addressed with PR #9

PR #9, opened 2017-03-14
Removing config.local.js from repo.

dtaylor113:
Addresses #8 "config.local.js should not be tracked"


spadgett:
It looked like config.local.js is required in various places? We should only use it if it exists and fall back to config.js.

jeff-phillips-18:
This needs to come along with using config.js if config.local.js does not exist.

dtaylor113:
config.js is in the repo?  Not sure what I need to udpate? 'npm run build' works on my computer, not sure why Travis is failing.

spadgett:
The idea is that config.local.js is optional, but we're requiring it in several places:

https://github.com/openshift/origin-web-catalog/search?utf8=%E2%9C%93&q=config.local.js

dtaylor113:
ah, ok, working on it

dtaylor113:
Hi @spadgett & @jeff-phillips-18 , this addresses config.local.js .vs config.js.  It also fixes Travis Unit Test timeouts.

PR #18, opened 2017-03-15
Update webpack and URIjs usage to remove dependencies from bundles.

### LINKED ISSUE #16: Dist file is >1MB
The dist file should be much smaller. Are we pulling in things we shouldn't be?

https://github.com/openshift/origin-web-catalog/blob/master/dist/origin-web-catalogs.js

@jeff-phillips-18 @dtaylor113 @jwforres

jeff-phillips-18:
Fixes #16


bower.json
```diff
@@ -44,8 +44,7 @@
     "moment-timezone": "0.5.3",
     "ng-sortable": "1.3.4",
     "registry-image-widgets": "0.0.2",
-    "ui-select": "angular-ui-select#0.19.4",
-    "uri.js": "1.18.0"
+    "ui-select": "angular-ui-select#0.19.4"
   },
```
dtaylor113: There is still 'uri.js' in 'dependencies':

"dependencies": {
   ...
    "uri.js": "1.18.0"

PR #19, opened 2017-03-16
Services View: Added OpenShift Imagestreams

### LINKED ISSUE #24: We should use `Logger` instead of `console.log`
This lets us enable and disable logging using log levels.

https://github.com/openshift/origin-web-console#enable--disable-console-log-output
https://github.com/openshift/origin-web-common/blob/82ec00ff08179d87621d8dd003fdbc5130779567/src/services/logger.js

dtaylor113:
Major enhancement of the Services View component which now displays Catalog Services alongside of OpenShift Imagestreams.   

I still need to reconcile [catalog categories](https://github.com/openshift/origin-web-catalog/blob/master/src/constants.ts#L12) with [consoles categories](https://github.com/openshift/origin-web-console/blob/master/app/scripts/constants.js#L301).  Need to deep-dive with someone and determine what else needs to be added to the Catalog's categories.

@jwforres , @spadgett, @jeff-phillips-18 

Fixes #24


src/components/services-view/services-view.controller.ts
```diff
@@ -1,156 +1,173 @@
...
+        item.ctrl.serviceToOrder = item;
+        item.ctrl.openOrderingPanel();
+    };
+
+    public openOrderingPanel() {
+        this.ctrl.orderingPanelvisible = true;
+    };
+
+    public closeOrderingPanel = () => {
+        this.ctrl.orderingPanelvisible = false;
```
jwforres: nit but should be orderingPanelVisible with a capital V

jwforres: Why do you need to track if the ordering panel is visible?

src/components/services-view/services-view.controller.ts
```diff
@@ -1,156 +1,173 @@
...
+        let subCatsRows: any = [];
+        for (let i = 0, len = subCats.length; i < len; i += this.subCatMaxRowLength) {
+            subCatsRows.push(subCats.slice(i, i + this.subCatMaxRowLength));
         }
-      }
+        return subCatsRows;
+    }
+
+    private getRowOfSubCategory(subCategory: string) {
+        for (let row = 0; row < this.ctrl.subCategories.length; row += 1) {
```
jwforres: i know you said this code is changing, but a general comment is that we use _.each instead of for loops, the code is cleaner to read without all the array indexing

src/components/services-view/services-view.controller.ts
```diff
@@ -1,156 +1,173 @@
...
+                    return row;
+                }
+            }
+        }
+    }
+
+    private normalizeData(type: string, items: any) {
+        let retSvcs = [];
+        let objClass: any;
+        _.each(items, (item: any, key: string) => {
```
jwforres: if we aren't going to use an argument we generally don't define it, so in the case of an _.each we typically leave off key

src/components/services-view/services-view.controller.ts
```diff
@@ -1,156 +1,173 @@
...
+            if (type === 'service') {
+                objClass = new CatalogServiceItem(item);
+            } else if (type === 'image') {
+                objClass = new OpenshiftImageItem(item);
+            }
+            // each item/card needs a call back to 'this' class
+            _.assign(objClass, {ctrl: this});
```
jwforres: shouldn't give individual model objects access to the controller, why do the items need this?

app/pages/home/homePageController.ts
```diff
@@ -30,17 +30,22 @@ export class HomePageController {
...
         resource: 'serviceclasses'
       }, {}, (resources: any) => {
-        this.ctrl.services = resources.by("metadata.name"); ;
-        this.ctrl.applications = this.constants.REDHAT_APPLICATIONS;
-        this.ctrl.categories = this.constants.SERVICE_CATALOG_CATEGORIES;
-        this.ctrl.loading = false;
+        this.ctrl.catalogServices = resources.by("metadata.name");
       });
     } catch (e) {
       console.log("Error Loading serviceclasses from Service Catalog.k8s.io: " + e);
```
spadgett: +1, I had opened #24 for this as well

src/components/services-view/service-view.classes.ts
```diff
@@ -0,0 +1,181 @@
...
+
+    private getIcon() {
+        if (this.annotations) {
+            return 'font-icon ' + this.annotations.iconClass;
+        }
+    }
+
+    private getName() {
+        if (this.data.metadata.annotations) {
+            return this.data.metadata.annotations['openshift.io/display-name'];
```
spadgett: I noticed @jeff-phillips-18 added this filter in https://github.com/openshift/origin-web-common/pull/8

app/pages/home/homePageController.ts
```diff
@@ -30,17 +30,22 @@ export class HomePageController {
...
-      this.ctrl.services = [];
-      this.ctrl.applications = this.constants.REDHAT_APPLICATIONS;
-      this.ctrl.categories = this.constants.SERVICE_CATALOG_CATEGORIES;
-      this.ctrl.loading = false;
+      this.ctrl.catalogServices = {};
     }
+
+    try {
```
dtaylor113: Hi, yes, getting a lot of runtime errors when/if 'addtionalServers' isn't setup in config.js, or other settings in the pipeline are not configured correctly (etc/hosts).

src/components/services-view/service-view.classes.ts
```diff
@@ -0,0 +1,181 @@
...
+    public data: any;
+
+    constructor (catalogService: any) {
+        this.data = catalogService;
+        this.icon = this.getIcon();
+        this.name = this.getName();
+        this.description = this.getDescription();
+        this.longDescription = this.getLongDescription();
+        this.subCategory = this.getSubCategory();
+        this.category = this.getCategory();
```
dtaylor113: Ugh, that would require a change to a lot of code in the service view.  How would an item/object be configured to be in multiple categories?  Would it have multiple 'tags' which equal multiple sub-categories?  I was afraid I didn't fully understand the categorization process yet :-(

src/components/services-view/services-view.controller.ts
dtaylor113: This variable is used with the ng-if to hide/show the ordering panel.

src/components/services-view/services-view.controller.ts
dtaylor113: Yes, I'm 99% sure this is going to go away, but point taken about always using _.each ;-)

src/components/services-view/services-view.controller.ts
dtaylor113: This is somewhat of a hack.  pf-card-view has an anonymous callback when a card is clicked, unfortunately the callback function cannot access the functions/variables of the ServiceViewController class, thus we stamp the 'this' in there.  
At one point I had this working without using the click callback, might be worth re-implementing that way?

app/pages/home/homePageController.ts
jwforres: Gotcha, that is a temporary issue while we are prototyping, for the real MVP (and beyond) we wont be using additionalServers as everything will be proxied behind the main API server hostname, can you add TODO comments to take these try/catches out?

src/components/services-view/service-view.classes.ts
jwforres: yeah they can have many tags, and you might have tags that put something into more then one category.  Our existing catalog supports this.  As an example, I might have a template that is an example with a database and a php app, i might tag that both with "database" and "php" so that it shows up in both places.  The idea is to catch people from multiple points of entry since they might look for things in different ways.

jwforres: also it isnt just multiple subcategories, it could be underneath multiple top level categories as well

src/components/services-view/services-view.controller.ts
jwforres: I see, I'd rather pass references to functions than the entire controller.  How many functions is it needing access to?

src/components/services-view/services-view.controller.ts
jwforres: Hmm I assumed the overlay was being done as a modal (like uibModal), I twill take a look at the overlay code and get back to you on this one.

src/components/services-view/services-view.controller.ts
jeff-phillips-18: If you define:

    public handleClick = (item: any, e: any) => {

wouldn't that give it access to 'this'?

app/pages/home/homePageController.ts
spadgett: @jwforres Should we change DataService so that passing an unrecognized resource doesn't cause an exception?

src/components/order-service/order-service.controller.ts
```diff
@@ -12,10 +12,12 @@ export class OrderServiceController implements angular.IController {
   }
 
   public $onInit() {
-    this.ctrl.serviceIcon = this.ctrl.service.imageURL ? this.ctrl.service.imageURL : 'pf pficon-service';
-    this.ctrl.serviceName = this.ctrl.service.displayName ? this.ctrl.service.displayName : this.ctrl.service.metadata.name;
-    this.ctrl.versions = this.ctrl.service.versions;
-    this.ctrl.selectedVersion = this.ctrl.service.versions[0];
+    this.ctrl.serviceIcon = this.ctrl.service.icon ? this.ctrl.service.icon : 'pf pficon-service';
```
dtaylor113: Is imageUrl under 'metadata' or 'osbMetadata'?   I can't find 'osbMetadata' definition in https://github.com/kubernetes-incubator/service-catalog/blob/master/pkg/apis/servicecatalog/v1alpha1/types.go#L133.  -thanks!

src/components/services-view/services-view.controller.ts
dtaylor113: I had thought I tried that in the early stages, but I will try it again.  -thanks!

src/components/order-service/order-service.controller.ts
jwforres: osbMetadata, its a raw extension blob which means that data is just getting passed back from the underlying info the broker sent to the service catalog API. You wont see the fields called out in the svc cat schema because of that.  There are no guarantees that any of that metadata will be there, its all just done by convention.

app/mockServices/mockData/services.ts
```diff
@@ -1,16 +1,15 @@
...
-      "name": "Wildfly",
-      category: 'languages',
-      subCategory: 'java',
+      "name": "test-serviceclass-java",
       "uid": "1",
     },
-    displayName: 'WildFly',
-    imageURL: 'font-icon icon-wildfly',
+    'spec': {
+      'tags': ['java']
```
dtaylor113: Fixed

app/mockServices/mockData/services.ts
```diff
@@ -1,16 +1,15 @@
...
+    },
+    displayName: 'Test ServiceClass Java',
```
dtaylor113: Fixed

app/pages/home/home-page.html
```diff
@@ -6,7 +6,7 @@
     </div>
   </landingheader>
   <landingbody>
-    <services-view services="$ctrl.services" categories="$ctrl.categories"></services-view>
+    <services-view catalog-services="$ctrl.catalogServices" openshift-image-streams="$ctrl.openshiftImageStreams"></services-view>
```
dtaylor113: Fixed

app/pages/home/homePageController.ts
```diff
@@ -14,7 +14,7 @@ export class HomePageController {
     this.$q = $q;
     this.ctrl.loading = true;
     this.ctrl.applications = [];
-    this.ctrl.services = [];
+    this.ctrl.catalogServices = [];
```
dtaylor113: Fixed

app/pages/home/homePageController.ts
dtaylor113: Fixed

src/components/services-view/service-view.classes.ts
dtaylor113: Ok, understood, this is going to take a little longer to address

src/components/services-view/service-view.classes.ts
```diff
@@ -0,0 +1,181 @@
...
+    private getAnnotations() {
+        if (this.data.spec && this.data.spec.tags && this.data.spec.tags.length > 1 &&
+            this.data.spec.tags[0].annotations) {
+            return this.data.spec.tags[0].annotations;
+        }
+    }
+
+    private getIcon() {
+        if (this.annotations) {
+            return 'font-icon ' + this.annotations.iconClass;
```
dtaylor113: Waiting on openshift/origin-web-common#8

src/components/services-view/service-view.classes.ts
dtaylor113: Waiting on openshift/origin-web-common#8

dtaylor113:
@jeff-phillips-18 , do you have any idea why Travis is failing with:
    modified:   dist/vendor-bundle.js
    The command "hack/verify-dist.sh" exited with 1.
I noticed #21 failed with the same error.  I've run 'hack/install-deps' and 'verify-dist' locally and they run fine.

src/components/services-view/service-view.classes.ts
dtaylor113: Hi, fixed ServiceClasses to be under multiple categories and sub-categories.  Waiting on Jeff's PR in order to use relevant filters for the builder images.

app/pages/home/homePageController.ts
dtaylor113: Added the TODO comments

src/components/services-view/services-view.controller.ts

src/components/order-service/order-service.controller.ts
dtaylor113: Added 'imageUrl'

src/components/services-view/service-view.classes.ts
dtaylor113: Waiting on openshift/origin-web-common#8

dtaylor113: fixed

src/components/services-view/service-view.classes.ts
dtaylor113: fixed

dtaylor113:
Hi @jwforres, I believe all PR issues have been addressed

@spadgett, just a heads up that I changed webpack-dev.js's 'dev-tool' back to 'eval-source-maps' because 'inline-source-maps' (for some reason we are still investigating) caused webpack to pause for a really long time at '92% optimization chunks'.

src/components/services-view/services-view.controller.ts
```diff
@@ -1,156 +1,190 @@
...
+            this.UpdateImageStreams();
+        }
+    }
+
+    public filterByCategory(category: string, subCategory: string, updateSubCategories: boolean) {
+        if (category === 'all' && subCategory === 'all') {
+            this.ctrl.filteredItems = this.ctrl.allItems;
         } else {
-          return  service.metadata.category === category && service.metadata.subCategory === subCategory;
+            this.ctrl.filteredItems = this.$filter('filter')(this.ctrl.allItems, (item: any) => {
```
jwforres: we prefer to use lodash's _.filter from the JS instead of $filter('filter')

src/components/services-view/services-view.controller.ts
```diff
@@ -1,156 +1,190 @@
...
+            this.ctrl.filteredItems = this.ctrl.allItems;
+        }
+    }
+
+    private updateServiceClasses() {
+        this.ctrl.allItems = this.ctrl.allItems.concat(this.normalizeData('service', this.ctrl.serviceClasses));
+        this.updateState();
+    }
+
+    private UpdateImageStreams() {
```
jwforres: use consistent casing, you have update above and Update on this one

src/components/services-view/services-view.controller.ts
```diff
@@ -1,156 +1,190 @@
...
+            } else if (type === 'image') {
+                objClass = this.catalog.getImageItem(item);
+            }
+            retSvcs.push(objClass);
+        });
+        return retSvcs;
+    }
+
+    private hasCategory(item: any, category: string) {
+        return _.some(item.catsBySubCats, (itemCategory: string) => {
```
jwforres:
```
return _.includes(item.catsBySubCats, category);
```

src/components/services-view/services-view.controller.ts
```diff
@@ -1,156 +1,190 @@
...
+            return itemCategory === category;
+        });
+    }
+
+    private hasSubCategory(item: any, subCategory: string) {
+        return _.some(Object.keys(item.catsBySubCats), (itemSubCategory: string) => {
```
jwforres: same as above, use _.includes

app/mockServices/mockData/services.ts
```diff
@@ -1,16 +1,16 @@
 export const servicesData = {
-  "wildfly": {
+  "test-serviceclass-java": {
```
dtaylor113: Arrgg...my apologies!  I meant to remove 'service-view.classes.ts'!  It has been replaced with 'src/services/catalog.service.ts'

src/components/services-view/service-view.classes.ts
```diff
@@ -0,0 +1,193 @@
...
+    constructor (image: any, $filter: any) {
+        this.data = image;
+        this.annotations = this.getAnnotations();
+        this.iconClass = this.getIcon();
+        this.name = this.getName();
+        this.description = this.getDescription();
+        this.longDescription = this.getLongDescription();
+        this.catsBySubCats = this.getCategoriesBySubCategories();
+        this.$filter = $filter;
+        console.log("construct $filter = " + JSON.stringify($filter, null, ' '));
```
dtaylor113: Arrgg...my apologies!  I meant to remove 'service-view.classes.ts'!  It has been replaced with 'src/services/catalog.service.ts'

src/components/order-service/order-service.controller.ts
jwforres: I see where this was fixed in the models and for the tiles, but in the order panel it looks like its still just using the class-based approach and not handling both ways.

src/services/catalog.service.ts
```diff
@@ -0,0 +1,198 @@
...
+
+  public getServiceItem(resource: any) {
+    return new ServiceItem(resource, this);
+  }
+
+  public getImageItem(resource: any) {
+    return new ImageItem(resource, this);
+  }
+
+  public getCategoryFromSubCategory(subCategory: string) {
```
jwforres: I had several comments about these methods against the other file, i think they all still apply here if you can just look back at those in the history.

src/services/catalog.service.ts
```diff
@@ -0,0 +1,198 @@
...
+      return this.resource.metadata.imageUrl;
+    }
+  }
+
+  private getName() {
+    if (this.resource.osbMetadata) {
+      return this.resource.osbMetadata.displayName ? this.resource.osbMetadata.displayName : this.resource.metadata.name;
+    } else {
+      return this.resource.metadata.name;
+    }
```
jwforres: this whole if/else can be simplified down to

```
return _.get(this.resource, 'osbMetadata.displayName', this.resource.metadata.name);
```

src/services/catalog.service.ts
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  private getDescription() {
+    if (this.resource.osbMetadata) {
+      return this.resource.osbMetadata.description;
+    }
```
jwforres: remove the if:
```
return _.get(this.resource, 'osbMetadata.description');
```

src/services/catalog.service.ts
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  private getLongDescription() {
+    if (this.resource.osbMetadata) {
+      return this.resource.osbMetadata.longDescription;
```
jwforres: same thing, _.get

src/services/catalog.service.ts
```diff
@@ -0,0 +1,198 @@
...
+    }
+  }
+
+  private getCategoriesBySubCategories() {
+    let catsBySubCats = {};
+
+    if (this.resource.osbTags) {
```
jwforres: the if isn't necessary, _.each already handles existence checks on what you pass to it so if this.resource.osbTags is null or undefined it won't care it just treats it like there are no items

src/services/catalog.service.ts
```diff
@@ -0,0 +1,198 @@
...
+  public catsBySubCats: any;
+  public description: string;
+  public longDescription: string;
+  public resource: any;
+  private catalogSrv: CatalogService;
+
+  constructor (serviceClass: any, catalogSrv: CatalogService) {
+    this.resource = serviceClass;
+    this.catalogSrv = catalogSrv;
+    this.imageUrl = this.getImage();
```
jwforres: a service class might have either an imageUrl or an iconClass, our own brokers we write specifically for openshift will use iconClass, although there was some discussion about prefixing icon class with something ('something.openshift.io/iconClass'), @bparees did you guys decide what this will be?

src/services/catalog.service.ts
```diff
@@ -0,0 +1,198 @@
...
+    this.catalogSrv = catalogSrv;
+    this.iconClass = this.getIcon();
+    this.name = this.getName();
+    this.description = this.getDescription();
+    this.longDescription = this.getLongDescription();
+    this.catsBySubCats = this.getCategoriesBySubCategories();
+  }
+
+  private getIcon() {
+    return 'font-icon ' + this.catalogSrv.$filter('imageStreamTagIconClass')(this.resource);
```
jwforres: repeating this comment that i put on the old file:
so i think i said before tagName was optional but I just realized that when you don't include it then it is assuming the tag is "latest" which we shouldn't really make the assumption that a latest tag exists :-\

so i think you need to actually grab the name of a valid tag off this imagestream, i think you can just do what you do above in annotations and just use the first tag you find in the list to get the name of a valid tag

src/services/catalog.service.ts
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  private getName() {
+    return this.catalogSrv.$filter('annotation')(this.resource, "displayName");
```
jwforres: should fallback to this.resource.metadata.name when displayName doesn't exist

src/services/catalog.service.ts
```diff
@@ -0,0 +1,198 @@
...
+
+  private getLongDescription() {
+    return null;
+  }
+
+  private getCategoriesBySubCategories() {
+    let catsBySubCats = {};
+    let tags =  this.catalogSrv.$filter('imageStreamTagTags')(this.resource);
+
+    this.isBuilder = _.some(tags, (tag) => {
```
jwforres: repeating from the other file:
its weird that a "getter" is also modifying the object. I think instead I would make "tags" a private instance variable and set it to an array that you generate once by splitting the tags string. Then you can have a simple line after that in the constructor for this.isBuilder = _.includes(this.tags, "builder") and then have this method use the already split tags to figure out the categories.

src/services/catalog.service.ts
bparees: i'm not sure we picked for certain, but i think it'll be `template.openshift.io/iconClass`.  @jim-minter can confirm/deny my recollection.

jwforres:
I wonder if that's the right prefix if we want other brokers we write to
use the same annotation.  Maybe we should do something like
console.openshift.io since it's really a convention for the console?

On Mar 22, 2017 8:05 PM, "Ben Parees" <notifications@github.com> wrote:

*@bparees* commented on this pull request.
------------------------------

In src/services/catalog.service.ts
<https://github.com/openshift/origin-web-catalog/pull/19#discussion_r107562517>
:

> +}
+
+export class ServiceItem implements IServiceItem {
+  public iconClass: string;
+  public imageUrl: string;
+  public name: string;
+  public catsBySubCats: any;
+  public description: string;
+  public longDescription: string;
+  public resource: any;
+  private catalogSrv: CatalogService;
+
+  constructor (serviceClass: any, catalogSrv: CatalogService) {
+    this.resource = serviceClass;
+    this.catalogSrv = catalogSrv;
+    this.imageUrl = this.getImage();

i'm not sure we picked for certain, but i think it'll be
template.openshift.io/iconClass. @jim-minter <https://github.com/jim-minter>
can confirm/deny my recollection.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/openshift/origin-web-catalog/pull/19#discussion_r107562517>,
or mute the thread

src/services/catalog.service.ts
bparees: > I wonder if that's the right prefix if we want other brokers we write to
use the same annotation.  Maybe we should do something like
console.openshift.io since it's really a convention for the console?

makes sense.  i could also see broker.openshift.io/iconClass, though it is a bit redundant since anything providing that value is, by definition, a broker.

src/components/services-view/service-view.classes.ts
```diff
@@ -0,0 +1,193 @@
...
+    private getCategoriesBySubCategories() {
+        let catsBySubCats = {};
+        if (this.annotations) {
+            let tags = this.annotations.tags.split(/\s*,\s*/);
+            this.isBuilder = _.some(tags, (tag) => {
+                return tag === 'builder';
+            });
+            let lastTag = tags.pop();
+            let subCat = getSubCategoryFromAlias(lastTag);
+            let cat = getCategoryFromSubCategory(subCat);
```
dtaylor113: Hi, fixed this in ''src/services/catalog.service.ts'

src/components/services-view/service-view.classes.ts
```diff
@@ -0,0 +1,193 @@
...
+    }
+}
+
+var categories = allCategories;
+
+function getCategoryFromSubCategory(subCategory: string) {
+    let retCategory: string = 'other';
+
+    let breakLoops: boolean = false;
+    _.each(categories, (category) => {
```
dtaylor113: Fixed this in ''src/services/catalog.service.ts', although I got a runtime error when I tried to use _.findWhere, but it should exist, we are using lodash 3.10.1?  I used _.findIndex instead.

src/components/services-view/service-view.classes.ts
```diff
@@ -0,0 +1,193 @@
...
+    });
+
+    return retCategory;
+}
+
+function getSubCategoryFromAlias(subCategory: string) {
+    let retSubCategory: string = subCategory;
+
+    let breakLoops: boolean = false;
+    _.each(categories, (category) => {
```
dtaylor113: Fixed this in ''src/services/catalog.service.ts'

src/components/services-view/service-view.classes.ts
```diff
@@ -0,0 +1,193 @@
...
+        if (breakLoops) {
+            return false;
+        }
+        _.each(category.subCategories, (cat_subCategory) => {
+            if (cat_subCategory.categoryAliases) {
```
dtaylor113: Fixed this in ''src/services/catalog.service.ts'

src/components/services-view/service-view.classes.ts
```diff
@@ -0,0 +1,193 @@
...
+                let result = _.findIndex(cat_subCategory.categoryAliases, (alias) => {
+                    return alias === subCategory;
+                });
```
dtaylor113: Fixed this in ''src/services/catalog.service.ts'

src/services/catalog.service.ts
jim-minter: the commit had `template.openshift.io/iconClass` but I agree it makes sense to change it.  Let me know what you choose (`console.openshift.io/iconClass` sounds pretty sensible to me) and I can do a fix-up on the template service broker side if you want me to.

jwforres:
I think console.openshift.io is the right prefix, and then the annotation
makes sense even outside the service catalog use case.

src/services/catalog.service.ts
bparees: console.openshift.io/iconClass works for me. let's do it.

app/mockServices/mockData/services.ts
```diff
@@ -1,16 +1,16 @@
...
     "metadata": {
-      "name": "Wildfly",
-      category: 'languages',
-      subCategory: 'java',
+      "name": "test-serviceclass-java-nodejs",
       "uid": "1",
+      "imageUrl": "http://www.rightstar.com/wp-content/uploads/2014/03/Java.png"
```
dtaylor113: Added issue #28 , I will remove the urls

src/services/catalog.service.ts
dtaylor113: Hi, these methods were removed and combined into 'getCategoriesBySubCategories'.

src/services/catalog.service.ts

src/services/catalog.service.ts
spadgett: @dtaylor113 If you use the `displayName` filter instead of `annotation`, it will do that for you.

```js
return this.catalogSrv.$filter('displayName')(this.resource);
```

app/pages/home/homePageController.ts
```diff
@@ -25,22 +28,30 @@ export class HomePageController {
   };
 
   public update() {
+
+    // TODO: Remove these try/catches once we are not relying on 'additionalServers' in config.js
```
spadgett: @jwforres I think unrecognized resource should be a promise failure instead of runtime error. Then if the service catalog is down or not installed, it can use our existing error handling.

app/pages/services/servicesPageController.ts
```diff
@@ -25,21 +28,28 @@ export class ServicesPageController {
...
     try {
       this.dataService.list({
         group: 'servicecatalog.k8s.io',
         resource: 'serviceclasses'
       }, {}, (resources: any) => {
-        this.ctrl.services = resources.by("metadata.name");
-        this.ctrl.categories = this.constants.SERVICE_CATALOG_CATEGORIES;
-        this.ctrl.loading = false;
+        this.ctrl.serviceClasses = resources.by("metadata.name");
+      });
```
spadgett: @dtaylor113 Your calls to `dataService.list` should use the promise returned rather than passing in the callback as a parameter.

src/components/services-view/services-view.component.ts
```diff
@@ -2,8 +2,8 @@ import {ServicesViewController} from './services-view.controller';
 
 export const servicesView: angular.IComponentOptions = {
   bindings: {
-    services: '<',
-    categories: '<'
+    serviceClasses: '<',
+    imageStreams: "<"
```
spadgett: Nit: let's consistently use either single quote or double quote styles in the same declaration

src/components/services-view/services-view.controller.ts
```diff
@@ -2,155 +2,188 @@ import * as angular from 'angular';
...
-      });
+
+        if (updateSubCategories) {
+            this.ctrl.subCategories = this.getSubCategories(category);
+        }
+
+        this.ctrl.expandSubCatRow = this.getRowOfSubCategory(subCategory);
+
+        this.ctrl.currentFilter = category;
+        this.ctrl.currentSubFilter = (subCategory !== undefined) ? subCategory : 'all';
```
spadgett: Any reason not to use...?

```js
this.ctrl.currentSubFilter = subCategory || 'all';
```

src/components/services-view/services-view.controller.ts
```diff
@@ -2,155 +2,188 @@ import * as angular from 'angular';
...
-      this.ctrl.filteredServices = this.ctrl.origServices;
-    } else {
-      this.ctrl.filteredServices = this.$filter('filter')(this.ctrl.origServices, (service: any) => {
-        if (category !== 'all' && subCategory === 'all') {
-          return service.metadata.category === category;
-        } else if (category === 'all' && subCategory !== 'all') {
-          return service.metadata.subCategory === subCategory;
+
+    public $onChanges(onChangesObj: angular.IOnChangesObject) {
+        if (onChangesObj.serviceClasses && !onChangesObj.serviceClasses.isFirstChange()) {
```
spadgett: If the service classes or image streams are loaded before the directive is used, will that mean that the directive is never initialized?

src/components/services-view/services-view.controller.ts
```diff
@@ -2,155 +2,188 @@ import * as angular from 'angular';
...
+        }
+
+        if (onChangesObj.imageStreams && !onChangesObj.imageStreams.isFirstChange()) {
+            this.ctrl.imageStreams = onChangesObj.imageStreams.currentValue;
+            this.imageStreamsLoaded = true;
+            this.UpdateImageStreams();
+        }
+    }
+
+    public filterByCategory(category: string, subCategory: string, updateSubCategories: boolean) {
```
spadgett: The approach we took with the previous catalog is to create maps of category and subcategories with the items once on initialization. Then we don't have to filter the lists every time the user clicks a tile.

src/components/services-view/services-view.html
```diff
@@ -35,10 +35,11 @@
         </div>
         <div ng-if="$ctrl.expandSubCatRow === $index"
              class="sub-category-row"
-             pf-card-view config="$ctrl.cardViewConfig" items="$ctrl.filteredServices">
+             pf-card-view config="$ctrl.cardViewConfig" items="$ctrl.filteredItems">
           <div class="card-container">
-            <div class="card-icon {{item.imageURL ? item.imageURL : 'pf pficon-service'}}"></div>
-            <div class="card-name">{{item.displayName ? item.displayName : item.metadata.name}}</div>
+            <div ng-if="!item.imageUrl" class="card-icon {{item.iconClass ? item.iconClass : 'pf pficon-service'}}"></div>
```
spadgett: Class `pf`? Should that be `pficon`? Not sure if `pficon-service` is really the default we want.

src/services/catalog.service.ts
```diff
@@ -0,0 +1,166 @@
...
+  public getImageItem(resource: any) {
+    return new ImageItem(resource, this);
+  }
+
+  public getCategoriesBySubCategories(tags: any) {
+    let catsBySubCats = {};
+    _.each(tags, (tag) => {
+      _.each(this.categories, (category) => {
+        let subCat: any = _.find(category.subCategories, (subCategory: any) => {
+              if (subCategory.id === tag) {
```
spadgett: indentation

src/services/catalog.service.ts
```diff
@@ -0,0 +1,166 @@
...
+                return true;
+              } else if (subCategory.categoryAliases) {
+                return _.includes(subCategory.categoryAliases, tag);
+              }
+            }
```
spadgett: I prefer the shorter

```js
  return subCategory.id === tag || _.includes(subCategory.categoryAliases, tag);
```

Lodash should handle undefined category aliases

src/services/catalog.service.ts
```diff
@@ -0,0 +1,166 @@
...
+        );
+        if (!_.isUndefined(subCat)) {
```
spadgett: Any reason not to use?

```js
if (subCat)
```

src/components/services-view/services-view.controller.ts
dtaylor113: Hi, please see UpdateState() which tests for empty or if the particular resources has been 'loaded'.  This is to handle when the directive is initialized with empty serviceClasses/imageStreams and is waiting on the DataService to return asynchronously.

src/components/services-view/services-view.html
dtaylor113: This has been removed in my local branch, not committed yet.

src/components/services-view/services-view.controller.ts
spadgett: It looks like `this.ctrl.loading` will never be set to true in `updateState()` if image streams or service classes are empty?

app/pages/home/homePageController.ts
jwforres: @spadgett opened an issue in common to track it, watch is a problem still https://github.com/openshift/origin-web-common/issues/17

app/pages/home/homePageController.ts
```diff
@@ -25,22 +28,26 @@ export class HomePageController {
...
-      });
-    } catch (e) {
-      console.log("Error Loading serviceclasses from Service Catalog.k8s.io: " + e);
-      this.ctrl.services = [];
-      this.ctrl.applications = this.constants.REDHAT_APPLICATIONS;
-      this.ctrl.categories = this.constants.SERVICE_CATALOG_CATEGORIES;
-      this.ctrl.loading = false;
-    }
+
+    // TODO: Remove these try/catches once we are not relying on 'additionalServers' in config.js
```
jwforres: since you switched to the promises can you remove this comment, since handling the promise rejection is a good thing to keep in case we ever get errors from the API or the server goes down

app/pages/services/servicesPageController.ts
jwforres: yeah agreed switch this to use the promise like you did in homePageController

src/components/services-view/services-view.controller.ts
dtaylor113: I believe the only way this.ctrl.loading will never be set to false (what we need to stop the spinners) if images streams or service classes are initially set to empty, and never updated again asynchronously.  After DataService.list sets those collections asynchronously (to empty obj. or otherwise) the flags (serviceClassesLoaded & imageStreamsLoaded) will be set to true and this.ctrl.loading will be set to false, thus stopping the spinners.

src/components/services-view/services-view.controller.ts

app/pages/services/servicesPageController.ts
```diff
@@ -25,21 +28,28 @@ export class ServicesPageController {
...
+    } catch (e) {
+      this.logger.log("Error Loading serviceclasses from Service Catalog.k8s.io: " + e);
```
dtaylor113: Fixed

src/components/services-view/service-view.classes.ts
```diff
@@ -0,0 +1,193 @@
...
+    private getAnnotations() {
+        if (this.data.spec && this.data.spec.tags && this.data.spec.tags.length > 1 &&
+            this.data.spec.tags[0].annotations) {
+            return this.data.spec.tags[0].annotations;
+        }
+    }
+
+    private getIcon() {
+        console.log("$filter = " + JSON.stringify(this.$filter, null, ' '));
+        let icon = this.$filter('imageStreamTagIconClass')(this.data, "icon");
```
dtaylor113: Fixed

src/components/services-view/service-view.classes.ts
```diff
@@ -0,0 +1,193 @@
...
+        if (this.annotations) {
+            return this.annotations.description;
+        }
+    }
+
+    private getCategoriesBySubCategories() {
+        let catsBySubCats = {};
+        if (this.annotations) {
+            let tags = this.annotations.tags.split(/\s*,\s*/);
+            this.isBuilder = _.some(tags, (tag) => {
```
dtaylor113: Fixed

src/components/services-view/service-view.classes.ts
```diff
@@ -0,0 +1,193 @@
...
+                return tag === 'builder';
+            });
+            let lastTag = tags.pop();
```
dtaylor113: Fixed

src/components/services-view/services-view.controller.ts

src/components/services-view/services-view.controller.ts

src/components/services-view/services-view.controller.ts

src/services/catalog.service.ts

src/services/catalog.service.ts

src/services/catalog.service.ts

src/services/catalog.service.ts

src/services/catalog.service.ts
```diff
@@ -0,0 +1,152 @@
...
+
+  public getImageItem(resource: any) {
+    return new ImageItem(resource, this);
+  }
+
+  public getCategoriesBySubCategories(tags: any) {
+    let catsBySubCats = {};
+    _.each(tags, (tag) => {
+      _.each(this.categories, (category) => {
+        let subCat: any = _.find(category.subCategories, (subCategory: any) => {
```
jwforres: you know what, just realized if this matches multiple subcategories only the first one will actually get matched, i dont think we can break out early at all here

src/services/catalog.service.ts

src/services/catalog.service.ts
dtaylor113: Fixed in next commit.

app/pages/services/servicesPageController.ts
dtaylor113: Fixed

app/pages/home/homePageController.ts
dtaylor113: Fixed

src/services/catalog.service.ts
```diff
@@ -0,0 +1,152 @@
...
+    this.longDescription = this.getLongDescription();
+    this.catsBySubCats = this.getCategoriesBySubCategories();
+  }
+
+  private getImage() {
+    return _.get(this.resource, 'metadata.imageUrl', '');
+  }
+
+  private getIcon() {
+    return 'fa fa-cubes';
```
jwforres: so based on the discussion with API folks, i think this becomes
```
return _.get(this.resource, ['osbMetadata', 'console.openshift.io/iconClass'], 'fa fa-cubes');
```
And we can debate later if there is a better default icon for service class items vs what we use for builder images.

src/components/services-view/services-view.component.ts
dtaylor113: Fixed

src/services/catalog.service.ts
```diff
@@ -0,0 +1,152 @@
...
+    this.imageUrl = this.getImage();
+    this.iconClass = this.getIcon();
+    this.name = this.getName();
+    this.description = this.getDescription();
+    this.longDescription = this.getLongDescription();
+    this.catsBySubCats = this.getCategoriesBySubCategories();
+  }
+
+  private getImage() {
+    return _.get(this.resource, 'metadata.imageUrl', '');
```
jwforres: osbMetadata not metadata

src/components/services-view/services-view.controller.ts
dtaylor113: I'll review that code and follow up with another PR.

src/components/services-view/services-view.controller.ts

src/services/catalog.service.ts

src/services/catalog.service.ts

dtaylor113:
Hi @jwforres ,
So for service class icons, is 'console.openshift.io/IconClass' under 'osbMetadata'?
Thanks

src/services/catalog.service.ts
dtaylor113: I don't follow.  A single tag like 'nodejs' is going to match a single sub-category (or alias).  A service item may have multiple tags, which may match to multiple categories and sub-categories.   We don't have a mapping of matching a single tag to multiple sub-categories?

app/pages/home/homePageController.ts
dtaylor113: Fixed

src/services/catalog.service.ts

src/services/catalog.service.ts
dtaylor113: Ok, thanks, fixed

src/services/catalog.service.ts
dtaylor113: This may come back to my initial comment: "I still need to reconcile [catalog categories](https://github.com/openshift/origin-web-catalog/blob/master/src/constants.ts#L12) with [consoles categories](https://github.com/openshift/origin-web-console/blob/master/app/scripts/constants.js#L301).  Need to deep-dive with someone and determine what else needs to be added to the Catalog's categories."

I see in  [consoles categories](https://github.com/openshift/origin-web-console/blob/master/app/scripts/constants.js#L301) there is something called 'subcategories', which really makes that structure 3 levels deep:  Categories -> Items -> subcategories.  I am currently using a categories constant which is two levels deep.

dtaylor113: Either way, it would be great if we could handle this in a follow up PR.

dtaylor113:
Hi @jwforres & @spadgett,  I believe I've addressed all issues (except the last one @jwforres raised, but hoping to push that to a follow up PR).
Thanks,
- Dave

jwforres:
Oh an please squash the commits :)

dtaylor113:
Created issue #29 so we can follow up on the subcategory matching logic

PR #20, opened 2017-03-17
Change dev server port to 9001 to avoid conflict with console

jeff-phillips-18:


spadgett:
We should change the URL in config.js oauth redirect too

spadgett:
Is it worth making this a parameter so I can choose the port?

jeff-phillips-18:
We could but it would need to be in config.local.js as well. So, you can change it in config.local.js and then use 'npm run start --port xxx' ?

spadgett:
Yeah I think it's reasonable to ask people to update config.local.js if they want to run on the non-default port.

README.md
```diff
@@ -11,13 +11,29 @@ A component library created for the cataloging components for [The OpenShift Web
...
+This will watch for asset changes.
+
+You may need to patch oc to allow for port 9001:
+```
+$ oc login -u system:admin
+$ oc patch oauthclient/openshift-web-console -p '{"redirectURIs":["https://localhost:9001/"]}'
+```
 
->Warning: Make sure you're using the latest version of Node.js and NPM
+You may also need to update your service catalog project's deployment to update teh apiserver YAML to allow for port 9001 if it is configured for port 9000 only
```
spadgett: teh -> the

README.md
```diff
@@ -11,13 +11,29 @@ A component library created for the cataloging components for [The OpenShift Web
...
+```
+# run the server
 $ npm run start
 ```
 
-Go to [https://localhost:9000](https://localhost:9000) in your browser.
+Go to [https://localhost:9001](https://localhost:9001) in your browser. (though it should auto-launch)
+This will watch for asset changes.
+
+You may need to patch oc to allow for port 9001:
```
spadgett: You will need to...

jwforres:
Yeah this is what we do in the main console.

README.md
spadgett: If this is for CORS allowed origins, it's already an option when deploying the catalog using the template... We should probably change the template default to allow 9001 as well.

cc @jwforres

jeff-phillips-18:
I've done it a bit differently. I've added a variable to config.js to allow specifying the port. This can then be set in config.local.js once and used throughout.

README.md
jwforres: Yeah you can take this warning out, I already updated the template to take the port off so it will work for any port.

jeff-phillips-18:
Updated Readme

README.md
```diff
@@ -11,13 +11,27 @@ A component library created for the cataloging components for [The OpenShift Web
...
+# run the server
 $ npm run start
 ```
 
-Go to [https://localhost:9000](https://localhost:9000) in your browser.
+Go to [https://localhost:9001](https://localhost:9001) in your browser. (though it should auto-launch)
+This will watch for asset changes.
 
->Warning: Make sure you're using the latest version of Node.js and NPM
+You will need to patch oc to allow for port 9001:
```
spadgett: Nit: you're not patching `oc`, you're patching the web console oauth client

jeff-phillips-18:
OK, back to where it was with commits squashed.

PR #22, opened 2017-03-17
Removing duplicate css (Font Awesome and Bootstrap are in PatternFly)

rhamilto:


spadgett:
@jeff-phillips-18 Are the files under `app/` for the showcase app and `src/` for the actual components?

jeff-phillips-18:
@spadgett Yes, exactly.

PR #25, opened 2017-03-20
Add responsive handling to landing page

jeff-phillips-18:
This handles responsive resizing in the landing page. This doesn't quite meet the mobile designs yet, but it at least handles small sizes adequately.


dist/origin-web-catalogs.css
```diff
@@ -60,37 +60,56 @@
   padding-bottom: 40px;
 }
 .landing-main-area {
-  margin: 15px 320px 15px 20px;
+  margin: 15px 20px;
   padding-right: 10px;
 }
+@media (min-width: 768px) {
```
dtaylor113: Should we be using constants for this?  I noticed them here:  https://github.com/openshift/origin-web-console/blob/master/app/scripts/app.js#L435

spadgett: We should be using the Bootstrap variables for media queries.

http://getbootstrap.com/css/#grid-media-queries

```css
@media (min-width: @screen-sm-min) { ... }
```

jeff-phillips-18:
Updated to use the variables and squashed commits

spadgett:
> Updated to use the variables

Hey @jeff-phillips-18 I don't see this update?

jeff-phillips-18:
Hmm... very odd. I know i made the change. I will update now.

jeff-phillips-18:
Something is diconnected here. My repo has the changes:

https://github.com/jeff-phillips-18/origin-web-catalog/blob/landing/src/styles/landing-page.less#L17

Should I close this PR and open a new one?

jeff-phillips-18:
@spadgett  My bad. The requested changes are in this PR. I was looking at the css file 😊

src/components/landing-page/landing-page.controller.ts
```diff
@@ -88,10 +92,20 @@ export class LandingPageController implements angular.IController {
     }
 
     angular.element(this.$window).bind('resize', this.onWindowResize);
```
spadgett: We probably want to debounce calls to this since resize triggers a lot of events

spadgett:
Thanks, missed that too!

src/components/landing-page/landing-page.controller.ts
```diff
@@ -88,10 +92,20 @@ export class LandingPageController implements angular.IController {
...
+    this.onWindowResize();
   };
 
-  private onWindowResize = (event: any) => {
-    angular.element(this.snapElement).css('min-height', this.$window.innerHeight + 'px');
+  private onWindowResize = () => {
+    var width = this.$window.innerWidth;
+    var w: any = $(window)[0];
+
+    if (width > w.patternfly.pfBreakpoints.tablet) {
```
spadgett: It looks like we have special handling for different widths, but then hiding the arrows at mobile in CSS?

src/components/landing-page/landing-page.controller.ts
jeff-phillips-18: Good call, I will add for both resize and scroll.

src/components/landing-page/landing-page.controller.ts
jeff-phillips-18: Yes, this code is adding enough height to the body area so that when snapping down, the top section (build applications) gets completely scrolled off screen. But, this only happens in non-mobile mode since there is no snapping in mobile mode.

spadgett: Got it, thanks

jeff-phillips-18:
OK, added debouncing and fixed unit tests.

PR #33, opened 2017-03-24
Bumping version to 0.0.3

jeff-phillips-18:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "origin-web-catalogs",
```
spadgett: Just noticed this does not match bower.json

jeff-phillips-18:
Thanks @spadgett lots of catalogs that should be catalog in there. Fixed.

package.json
```diff
@@ -17,14 +17,14 @@
   },
   "repository": {
     "type": "git",
-    "url": "https://github.com/origin/origin-web-catalogs.git"
+    "url": "https://github.com/origin/origin-web-catalog.git"
   },
-  "author": "Jeff Phillips, David Taylor",
+  "author": "Red Hat, Inc",
   "license": "MIT",
```
spadgett: @jeff-phillips-18 This should probably be Apache License 2.0 since the rest of OpenShift is

cc @jwforres

jwforres: Yes, openshift is Apache 2.0

jwforres: I'm not seeing a LICENSE file at all, can we go ahead and fix that too, grab the one in origin-web-console

PR #35, opened 2017-03-24
Add content to Projects Summary component

jeff-phillips-18:
This is mostly functionality based. The final css styling and layout is still being worked on but this PR gives us the ability to view and perform CRUD operations on projects from the projects summary and hence landing page.


app/mockServices/mockData.service.ts
```diff
@@ -19,30 +28,188 @@ class Data {
...
+  }
+
+  public delete(resource: any, name: any, context: any, opts: any) : angular.IPromise < any > {
+    let deferred = this.$q.defer();
+    var data: any = this.getMockData(resource, context);
+
+    _.remove(data.by(), function(nextData: any) {
+      return nextData.metadata.name === name;
+    });
+    setTimeout(() => {
```
spadgett: Should this be `$timeout` so deferred is resolved in a digest loop? Here and below.

src/components/projects-summary/projects-summary.controller.ts
```diff
@@ -1,23 +1,144 @@
...
+  public onProjectsUpdate = (projectData: any) => {
+    var projects: any = _.toArray(projectData.by('metadata.creationTimestamp'));
+    var orderByDate: any = this.$filter('orderObjectsByDate');
+    this.ctrl.projects = orderByDate(projects, true);
+
+    this.ctrl.maxDisplayProjectsReached = this.ctrl.projects.length > this.maxDisplayProjects;
+    this.ctrl.totalProjects = this.ctrl.projects.length;
+    if (this.ctrl.maxDisplayProjectsReached) {
+      this.ctrl.projects.splice(this.maxDisplayProjects, this.ctrl.projects.length);
+    }
```
spadgett: I believe this could be

```js
this.ctrl.projects = _.take(this.ctrl.projects, this.maxDisplayProjects);

File: src/components/projects-summary/projects-summary.html [hunk 3]
```diff
@@ -1,11 +1,75 @@
...
+  </div>
+
+  <div ng-if="$ctrl.showGetStarted">
+    <h2>Getting Started</h2>
+    <div class="getting-started-panel">
+      {{$ctrl.gettingStartedText}}
+    </div>
+  </div>
+  <div ng-if="!$ctrl.showGetStarted" class="catalogs-project-summary-list">
+    <div ng-if="$ctrl.maxDisplayProjectsReached" class="projects-count-area">
```
spadgett: Maybe just check if total > max rather than having a separate variable tracking this?

src/components/projects-summary/projects-summary.controller.ts
```diff
@@ -1,23 +1,144 @@
...
+  };
+
+  public editProject = (project: any, parentId: string) => {
+    var projectDiv = jQuery(this.$element).find('#' + parentId)[0];
+    var listDiv = jQuery(this.$element).find('#catalogs-projects-summary-list')[0];
+    var width = this.$window.innerWidth;
+    var w: any = jQuery(window)[0];
+
+    if (width > w.patternfly.pfBreakpoints.tablet) {
+      this.ctrl.editPanelTopPosition = listDiv.offsetTop + projectDiv.offsetTop + projectDiv.clientHeight - 10;
```
spadgett: I'd to see if there's a way to do this with CSS, but can be a follow on.

src/components/projects-summary/projects-summary.html
```diff
@@ -1,11 +1,75 @@
...
+      <span class="counter-text">{{$ctrl.maxDisplayProjects}}</span>
+      <span>of</span>
+      <span class="counter-text">{{$ctrl.totalProjects}}</span>
+      <span>Projects</span>
+      <a href="#" class="projects-view-all" ng-click="$ctrl.showAllProjects()">View All</a>
+    </div>
+    <div id="catalogs-projects-summary-list">
+      <div class="catalogs-edit-project-panel" ng-if="$ctrl.showEditProjectPanel" style="top: {{$ctrl.editPanelTopPosition}}px;">
```
spadgett: `ng-style` here?

src/components/projects-summary/projects-summary.html
```diff
@@ -1,11 +1,75 @@
...
+      {{$ctrl.gettingStartedText}}
+    </div>
+  </div>
+  <div ng-if="!$ctrl.showGetStarted" class="catalogs-project-summary-list">
+    <div ng-if="$ctrl.maxDisplayProjectsReached" class="projects-count-area">
+      <span class="counter-text">{{$ctrl.maxDisplayProjects}}</span>
+      <span>of</span>
+      <span class="counter-text">{{$ctrl.totalProjects}}</span>
+      <span>Projects</span>
+      <a href="#" class="projects-view-all" ng-click="$ctrl.showAllProjects()">View All</a>
```
spadgett: Can't this be `href=""` here and below? Then we're not adding or changing the URL hash.

https://docs.angularjs.org/api/ng/directive/a

> Modifies the default behavior of the html a tag so that the default action is prevented when the href attribute is empty.

src/components/projects-summary/projects-summary.html
```diff
@@ -1,11 +1,75 @@
...
+        <div class="catalogs-create-project-title">
+          <span>Edit Project - </span>
+          <span>{{$ctrl.edittingProject | displayName}}</span>
+          <a class="pull-right" ng-click="$ctrl.closeEditProjectPanel()">
+            <span class="catalogs-create-project-close pficon pficon-close"></span>
+          </a>
+        </div>
+        <edit-project project="$ctrl.edittingProject" is-dialog="true" alerts="$ctrl.alerts" redirect-action="$ctrl.onEditProject" on-cancel="$ctrl.closeEditProjectPanel"></edit-project>
+      </div>
+      <div ng-repeat="project in $ctrl.projects track by $index" class="project-tile">
```
spadgett: `track by (project | uid)`

src/components/projects-summary/projects-summary.html
```diff
@@ -1,11 +1,75 @@
...
+      <span>of</span>
+      <span class="counter-text">{{$ctrl.totalProjects}}</span>
+      <span>Projects</span>
+      <a href="#" class="projects-view-all" ng-click="$ctrl.showAllProjects()">View All</a>
+    </div>
+    <div id="catalogs-projects-summary-list">
+      <div class="catalogs-edit-project-panel" ng-if="$ctrl.showEditProjectPanel" style="top: {{$ctrl.editPanelTopPosition}}px;">
+        <div class="arrow"></div>
+        <div class="catalogs-create-project-title">
+          <span>Edit Project - </span>
```
spadgett: `&ndash;`?

Do we need the extra spans?

src/components/projects-summary/projects-summary.html
```diff
@@ -1,11 +1,75 @@
...
+                  stay-on-current-page="true"
+                  alerts="$ctrl.alerts">
+              </delete-link>
+            </li>
+          </ul>
+        </div>
+        <div id="{{'catalogs-summary-project-' + $index}}" class="project-tile-header">
+          <span ng-if="project.statusIconClass" class="project-status {{project.statusIconClass}}"></span>
+          <a href="#" ng-click="$ctrl.handleProjectClick(project)" class="project-title">{{project | displayName}}</a>
+          <span class="project-date">{{project.metadata.creationTimestamp | date: 'MMM d yyyy'}}</span>
```
spadgett: `date : 'mediumDate'`? Then it's localized.

https://docs.angularjs.org/api/ng/filter/date

src/components/projects-summary/projects-summary.html
```diff
@@ -1,11 +1,75 @@
...
+        </div>
+        <edit-project project="$ctrl.edittingProject" is-dialog="true" alerts="$ctrl.alerts" redirect-action="$ctrl.onEditProject" on-cancel="$ctrl.closeEditProjectPanel"></edit-project>
+      </div>
+      <div ng-repeat="project in $ctrl.projects track by $index" class="project-tile">
+        <div class="dropdown  dropdown-kebab-pf" uib-dropdown="">
+          <button class="btn btn-link uib-dropdown-toggle" type="button" id="dropdownKebab" aria-haspopup="true" aria-expanded="true" uib-dropdown-toggle>
+            <span class="fa fa-ellipsis-v"></span>
+          </button>
+          <ul class="uib-dropdown-menu dropdown-menu-right" aria-labelledby="dropdownKebab">
+            <li><a href="#" ng-click="$ctrl.editProject(project, 'catalogs-summary-project-' + $index)">Edit</a></li>
```
spadgett: Are we going to have problems where the dialog is in the wrong place if the list changes during an edit?

src/components/projects-summary/projects-summary.html
```diff
@@ -1,11 +1,75 @@
...
+            </li>
+          </ul>
+        </div>
+        <div id="{{'catalogs-summary-project-' + $index}}" class="project-tile-header">
+          <span ng-if="project.statusIconClass" class="project-status {{project.statusIconClass}}"></span>
+          <a href="#" ng-click="$ctrl.handleProjectClick(project)" class="project-title">{{project | displayName}}</a>
+          <span class="project-date">{{project.metadata.creationTimestamp | date: 'MMM d yyyy'}}</span>
+        </div>
+        <div class="project-description">
+          {{project | description}}
```
spadgett: Do long descriptions get truncated properly?

src/components/projects-summary/projects-summary.html
```diff
@@ -1,11 +1,75 @@
...
     <span class="create-button-text">Create Project</span>
   </button>
-  <h2>Getting Started</h2>
-  <div class="getting-started-panel">
-    {{$ctrl.gettingStartedText}}
+  <div class="catalogs-create-project-panel" ng-if="$ctrl.showNewProjectPanel">
+    <div class="arrow"></div>
+    <div class="catalogs-create-project-title">
+      <span>Create Project</span>
+      <a class="pull-right" ng-click="$ctrl.closeNewProjectPanel()">
```
spadgett: `href=""` so you can focus it with the keyboard.

src/styles/projects-summary.less
```diff
@@ -17,3 +17,140 @@
...
+  }
+
+  .catalogs-create-project-title {
+    font-size: 16px;
+    font-weight: 500;
+  }
+  .catalogs-create-project-close {
+    font-size: 18px;
+    float: right;
+    cursor: pointer;
```
spadgett: Shouldn't be necessary if you give the `a` element an `href`

src/styles/projects-summary.less
```diff
@@ -17,3 +17,140 @@
   }
 }
 
+.catalogs-create-project-panel, .catalogs-edit-project-panel {
```
spadgett: Nit: Should these names start with `catalog-*` instead of `catalogs-`?

spadgett:
I pulled down the changes, and the project list works really well. Thanks @jeff-phillips-18 

One comment after using it: I think we should make the entire row clickable and highlight on hover. The pointer cursor hints that I can click anyway, but only the link is actually clickable.

spadgett:
We'll also need a way to disambiguate two projects with the same display name.

src/components/projects-summary/projects-summary.html
spadgett: Opened https://github.com/openshift/origin-web-common/pull/21

app/mockServices/mockAPI.service.ts
```diff
@@ -0,0 +1,42 @@
...
+export class APIService implements IAPIService {
+  public static $inject = ['$filter'];
+
+  private $filter : any;
+
+  constructor ($filter: any) {
+    this.$filter = $filter;
+  }
+
+  public kindToResource(kind: string, humanize: boolean) {
```
jeff-phillips-18: When we are in mock mode, we do not include the common services at all. Doing so requires a bunch of checks that fail when not having a backend to connect to. So, we do need to mock any particular service calls that get used even when we are in mock mode.

app/mockServices/mockAlertMessage.service.ts
```diff
@@ -0,0 +1,46 @@
+interface IAlertMessageService {
+  addAlert(alert: any) : any;
+  getAlerts() : any;
+  clearAlerts(): any;
+  isAlertPermanentlyHidden(alertID: any, namespace: any): any;
+  permanentlyHideAlert(alertID: any, namespace: any): any;
+}
+
+
+export class AlertMessageService implements IAlertMessageService {
```
jeff-phillips-18: Same issue here. The idea is not to duplicate the real implementation just to have a basic implementation that allows for testing without using the common services.

src/components/projects-summary/projects-summary.controller.ts
jeff-phillips-18: We could if we added the editPanel to each project tile (similar to uib dropdowns). This implementation requires only 1 panel that gets placed based on the project being edited.

src/components/projects-summary/projects-summary.html
jeff-phillips-18: Yeah, agreed. I did this originally to ease unit testing but found it didn't really make a difference.

src/components/projects-summary/projects-summary.html
jeff-phillips-18: I guess that would be a problem. Thanks I will change this implementation.

src/components/projects-summary/projects-summary.html
jeff-phillips-18: yup... thanks

jeff-phillips-18:
Didn't realize they could have the same display name. what is done in console?

src/components/projects-summary/projects-summary.html
spadgett: It might be weird for the dialog to move as you type, too. I can't decide what's best. It was more an observation :/

spadgett:
When there's a display name, we show both name and display name in the projects list. Some users might only recognize `project.metadata.name` since that's the value that's used with the CLI.

screen shot 2017-03-26 at 6 55 43 pm

src/components/projects-summary/projects-summary.html
jeff-phillips-18: Yes, they are limited to one line here. Waiting on final designs on the list items

jeff-phillips-18:
@spadgett I'd like to wait on making the entire tile clickable as well as adding the project.metadata.name until final UX designs are complete. I know there are other changes as well and it would be better to handle them all at once.

spadgett:
> I'd like to wait on making the entire tile clickable as well as adding the project.metadata.name until final UX designs are complete

OK for me.

app/mockServices/mockProjects.service.ts
```diff
@@ -0,0 +1,52 @@
...
+  private $q : any;
+
+  constructor ($q: angular.IQService) {
+    this.$q = $q;
+  }
+
+  public canCreate() {
+    let deferred = this.$q.defer();
+
+    setTimeout(() => {
```
dtaylor113: I've seen usage of $timeout and setTimeout() in origin-web-catalog, do we have a preference/best practice?

spadgett: We use `$timeout` to trigger a digest loop that rerenders the views.

We use `setTimeout` specifically when we know the view does not need to be reevaluated (e.g., updating duration values or updating c3 charts using the c3.js API)

src/components/projects-summary/projects-summary.controller.ts
```diff
@@ -1,23 +1,144 @@
 import * as angular from 'angular';
+import * as _ from 'lodash';
+import * as jQuery from 'jquery';
```
dtaylor113: I see we have 'import * as $ from 'jquery';' in the landing-page.controller.js, it would be good to standardize on one or the other.

jeff-phillips-18:
@spadgett I believe I have addressed  all of your comments.

rhamilto:
@jeff-phillips-18, how close do you think this is?  I'm hoping to work on sidebar styling.

app/mockServices/mockAlertMessage.service.ts
spadgett: I'd suggest then we don't change localStorage for mock data.

src/components/projects-summary/projects-summary.html
```diff
@@ -1,11 +1,74 @@
...
+        <span class="catalog-create-project-close pficon pficon-close"></span>
+      </a>
+    </div>
+    <create-project alerts="$ctrl.alerts" is-dialog="true" redirect-action="$ctrl.onNewProject" on-cancel="$ctrl.closeNewProjectPanel"></create-project>
+  </div>
+
+  <div ng-if="$ctrl.showGetStarted">
+    <h2>Getting Started</h2>
+    <div class="getting-started-panel">
+      {{$ctrl.gettingStartedText}}
```
spadgett: Where is this getting set?

jeff-phillips-18:
@rhamilto I see you have been talking with @cshinn. He is the contact for where that is at.

src/components/projects-summary/projects-summary.html
jeff-phillips-18: It isn't yet. Again, waiting for final designs so this is just blank for now.

jeff-phillips-18: removed the localStorage usage

jeff-phillips-18:
Added #39 and #40

jeff-phillips-18:
Squashed.

Issue #27, opened 2017-03-21
Builder image create flow


spadgett:
I wanted to start discussion on supported builder images in the catalog. These are the source-to-image language builders like Wildfly, Node.js, Ruby, etc., that we support today. I have some concerns, both technical and around the UX.

Note I'm talking specifically about the language builders and not services in the service catalog.

### It's a complex form

For context, here is the create form will all sections expanded. It's enormous. I've scaled down the image so you can see how tall the form is. For some fields, it pops up additional dialogs and file browsers.

![create-form](https://cloud.githubusercontent.com/assets/1167259/24164409/ff08e78a-0e43-11e7-920e-8c1a34245099.png)

I think this form is too complex as-if for the ordering panel. We have a card to break the form into a wizard in our backlog.

https://trello.com/c/wxH53dyz

### Many dependencies

The related problem is the number of services and Angular directives this form uses. The list below is not everything, but these are some of the larger services and directives it uses. (I have some notes with a more complete list. It's too large to include here.) These directives are reused for other editors, so changes to them will affect many places in the web console.

*I'd guess pulling all of these in will put close to half of the web console code into the common repo.* Pulling this much code into the common repository will make maintenance of origin-web-console difficult.

**Services**

ApplicationGenerator
LimitRangesService
MetricsService (no metrics HPA warning, could be factored out)
HPAService
QuotaService
SecretsService
ImagesService
ModalsService (used, but could be factored out)

**Editor Directives**

osc-secrets
create-secret
osc-routing
osc-file-input
key-value-editor
osc-autoscaling
edit-request-limit
label-editor

**Other Dependencies**

Ace Editor

### Proposal

I suggest we don't try to pull all of this into the catalog repo. The tiles can have a click handler registered with the catalog component. Then we can decouple the ordering panel from the catalog itself, and some of the actions can be handled by existing code in the origin-web-console repo when needed.

Normal service classes can use the ordering panel as in the current designs. For builders, we might have two forms:

1. A simple form that uses the ordering panel and prompts for a few fields (name, Git repository, and possibly source secrets for private Git repositories)
1. A more complex form -- like the current form -- that is linked out from the ordering panel.

Longer term we can look at improving the create form by making it multiple wizard steps or possibly removing less important items to keep it manageable.

When the catalog is used outside of origin-web-console, we can show the simple form with a link out to the web console for the more advanced create options.

[...]

PR #36, opened 2017-03-24
Add basic support for provisioning a service

spadgett:
This adds initial support for creating a service instance from the ordering panel.

@jwforres @jeff-phillips-18 @dtaylor113


spadgett:
OK the basic function is working now.

- [x] Removed any hard-coded, mock data from the views
- [x] Real plans are displayed with a description if set
- [x] You can select from a list of real projects when provisioning
- [x] The service instance is successfully created
- [x] All dialog buttons should be functional now

Follow-on work:

- [ ] Add parameters
- [ ] Handle async provisioning
- [ ] Let users create a project from the ordering panel
- [ ] Decide how to handle what happens when there is a single plan. Right now the template broker always has a single "default" plan with no description, which doesn't look great.

I haven't mocked any project data since it looks like @jeff-phillips-18 is adding that in #35, so you won't be able to get past the configuration panel with mock data. Also @jeff-phillips-18 I am still unable to run `npm test`, so it's difficult to add tests.

@pmorie @jim-minter @bparees When I create a service instance for one of the template broker service classes, the service instance is created, but nothing else is actually created. Not sure if this is expected. I'm not passing any parameters, but the templates I'm using should have defaults for anything required. This is the service instance I'm creating:

https://github.com/openshift/origin-web-catalog/pull/36/files#diff-7a34cbbd85f9a19385ee34c6b893075fR109

I also can't delete service instances created using template broker service classes.

src/components/order-service/order-service-plans.html
```diff
@@ -0,0 +1,22 @@
+<!--
+  TODO:
+  - Improve plan display
+  - Skip plan selection if only one
+-->
+<div ng-repeat="plan in $ctrl.serviceClass.resource.plans" class="col-md-6">
+  <div class="title">Plan {{plan.name}}</div>
```
jwforres: plan.osbMetadata.displayName || plan.name

src/components/order-service/order-service-plans.html
```diff
@@ -0,0 +1,22 @@
...
+  <!-- TODO : truncate long text -->
+  <div ng-if="plan.description" class="sub-title first">{{plan.description}}</div>
```
jwforres: I'm thinking we may want to consider having plan.osbMetadata.bullets show up underneath the plan description, bullets should be a ul of some kind.  a TODO here would be good for now.

src/components/order-service/order-service-review.html
```diff
@@ -1,27 +1,42 @@
...
-      Launch Service
-    </button>
+  <div ng-if="!$ctrl.error">
+    <div class="title">Your Order is Complete! <span class="fa fa-check success-check"></span></div>
+    <div class="sub-title center">
+      Continue to your project to bind the service to your application.
+    </div>
+    <div class="launch-service">
+      <!-- FIXME: Only works in the console -->
+      <!-- TODO: Provide direct link to bind action? -->
```
jwforres: we may want the option to go straight to bind OR just go to the project, maybe a button and a link or something?

src/components/order-service/order-service.controller.ts
```diff
@@ -35,9 +44,16 @@ export class OrderServiceController implements angular.IController {
   }
 
   public stepClick(step: any) {
-    if (step.visited) {
-      this.gotoStep(step);
+    // Prevent returning to previous steps if the order is complete.
+    if (this.ctrl.orderComplete) {
```
jwforres: the cursor is still showing a hand even though the action is prevented

src/components/order-service/order-service.controller.ts
```diff
@@ -48,20 +64,25 @@ export class OrderServiceController implements angular.IController {
...
+    this.DataService.create({
+      group: 'servicecatalog.k8s.io',
+      resource: 'instances'
+    }, null, serviceInstance, {
+      namespace: this.ctrl.selectedProject.metadata.name
+    }).then(() => {
+      // TODO: Handle async responses
+      this.ctrl.orderComplete = true;
+      this.ctrl.error = null;
+      this.gotoStep(this.ctrl.steps[2]);
```
jwforres: we will probably want named references to steps since the number of steps is going to change depending on the service class, but we can deal with that when we start changing the number of steps.

src/components/order-service/order-service.controller.ts
```diff
@@ -72,7 +93,34 @@ export class OrderServiceController implements angular.IController {
...
+  }
+
+  private makeServiceInstance() {
+    let serviceClassName = _.get(this, 'ctrl.serviceClass.resource.metadata.name');
+    let serviceInstance = {
+      kind: 'Instance',
+      apiVersion: 'servicecatalog.k8s.io/v1alpha1',
+      metadata: {
+        namespace: this.ctrl.selectedProject.metadata.name,
+        generateName: serviceClassName + '-'
```
jwforres: does generateName have a length restriction? just wondering if we get into issues when serviceClassName is that max length for service classes

src/components/order-service/order-service.controller.ts
```diff
@@ -72,7 +93,34 @@ export class OrderServiceController implements angular.IController {
...
+      },
+      spec: {
```
jwforres: I'm not sure when we update this to handle async whether thats a URL param or if its a field on the object,  @pmorie ?  not seeing anything in the InstanceSpec in types.go

src/components/order-service/order-service.html
```diff
@@ -12,13 +12,20 @@
             <div ng-include="'components/order-service/order-service-details.html'"></div>
           </span>
            <span ng-if="$ctrl.currentStep.id === 2 || $ctrl.currentStep.id === 3">
-            <div class="long-description">
-              {{$ctrl.serviceLongDescription}}
+            <div ng-if="!$ctrl.selectedPlan.description">
+              <div ng-include="'components/order-service/order-service-details.html'"></div>
+              <div ng-if="$ctrl.selectedPlan" class="h3">
+                Selected Plan: {{$ctrl.selectedPlan.name}}
```
jwforres: selectedPlan.osbMetadata.displayName if it exists

src/components/order-service/order-service.controller.ts
spadgett: Yeah, I'm not sure. I had some code to handle it on the client, but ran into #37

src/components/order-service/order-service.controller.ts
spadgett: Agreed

src/components/order-service/order-service.controller.ts
spadgett: Yeah. We have the problem already for unvisited steps, too. Was holding off trying to fix too many styles since there is some attention needed there

src/components/order-service/order-service-details.html
```diff
@@ -11,9 +11,13 @@
     </li>
   </ul>
 </div>
-<h2 class="sub-title">General Information</h2>
+<h2 class="sub-title">
+  General Information
+</h2>
 <div>{{$ctrl.description}}</div>
+<!-- TODO: truncate long text -->
```
spadgett: Our `linkify` is not yet in common. I'll update to use Angular `linky` for now

src/components/order-service/order-service.controller.ts
spadgett: From my reading of the godoc, the prefix will be truncated if necessary.

https://github.com/spadgett/origin/blob/9b26d5a134bd12767ab7a466c44aeef19a1af13a/vendor/k8s.io/client-go/pkg/api/types.go#L68-L73

> The provided value must be valid within the rules for Name, and may be truncated by the length of the suffix required to make the value unique on the server.

src/components/order-service/order-service-configure.html
```diff
@@ -1,77 +1,31 @@
...
+        <!-- TODO: let users create projects -->
+        <select
+            id="project"
+            ng-model="$ctrl.selectedProject"
+            ng-options="project as (project | displayName) for project in $ctrl.projects track by (project | uid)"
+            class="form-control">
+        </select>
       </div>
     </div>
+    <!-- TODO: add parameters -->
```
spadgett: Added one back.

spadgett:
Thanks @jwforres, updated

src/components/order-service/order-service.controller.ts
jwforres: so we could still potentially violate the rules for name since we are adding the '-' character

jwforres: meaning that how read that comment is that it has to be a valid name, and if they need to truncate it to append to it then they will.  to be a valid name it has length limitations i thought

spadgett: Pretty sure a trailing `-` is used with generate name frequently. E.g.

```
apiVersion: v1
kind: Pod
metadata:
  name: hello-openshift-1-mdd3f
  generateName: hello-openshift-1-
```

>  may be truncated by the **length of the suffix**

"Length of the suffix" implies to me it's truncated to meet length limits, but I could be wrong.

spadgett: Or are you saying we'd be one character over length potentially by adding the `-`?

jwforres:
Yeah i'm saying i think the prefix we submit has to be no longer than the
allowed length for name, and we could be one over.

spadgett:
Looks like `maskTrailingDash` is called before `generateName` is validated, which should be OK even if one over the maxlength.

https://github.com/spadgett/origin/blob/9b26d5a134bd12767ab7a466c44aeef19a1af13a/vendor/k8s.io/kubernetes/pkg/api/validation/validation.go#L240-L247

Then `GenerateName` doesn't ever make a name longer than 63 chars.

https://github.com/spadgett/origin/blob/9b26d5a134bd12767ab7a466c44aeef19a1af13a/vendor/k8s.io/client-go/pkg/api/generate.go#L47-L64

jwforres:
Ok

On Mar 25, 2017 5:32 PM, "Sam Padgett" <notifications@github.com> wrote:

Looks like maskTrailingDash is called before generateName is validated,
which should be OK even if one over the maxlength.

https://github.com/spadgett/origin/blob/9b26d5a134bd12767ab7a466c44aee
f19a1af13a/vendor/k8s.io/kubernetes/pkg/api/validation/
validation.go#L240-L247

Then GenerateName doesn't ever make a name longer than 63 chars.

https://github.com/spadgett/origin/blob/9b26d5a134bd12767ab7a466c44aee
f19a1af13a/vendor/k8s.io/client-go/pkg/api/generate.go#L47-L64

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/openshift/origin-web-catalog/pull/36#issuecomment-289240908>,
or mute the thread

bparees:
> @pmorie @jim-minter @bparees When I create a service instance for one of the template broker service classes, the service instance is created, but nothing else is actually created.

what do you mean by the service instance is created?  like the service catalog creates the object?

I don't know much about how you've wired this up, but a few thoughts:

1) the template broker is disabled by default, have you enabled it in the master config?
2) there are some required parameters for any template broker provision request:  we need to know the userid we're supposed to impersonate and the namespace we're supposed to provision into:
https://github.com/openshift/origin/blob/master/pkg/template/api/constants.go#L26-L38


I would have expected you to get a more clear error if any of those things were problems, but i don't for sure 1) what the template broker returns if there's insufficient information or 2) what the service catalog does w/ a failed response from the broker.

spadgett:
> what do you mean by the service instance is created? like the service catalog creates the object?

Yes. I make a POST request for a `servicecatalog.k8s.io/Instance`. The catalog responds with 201 created and the object is there.

> the template broker is disabled by default, have you enabled it in the master config?

Yes. I'm seeing the service classes, so I believe I have this set up correctly.

> there are some required parameters for any template broker provision request: we need to know the userid we're supposed to impersonate and the namespace we're supposed to provision into

I'm sure this is the problem. Is there are any reason you can't use the namespace of the service instance, or is that not passed through?

spadgett:
> there are some required parameters for any template broker provision request: we need to know **the userid we're supposed to impersonate**

Is this temporary? Otherwise won't I be able to impersonate other users through the template broker?

spadgett:
Still doesn't create anything.

```yaml
apiVersion: servicecatalog.k8s.io/v1alpha1
kind: Instance
metadata:
  name: nodejs-mongo-persistent-6sfz7
  generateName: nodejs-mongo-persistent-
  namespace: provision-test
  selfLink: >-
    /apis/servicecatalog.k8s.io/v1alpha1/namespaces/provision-test/instances/nodejs-mongo-persistent-6sfz7
  uid: 26b11d50-122c-11e7-b814-0242ac110005
  resourceVersion: '77'
  creationTimestamp: '2017-03-26T13:57:29Z'
  finalizers:
    - kubernetes
spec:
  serviceClassName: nodejs-mongo-persistent
  planName: default
  parameters:
    template.openshift.io/namespace: provision-test
    template.openshift.io/requester-username: sam
  osbGuid: 1fb6d16e-b689-4cde-b87c-8bfabf134699
status:
  conditions: []
```

bparees:
> I'm sure this is the problem. Is there are any reason you can't use the namespace of the service instance, or is that not passed through?

we don't even see the service instance object.  that's a resource created by the SC, for the SC, for its own tracking purposes.  All the broker gets is a call from the SC passing the provisionrequest object.

(and i would not have expected the service instance object to be created in the user's project, i'd expect it to be created in some system namespace used by the SC, but i'm less confident of that.  The point being that the SC doesn't even have to be running in the same cluster as the template broker, nor does the broker even have to be running in a cluster, etc... so the SC and the Broker are (should be) very decoupled).

> Is this temporary? Otherwise won't I be able to impersonate other users through the template broker?

the assumption/design is that the SC confirms the user's identity and passes it to the broker.  The broker trusts the SC and trusts that it has confirmed the userid that is being passed, so the broker will impersonate that user on the basis of its assumption that the SC has validated the user already.  (And of course the broker will not accept provisionrequests from anything except the SC that it trusts)

at least that is the rough plan as i understand it today, but there are still some on-going conversations about how we get there.

> Still doesn't create anything.

ok, we'll probably need to do some debugging w/ Jim on monday to ensure that the templatebroker is even receiving a provisionrequest call from the service catalog, and if so, where things are going wrong from there (and if not, why not).  (And why the SC is not reporting a failure/error when clearly the flow is not successful)

spadgett:
@bparees We should catch up early this week. Our team had different assumptions.

> we don't even see the service instance object. that's a resource created by the SC, for the SC, for its own tracking purposes. All the broker gets is a call from the SC passing the provisionrequest object.

Understood, and I don't know what's in the provision request. I assumed the namespace would be included (somehow).

> (and i would not have expected the service instance object to be created in the user's project, i'd expect it to be created in some system namespace used by the SC, but i'm less confident of that. 

This is not at all what I understood. The web console is creating these directly in the user's namespace to provision services. The web console will also create the binding resources.

>The point being that the SC doesn't even have to be running in the same cluster as the template broker, nor does the broker even have to be running in a cluster, etc... so the SC and the Broker are (should be) very decoupled).

Agree that the broker can run anywhere. I don't follow how having service catalog resources like service instances, bindings, and pod presets in the user's namespace prevents that.

> the assumption/design is that the SC confirms the user's identity and passes it to the broker. The broker trusts the SC and trusts that it has confirmed the userid that is being passed, so the broker will impersonate that user on the basis of its assumption that the SC has validated the user already. (And of course the broker will not accept provisionrequests from anything except the SC that it trusts)

Doesn't this require special knowledge in the service catalog that is talking to the template broker? How does it inject these extra parameters otherwise?

bparees:
> @bparees We should catch up early this week. Our team had different assumptions.

i think we may simply be talking about different pieces of the system. anyway yes, let's talk monday.

> This is not at all what I understood. The web console is creating these directly in the user's namespace to provision services. The web console will also create the binding resources.

i wouldn't have expected the web console to create anything, i would have expected it to invoke the SC which in turn invokes the broker.  (And whether the SC creates any resources, either in a global or user namespace to track service instances/bindings would be up to the SC, not the web console) But i haven't been involved in any of the webconsole/SC architecture so certainly you'd know that design better than I.

Regardless, the broker doesn't care about any resources created by the SC, it only cares that it receives a provisionrequest api call from the SC. (and bind call, etc).

> Agree that the broker can run anywhere. I don't follow how having service catalog resources like service instances, bindings, and pod presets in the user's namespace prevents that.

my only point is that the broker doesn't care about those resources and may not even have a way to see those resources.


> Doesn't this require special knowledge in the service catalog that is talking to the template broker? How does it inject these extra parameters otherwise?

the template broker will list the userid and namespace as required parameters for the template provision action (along w/ any other parameters the template itself defines).  But yes, there is a requirement that the SC respect and populate those parameters (as opposed to soliciting them from the user).  This is all stuff that's been discussed w/ @pmorie but I don't know the details of how the SC intends to implement it.  I think there's a general assumption that the namespace being provisioned into, at least, will *always* be a provided parameter on a provision request.  it would be logical to me that the requesting username would also be always provided.

src/components/order-service/order-service-review.html
```diff
@@ -1,27 +1,42 @@
 <div class="col-md-12 center">
-  <div class="title">Your Order is Complete! <span class="fa fa-check success-check"></span></div>
-  <div class="sub-title center">Launch this service now to access your project directly.</div>
-  <div class="launch-service">
-    <button type="button" class="btn btn-default btn-primary order-btn" ng-click="$ctrl.cancelOrder()">
-      Launch Service
-    </button>
+  <div ng-if="!$ctrl.error">
+    <div class="title">Your Order is Complete! <span class="fa fa-check success-check"></span></div>
```
jeff-phillips-18: can we remove the !  It's not that exciting ;)

src/styles/overlay-panel.less
```diff
@@ -16,6 +16,7 @@
   border: 1px solid @color-pf-black-400;
   height: 50%;
   left: 20px;
+  min-height: 350px;
```
jeff-phillips-18: This is fine for now but I believe UX has plans to specify two settings for the overlay panel height. Callers can choose either large or small to set the height/max-height.

src/components/order-service/order-service-review.html
spadgett: Agree.

I think we need to look at the wording used in the dialog in general. I'm not sure we're using the right terminology.

jeff-phillips-18: Yeah, if you want to wait and we can address it all at once, I'm OK with that.

src/styles/overlay-panel.less
spadgett: Why not just let the panel size for the content? Then small/large isn't needed at all, and panel would probably work better on mobile.

Then we can give the order service wizard a fixed height if needed so switching steps doesn't resize.

jeff-phillips-18: Can't say I disagree, just that I had heard it from designers that they wanted to see two sizes available for the panel. Let's leave this for now and continue discussions as the designs firm up.

spadgett:
@bparees @jim-minter @pmorie 

This seems to be why provision wasn't working.

```
I0327 12:26:55.299959       1 controller.go:423] Processing Instance provision-test/nodejs-mongo-persistent-fsqqx
I0327 12:26:55.300005       1 controller.go:477] Creating client for Broker template-broker, URL: https://kubernetes.default.svc:443/brokers/template.openshift.io
I0327 12:26:55.300022       1 controller.go:481] Adding/Updating Instance provision-test/nodejs-mongo-persistent-fsqqx
E0327 12:26:55.303013       1 controller.go:501] Failed to get namespace during instance create (User "system:serviceaccount:service-catalog:default" cannot get namespaces in project "provision-test")
```

Adding granting additional authority to the service account, I get this error:

```
I0327 12:35:59.101000       1 controller.go:423] Processing Instance provision-test/nodejs-mongo-persistent-l0814
I0327 12:35:59.101035       1 controller.go:477] Creating client for Broker template-broker, URL: https://kubernetes.default.svc:443/brokers/template.openshift.io
I0327 12:35:59.101124       1 controller.go:481] Adding/Updating Instance provision-test/nodejs-mongo-persistent-l0814
I0327 12:35:59.107406       1 controller.go:515] Provisioning a new Instance provision-test/nodejs-mongo-persistent-l0814 of ServiceClass nodejs-mongo-persistent at Broker template-broker
I0327 12:35:59.115371       1 utils.go:69] {
  "error": "AsyncRequired",
  "description": "This service plan requires client support for asynchronous service operations."
 }
E0327 12:35:59.115408       1 controller.go:518] Error provisioning Instance provision-test/nodejs-mongo-persistent-l0814 of ServiceClass nodejs-mongo-persistent at Broker template-broker: Broker only supports this action asynchronously
I0327 12:35:59.115441       1 controller.go:617] Updating Ready condition for Instance provision-test/nodejs-mongo-persistent-l0814 to False
I0327 12:35:59.298922       1 request.go:633] Throttling request took 183.37814ms, request: PUT:http://172.30.185.89:80/apis/servicecatalog.k8s.io/v1alpha1/namespaces/provision-test/instances/nodejs-mongo-persistent-l0814/status
```

I guess I need to look at adding async to our client.

src/components/order-service/order-service.component.ts
```diff
@@ -2,7 +2,8 @@ import {OrderServiceController} from './order-service.controller';
 
 export const orderService: angular.IComponentOptions = {
   bindings: {
-    service: '<'
+    serviceClass: '<',
```
dtaylor113: 'serviceClass'?  This could be a 'imageStream' as well, right?  I thought from the Catalog users can click on serviceClasses or imageStreams to order.

src/components/order-service/order-service.controller.ts
```diff
@@ -1,31 +1,40 @@
...
     this.$scope = $scope;
+    this.$filter = $filter;
+    this.DataService = DataService;
+    this.Logger = Logger;
   }
 
   public $onInit() {
-    this.ctrl.serviceIcon = this.ctrl.service.icon ? this.ctrl.service.icon : 'pf pficon-service';
-    this.ctrl.serviceName = this.ctrl.service.name;
+    this.ctrl.serviceIcon = this.ctrl.serviceClass.icon || 'fa fa-cubes';
```
dtaylor113: I believe these fields are now called:  'iconClass' and 'imageUrl', where 'imageUrl' should be used if defined.

src/components/order-service/order-service.component.ts
spadgett: I assumed we're going to have a different component for that. They're vastly different.

dtaylor113: Hmm...I thought the ordering panel/service would be able to handle both.  I thought there were only a few differences such as imageStreams have a 'versions' dropdown, serviceClasses have 'plans'  -both have params, right?

jwforres:
No the builder image workflow is significantly different, its a complex
form. See https://github.com/openshift/origin-web-catalog/issues/27

jwforres:
Even if we do a simplified form as @spadgett suggests in the issue, the
interactions it has to make with the API are significantly different.  The
form isn't a simple list of params that map into a single model.

dtaylor113:
Ok, I didn't read #27 closely enough, I had thought it would just present the params differently between the two.  Thanks for the info.

src/components/order-service/order-service.component.ts
spadgett: No, builder images don't have parameters.

I don't think anything in the current order service controller will work with builder images. Even the metadata is different from service classes.

spadgett:
@jwforres I'm able to provision a service from the ups broker example.

spadgett:
Rebased and updated to use imageUrl/iconClass

src/components/order-service/order-service.controller.ts
jwforres: I guess my comment never made it through on this, odd.  Yes my concern was that we might be over by one character.

src/components/order-service/order-service.controller.ts
```diff
@@ -1,31 +1,41 @@
 import * as angular from 'angular';
+import * as _ from 'lodash';
+
+require("./order-service-plans.html");
+require("./order-service-configure.html");
+require("./order-service-review.html");
```
spadgett: @jeff-phillips-18 Is this the correct way to include these files?

jeff-phillips-18: Yes.

jeff-phillips-18:
+1

PR #41, opened 2017-03-27
Service View: Wrap Sub Cateogry Cards

dtaylor113:
This PR removes hard coded rows and columns which had existed in the catalog service view/list.  CSS has been updated to allow Sub Category cards to wrap.  Positioning of the expansion card/row is handled by the controller and adjusts on browser resize. 

image

Fine tuning of positioning calculations still needs to be done.  Follow up CSS adjustments will be done by @cshinn.   Works in mobile screen size, but could use some tweeks.


src/components/services-view/services-view.controller.ts
```diff
@@ -2,178 +2,188 @@ import * as angular from 'angular';
...
+        subCats = subCats.concat(categoryObj.subCategories);
+      }
+    });
+    return subCats;
+  };
+
+  public positionExpansionCard() {
+    // set the top position of the expansion card
+    // do in next cycle, after .active card set
+    this.$timeout(() => {
```
spadgett: Should only need to use `$timeout` when we know we have to wait for the digest cycle. (Shouldn't be needed when called from the resize listener).

I don't think you need to pass any delay to timeout. It should be able to run immediately after the digest cycle completes.

src/components/services-view/services-view.controller.ts
```diff
@@ -2,178 +2,188 @@ import * as angular from 'angular';
...
+      // reset all sub-cat-card 'gaps'/margin-bottoms
+      var cards = document.querySelectorAll(".services-sub-categories .sub-cat-card");
+      _.each(cards, (card) => {
+        angular.element(card).css('margin-bottom', 8 + 'px');
+      });
```
spadgett: If you use jQuery, you should be able to set this in one call without the `_.each` loop.

Suggest making the normal margin part of the `.sub-cat-card` style and just removing any inline style here instead of setting a value.

dtaylor113:
I asked Serena about that when the hifi mocks came out, she said it should wrap like we've been discussing.  The hifi mocks do not show that however.

src/components/services-view/services-view.controller.ts
dtaylor113: It is also called whenever a new filter is applied and when a card is clicked on twice (2nd time hides the row).   I did try to remove the delay, but it didn't render correctly.   I did have it working with javascript's setTimeout(), I can change it back if desired.

src/components/services-view/services-view.controller.ts
dtaylor113: The bottom-margin is being dynamically set to the current size of the expansion card/row, which changes depending on the browser window (wrapping).

spadgett: Here's it's always being set to 8px?

dtaylor113: There it is being reset to 8px (which is the default in the css) which clears any previous gaps/bottom margins of previous active card.

src/components/services-view/services-view.controller.ts
spadgett: We should understand why the delay is needed. I'd rather not set it to an arbitrary value. If my laptop is slower then yours, will the timeout fire too soon?

It's better to only call `$timeout` the times we know we need to and leave it off for resize events.

src/components/services-view/services-view.controller.ts
spadgett: Right, I'm saying instead of setting it to 8px, clear the inline margin-bottom style (which will fall back to what's set in the CSS).

dtaylor113: For example, say one clicks on the 'All' sub-cat-card which has three rows of catalog items in the expansion card/row.  The bottom-margin for the 'All' active card is set dynamically to accomidate.  Next they click on 'CI/CD' which only has one row of catalog items.  The 'All' bottom-margin is reset, and the 'CI/CD' bottom-margin is set to accommodate the one row expansion card.

jeff-phillips-18:
@rhamilto Are you working on this as well or leaving this and working the rest of the landing page layouts?

src/components/services-view/services-view.controller.ts
dtaylor113: Ah..got it, ok will fix.

spadgett: Right, I understand. I'm saying reset the style by removing it. Please try

```js
$(".services-sub-categories .sub-cat-card").css('margin-bottom', '');
```

spadgett:
> I asked Serena about that when the hifi mocks came out, she said it should wrap like we've been discussing.

Thanks @dtaylor113. I'm still unclear if the expanded section should have a max height and if the first option is expanded by default. I'm worried the user might completely overlook anything after the first row of tiles if it's below the fold.

src/components/services-view/services-view.controller.ts
dtaylor113: When I remove the delay, every 3 or 4th click on a sub-cat-card causes the expansion row's top to not be set, so it defaults back to CSS and appears at top of page:

image

The calculations are being done and the .css call is being made, but not set in the final HTML?

dtaylor113:
@spadgett , me too!   Per Serena's request I made the max-height of the expansion row to accommodate three rows of catalog items.  I'm concerned that pushes 2nd row of sub-cat-cards off screen and maybe missed by user.

src/components/services-view/services-view.controller.ts
dtaylor113: That worked great!  Fixed.

dtaylor113:
I know it's not as stylish but I'd almost want to have the sub-cat-cards just wrap as default/expected -no gap between rows.   The catalog item's area is always underneath all sub-categories, and the active card has a different 'selected' border.  This way we could avoid all these crazy calculations and users would always see all the sub-category tiles.

rhamilto:
@dtaylor113, mind rebasing this branch on master so that it incorporates @jeff-phillips-18's navbar changes?  @cshinn and I are working from this branch since it contains the latest cards bits.

src/components/services-view/services-view.controller.ts
rhamilto: Given the [mobile design for the cards is an accordion](http://openshift.github.io/openshift-origin-design/web-console/1-homepage/project-panel#responsive-behavior-mobile-view), shouldn't this styling be made specific to non-mobile viewport sizes?

dtaylor113:
Ok, rebased to latest.

src/components/services-view/services-view.controller.ts
dtaylor113: @rhamilto, yes it should be modified to adhere to the latest designs.  This is a WIP.  The goal of this PR is to get it closer to what we want and to remove the hard-coded rows and columns it previously had.

spadgett:
Hey @dtaylor113 I had a change from before that was able to do something similar with just HTML and CSS. Can we take a look at that as well together? The one limitation is that it set a fixed height on the expanded content.

I'd need to rebase on my branch on the latest master.

dtaylor113:
> Hey @dtaylor113 I had a change from before that was able to do something similar with just HTML and CSS. Can we take a look at that as well together?

Sure, that would be great

dtaylor113:
Incorporated @spadgett changes (thanks @spadgett !).  Hid 'Other' sub-category tile when viewing 'Other' category.

spadgett:
Thanks @dtaylor113 !

spadgett:
Code LGTM, but travis is complaining

PR #42, opened 2017-03-27
Change App to use Horizontal Navigation

jeff-phillips-18:
Clean up unused classes and icons
Minor adjustment to landing page top margin for better alignment

@rhamilto


dist/origin-web-catalogs.css
```diff
@@ -76,7 +76,7 @@
   background-color: #ededed;
   border-left: 1px solid #bbb;
   padding: 0 15px 0 10px;
-  margin: 10px 15px 10px 0;
+  margin: 0px 15px 10px 0;
```
spadgett: Nit: zero values don't need a unit

```css
margin: 0 15px 10px 0
```

jeff-phillips-18: Yeah but deleting 2 chars is twice as much work as deleting 1 😉

Issue #45, opened 2017-03-29
Need to truncate long text in SaaS Offering Tiles


dtaylor113:
Need to use `<truncate-long-text>` in web-catalog's SaaS Offering tiles:

![image](https://cloud.githubusercontent.com/assets/12733153/24458114/fae42e8a-1465-11e7-8d03-20a12741b1c8.png)

jeff-phillips-18:
This directive was already added to origin-web-common:

https://github.com/openshift/origin-web-common/pull/21

PR #46, opened 2017-03-29
Added mobile responsiveness to SaaS Offerings component

dtaylor113:
- Refactored 'rh-apps-list' to 'saas-list'
- Added 'title' binding to 'saas-list' component
- Added mobile responsiveness
- Added unit tests

Per [latest designs](https://github.com/openshift/openshift-origin-design/blob/master/web-console/1-homepage/offerings.md#responsive-designs)

Created issue #45 

image


src/components/saas-list/saas-list.component.ts
```diff
@@ -0,0 +1,10 @@
+import {SaasListController} from './saas-list.controller';
+
+export const saasList: angular.IComponentOptions = {
+  bindings: {
+    title: '<?',
```
spadgett: I think I like it better when title was outside the component, particularly if the page embedding might change it. Any compelling reason to move it here?

src/styles/saas-list.less
```diff
@@ -14,21 +14,37 @@
...
+    @media (min-width: @screen-sm-min) {
+      margin: 2px 0 10px 0;
+    }
   }
 
   .card-view-pf .card {
-    height: 200px;
-    width: 260px;
+    height: 100%;
+    width: 100%;
```
spadgett: Do the patternfly defaults for height/width/margin not work well? It seems odd that we have to set this.

src/styles/saas-list.less
```diff
@@ -14,21 +14,37 @@
...
+    }
   }
 
   .card-container {
     padding: 6px;
 
     .card-header {
       display: table;
-      height: 70px;
+      height: 100%;
```
spadgett: Is height: 100% necessary?

src/styles/saas-list.less
```diff
@@ -48,12 +64,21 @@
...
+        display: table-cell;
+        height: 80px;
+      }
+
       .card-description {
         font-size: 14px;
+        display: none;
+
+        @media (min-width: @screen-sm-min) {
+          display: inherit;
```
spadgett: Maybe just display: block?

src/components/saas-list/saas-list.controller.ts
```diff
@@ -11,21 +11,19 @@ export class RHListController implements angular.IController {
...
     window.open(item.url, '_blank');
   };
 
   public $onChanges(onChangesObj: angular.IOnChangesObject) {
-    if ((onChangesObj.applications && !onChangesObj.applications.isFirstChange()) ) {
-      this.ctrl.applications = onChangesObj.applications.currentValue;
-      this.ctrl.loading = false;
+    if ((onChangesObj.saasOfferings && !onChangesObj.saasOfferings.isFirstChange()) ) {
+      this.ctrl.saasOfferings = onChangesObj.saasOfferings.currentValue;
     }
```
spadgett: Doesn't this trigger an infinite loop? This should be unnecessary.

src/components/saas-list/saas-list.component.ts
dtaylor113: If there are no SaaS Offerings defined, then the whole component should be hidden -title and all.  If title were separate, then additional code would be needed to hide the title as well as the component.

src/components/saas-list/saas-list.controller.ts
dtaylor113: $onChanges is triggered when the parent changes the value of a one-way binding.  $onChanges doesn't get called when the variable is changed from within the component.   $onChanges probably isn't needed at this point since we're setting the saas offerings synchronously from the constants file.  I left it in because there is talk that in the future they might be loaded via an asynchronous API.

src/styles/saas-list.less
```diff
@@ -48,12 +64,21 @@
       }
     }
     .card-footer {
-      display: table-cell;
```
jeff-phillips-18: You could avoid the media settings here and just add 'hidden-xs" to the card footer html element.

src/styles/saas-list.less
dtaylor113: I removed them and the cards remained their original 'large' height and width.  I then set them to 'inherit' and the height was ok, but the width fit the width of the content and didn't stretch to the end.

src/components/saas-list/saas-list.component.ts
spadgett: OK

src/components/saas-list/saas-list.controller.ts
spadgett: OK, looks fine. Still learning components :)

src/styles/saas-list.less
spadgett: OK

src/styles/saas-list.less
```diff
@@ -14,21 +14,37 @@
   .card-view-pf .card-content {
     background-color: @color-pf-white;
     cursor: pointer;
+    margin: 0;
+
+    @media (min-width: @screen-sm-min) {
+      margin: 2px 0 10px 0;
```
dtaylor113: Fixed

src/styles/saas-list.less
dtaylor113: Fixed

src/styles/saas-list.less
dtaylor113: Fixed

app/pages/home/home-page.html
```diff
@@ -1,8 +1,7 @@
 <landing-page class="showcase-home-page" search-placeholder="Search Catalog">
   <landingheader>
     <div class="build-applications-view">
-      <h1>What Do You Want to Build?</h1>
-      <rh-apps-list applications="$ctrl.applications"></rh-apps-list>
+      <saas-list title="$ctrl.saasOfferingsTitle" saas-offerings="$ctrl.saasOfferings"></saas-list>
```
dtaylor113: Changed to 'saasTitle'

src/styles/saas-list.less
dtaylor113: Thanks, fixed

PR #47, opened 2017-03-29
Cleanup of layout to make <landing-page> less contextually specific

rhamilto:
changes include:

* switches layout to use flex (eliminates fixed positioning since that
could be problematic when <landing-page> is put in other contexts (like
origin-web-console))
* moves app-specific styling to app.less since landing-page.less should
be contextually agnostic (the app will need to control “advanced”
layout features like inner scrolling)
* the panel now slides down in front of the overlay

known issues:

* the layout of the content on the landing page is a bit janky, but
@cshinn is working on that in parallel to this effort
* the snap-up icon isn’t fully visible on clicking snap-down, but I
think the entire mechanism needs rethinking now that @spadgett changed
the layout the cards/tiles


src/styles/animations.less
```diff
@@ -0,0 +1,38 @@
...
+    margin-top: -50%;
+  }
+  100% {
+    margin-top: 0%;
+    opacity: 1;
+  }
+}
+
+@keyframes modalSlideDown {
+  0% {
```
dtaylor113: Do we need all four of these annimations?  I'm not sure the 'simple' animations are being used.
I see in overlay-panel.less:
```
.catalogs-overlay-window {
  &.ng-enter {
    animation: simpleSlideDown 1s ease forwards;
  }
  &.ng-leave {
    animation: simpleSlideUp 1s ease forwards;
  }
}
```
However, I don't see the class 'catalogs-overlay-window' used anywhere, so we maybe able to get rid of the 'simple' animations.

rhamilto: Good catch, @dtaylor113!  I've removed the animations and the rules as they weren't in use.

jeff-phillips-18:
The design of the landing page is for the search component and the side bar to be fixed. @serenamarie125 @beanh66

jeff-phillips-18:
Shouldn't there be a left margin on the page? The text runs right up against the window edge.

jeff-phillips-18:
The edit and create project panels no longer show the panel titles or field labels.

jeff-phillips-18:
The edit and create project panels are fixed but the sidebar isn't. If the user scrolls, the panel does not move with the target it points to.

rhamilto:
> The design of the landing page is for the search component and the side bar to be fixed. @serenamarie125 @beanh66

@jeff-phillips-18, that position fixing needs to be done at the app level, which I can add back to this demo app, but it will need to be addressed in each context (app) that utilizes <landing-page> since position: fixed is relative to the viewport, and we won't know what's at the top, right, left, and bottom of the viewport.  For example, when <landing-page> gets put in to origin-web-console, the offset properties are going to vary from what's in this demo app since the navbar height is going to differ (and is even customized by customers in some instances).

rhamilto:
> Shouldn't there be a left margin on the page? The text runs right up against the window edge.

We'll fix that when we style the content.  @cshinn is working on the apps section in parallel to this effort.

rhamilto:
> The edit and create project panels no longer show the panel titles or field labels.

Sorry.  I meant to document that as a known issue.  That'll get fixed when I style the sidebar stuff (which is blocked by these layout change needs).

spadgett:
> The design of the landing page is for the search component and the side bar to be fixed. 

What happens if the content of the sidebar is too long to fit on the screen?

spadgett:
> What happens if the content of the sidebar is too long to fit on the screen?

To be clear, I'm not saying it shouldn't be fixed, but there are some problems we'll have to solve like different banner heights in different contexts (as @rhamilto points out) and content that overflows.

rhamilto:
> The edit and create project panels are fixed but the sidebar isn't. If the user scrolls, the panel does not move with the target it points to.

Missed that one since I wasn't focused on the contents.  Lemme fix.

rhamilto:
>> The edit and create project panels are fixed but the sidebar isn't. If the user scrolls, the panel does not move with the target it points to.

> Missed that one since I wasn't focused on the contents. Lemme fix.

Done.

rhamilto:
Ok.  I believe I've fixed all the noted issues.  PTAnotherL.

src/styles/landing-page.less
```diff
@@ -1,88 +1,34 @@
...
-
 .landing-main-area {
-  margin: 15px 20px;
-  padding-right: 10px;
-
-  // TODO: Remove corresponding style in services-view.less (card-expansion)
-  // when sidebar is no longer fixed position.
-  @media (min-width: @screen-sm-min) {
-    margin-right: @sidebar-width;
+  padding: (@grid-gutter-width / 2); // this is temp
```
rhamilto: @cshinn, you're gonna want to remove this to get the styling of the search and apps area right, but I ended up putting it in to make the page look ok in the interim.

cshinn: Thanks for the heads up!

PR #48, opened 2017-03-29
Addresses Issue #44 'Catalog showing image streams that aren't builder images'

### LINKED ISSUE #44: Catalog showing image streams that aren't builder images
We should only be showing image streams have are tagged as builders. See

https://github.com/openshift/origin-web-console/blob/master/app/scripts/services/catalog.js#L91-L94

cc @jwforres

dtaylor113:
Fixes #44 
@spadgett


src/services/catalog.service.ts
```diff
@@ -158,6 +160,7 @@ export class ImageItem implements IServiceItem {
     this.longDescription = this.getLongDescription();
     this.catsBySubCats = this.getCategoriesBySubCategories();
     this.isBuilder = _.includes(this.tags, 'builder');
```
spadgett: We need to check that there is a corresponding tag in `status.tags` that matches the builder spec tag.

Image streams have `spec.tags` that are the tag definitions and `status.tags` that are the tags which are currently loaded / exist.

spadgett: OK to do this as a follow on.

dtaylor113: Ok, created issue #52

dtaylor113: So, I'm using 
`this.tags = this.catalogSrv.$filter('imageStreamTagTags')(this.resource);`
Since I'm not specifing a tagName, it is using 'latest' as tagName.  Are you saying I need to verify that 'latest' exists in status.tags?  ...and if it doesn't, do not use the image?

spadgett: The latest tag references another tag. Tags can points to other tags in the same image stream or another image stream. We combine these in the version dropdown currently.

spadgett: https://github.com/openshift/origin-web-console/blob/master/app/scripts/services/catalog.js#L151-L160

The logic around all this is complicated unfortunately.

spadgett: You also need to check all the spec tags for builders, not just latest.

spadgett:
@dtaylor113 Please open a follow on PR to check all spec tags for `builder`. Not all image streams have a `latest` tag (or a `latest` that's tagged as a builder), so it's possible for things to be hidden entirely that shouldn't.

dtaylor113:
@spadgett I added issue #52 yesterday, and just updated it with your comment above.   I need to meet up with you at some point to understand what is required.  Especially what I need to do with 'status.tags'.  -thanks!

spadgett:
@dtaylor113 Sure, let me add some comments in the issue explaining.

PR #50, opened 2017-03-29
Align order service wizard with design

spadgett:
In this PR:

- [x] Move steps to top of wizard
- [x] Add wizard buttons fixed to bottom for all steps
- [x] Use a radio button for selecting plans
- [x] Use Patternfly wizard styles

Follow-on work:

- [ ] Mobile styles
- [ ] Truncate descriptions
- [ ] Show plan bullets
- [ ] Add tags to dialog #54

screen shot 2017-03-29 at 2 19 42 pm


spadgett:
Note that more work is still needed for mobile.

spadgett:
@serenamarie125 

It's currently

> Plans -> Configure -> Results

Should it be...?

> Plans -> Configuration -> Results

Then it's noun - noun - noun instead of noun - verb - noun

dtaylor113:
In the 'Select a Plan' step, if there are a lot of plans, where does it scroll?  Does the 'Select a Plan' title remain in place but the radio-buttons scroll?

spadgett:
"Select a Plan" scrolls together with the plans right now.

spadgett:
Rebased and cleaned up step includes based on feedback from @jwforres

serenamarie125:
@spadgett you are correct, it should be Configuration.  Thanks for pointing that out!  I'll update the associated design documentation as well

serenamarie125:
@spadgett I don't see tags in your mocks.  Are you going to try to add those as well?  We have a small change for that, I'm going to update design docs tomorrow.  This is more inline with how tags are currently shown, and consistent with the way we are showing it in search results.

screen shot 2017-03-30 at 12 17 46 am

spadgett:
@serenamarie125 Thanks, I'll change to "Configuration".

I haven't done the tags yet. If you don't mind, I'll open an follow-up issue and look at it next since @jwforres has already reviewed this pull request and I think it gets us closer.

If we keep the "BUILDS SOURCE CODE" text for builders, we should look at the style since it's similar to the proposed tags style.

Just a note that each builder image version can have its own tags and descriptions, so the values might change when you select a different version from the dropdown. I assume the left panel should simply update? (This is only for source code builders and not service classes from the service catalog.)

spadgett:
@jeff-phillips-18 PTAL

- Updated the results steps buttons to match design
- Added correctly Patternfly button classes to all wizard buttons
- Switched step name to "Configuration"

screen shot 2017-03-30 at 7 41 01 am

spadgett:
Opened #54 for adding tags.

src/components/order-service/order-service-plans.html
```diff
@@ -1,21 +1,20 @@
...
+    <div ng-repeat="plan in $ctrl.serviceClass.resource.plans" class="radio">
+      <label>
+        <input
+          type="radio"
+          ng-model="$ctrl.planIndex"
+          ng-change="$ctrl.selectPlan(plan)"
+          value="{{$index}}">
+        <span class="plan-name">{{plan.osbMetadata.displayName || plan.name}}</span>
+        <!-- TODO: truncate long text -->
+        <div ng-if="plan.description">{{plan.description}}</div>
```
jeff-phillips-18: origin-web-common 0.0.8 has this directive available

spadgett: > origin-web-common 0.0.8 has this directive available

Thanks, will add this in a follow on.

Issue #32, opened 2017-03-24
Catalog showcase goes into redirect loop with invalid config


spadgett:
It's very easy to get into a redirect loop if the catalog config is incorrect. We should improve our error handling.

jeff-phillips-18:
One problem is that the additional servers are treated like the necessary servers. It is easy to not have the additionalServers setup correctly (since they may need to be created). Should we continue on if we fail  to connect to the additionalServers in the config?

@spadgett @jwforres

spadgett:
@jeff-phillips-18 I think it's OK if it fails, but it shouldn't keep reloading the page again and again. If nothing else, it makes it difficult to troubleshoot the problem. I'm seeing what @benjaminapetersen saw in #14.

We should also change the default config.js. The additional servers [1] won't likely work for anyone. Maybe use mock data by default for now? There is some set up to show real service classes.

[1] https://github.com/openshift/origin-web-catalog/blob/master/app/config.js#L24-L27

benjaminapetersen:
Yes, definitely.  Ideally fail once & give a good error message?  Looping is a pain, console gets cleared so there is no good way to trouble shoot (other than pester other people who might know the cause).

jeff-phillips-18:
@benjaminapetersen  Can you try this again with the latest commits? There have been a few changes that may have resolved this as I don't see the issue anymore.

spadgett:
@jeff-phillips-18 I saw this yesterday morning and again this morning with a valid config. Both times stopping and restarting `npm run start` fixed it. I had to accept the localhost:9001 cert again as well, which might be related.

jeff-phillips-18:
I see that issue (typically every morning) but it goes to the error page correctly without looping. Eventually it seems the service catalog server responds and it works (no npm restarting). Same issue in the console.

dtaylor113:
This is the behavior I am seeing as well.  I haven't seen the looping since we added the error handling page/route.

benjaminapetersen:
Haven't tried it again yet but will take a look.

jwforres:
I still see this problem, at least from the main console, if the service catalog cert hasn't been accepted. I believe we had discussed handling errors on additionalServers loading differently than we handle the main API discovery.  Since additionalServers is supposed to always be for prototyping purposes.  But any change to that will be in common in the api discovery code.

benjaminapetersen:
Yup. still kinda fiddly to get all setup.

spadgett:
[...]

PR #53, opened 2017-03-29
Comment out additional-servers block, leave as an example

jeff-phillips-18:
Updated this to work on #32 but I don't see that login loop happening anymore. This fix is still relevant though.

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -21,17 +21,23 @@ window.OPENSHIFT_CONFIG = {
       prefix: "/api"
     }
   },
-  additionalServers: [{
-    protocol: "https",
-    hostPort: "apiserver-svc-cat.10.0.0.25.xip.io:6443",
-    prefix: "/apis"
-  }],
   auth: {
     oauth_authorize_uri: "https://localhost:8443/oauth/authorize",
     oauth_redirect_base: "https://localhost:" + window.DEV_SERVER_PORT,
     oauth_client_id: "openshift-web-console",
     logout_uri: ""
   },
+/*
+ * Add Any additional servers to an 'additionalServers' block, example shown below:
+ *
+   additionalServers: [{
+    protocol: "https",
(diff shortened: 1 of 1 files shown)


app/config.js
```diff
@@ -23,7 +23,7 @@ window.OPENSHIFT_CONFIG = {
   },
   additionalServers: [{
     protocol: "https",
-    hostPort: "apiserver-svc-cat.10.0.0.25.xip.io:6443",
+    hostPort: "apiserver-svc-cat.10.0.0.25.xip.io",
```
spadgett: This value still won't work for most people unfortunately. If you're running cluster up locally on your laptop, it will be 127.0.0.1. I assume you're running cluster up in vagrant? It also relies on you having deployed the template, and eventually it won't be required.

We might just comment the stanza out in the default config.js and explain above how it's used. Perhaps default to mock data.

jeff-phillips-18: we can comment it out. I was hoping to be able to ignore the errors by adding a flag to an additional server block but that PR (https://github.com/openshift/origin-web-common/pull/22) is still in limbo.

spadgett: Yeah I'm undecided on that. I think I'd want to see the connection error to know something isn't working?

jeff-phillips-18: I don't believe there is a good way to see the error. Same as in console. The discovery is done during the angular bootstrap process and if it fails the page is redirected so any failure message is not retrievable.

spadgett: Ah I see. Approved https://github.com/openshift/origin-web-common/pull/22#pullrequestreview-29994970

spadgett:
I guess this is better except it will still only work if

- You ran cluster up inside of a VirtualBox VM
- You deployed the catalog template in a project named `svc-cat`

It seems very specific to one environment to leave as a default.

jeff-phillips-18:
Updated port number and commented the additional servers block out.

PR #55, opened 2017-03-30
Bump origin-web-common to 0.0.8

spadgett:


jeff-phillips-18:
Need to bump the version in package.json as well.

It is in bower.json so that the dependency is known by those installing with bower but in order to use it in catalogs and not include it in vendor.js it has to be used from the npm install.

spadgett:
Thanks, I updated package.json now too, but I don't see any changes in the dist folder after reinstalling / rebuilding. Did I miss something?

jeff-phillips-18:
No, that change shouldn't effect it.

spadgett:
@jeff-phillips-18 Spoke with @jwforres and she was going to update the panel styles for the large panel. Then I think we should cut a release with everything and update both origin-web-catalog and origin-web-common in the console repo at the same time to avoid bower conflicts.

Issue #29, opened 2017-03-23
Need to review subcategory matching logic


dtaylor113:
The question came up whether a service item (service-class or image-stream)'s tag can match multiple sub-categories. 

Currently the logic in the catalog service's getCategoriesBySubCategories() finds the first sub-category which matches a given tag.  A single tag like 'nodejs' is going to match a single sub-category (or alias). A service item may have multiple tags, which may match to multiple sub-categories and categories. We don't have a mapping of matching a single tag to multiple sub-categories.

"Potentially the icon, description, and tags can be different for each spec tag"

This may come back to my initial comment: "I still need to reconcile [catalog categories](https://github.com/openshift/origin-web-catalog/blob/master/src/constants.ts#L12) with [consoles categories](https://github.com/openshift/origin-web-console/blob/master/app/scripts/constants.js#L301).  Need to deep-dive with someone and determine what else needs to be added to the Catalog's categories."

I see in  [consoles categories](https://github.com/openshift/origin-web-console/blob/master/app/scripts/constants.js#L301) there is something called 'subcategories', which really makes that structure 3 levels deep:  Categories -> Items -> subcategories.  I am currently using a categories constant which is two levels deep.

PR #57, opened 2017-03-30
Fix for issue #52: Catalog Service View: Need to match 'status.tags'

### LINKED ISSUE #52: Catalog Service View: Need to match 'status.tags'
We need to check that there is a corresponding tag in status.tags that matches the builder spec tag.

Image streams have spec.tags that are the tag definitions and status.tags that are the tags which are currently loaded / exist.

Check all spec tags for builder. Not all image streams have a latest tag (or a latest that's tagged as a builder), so it's possible for things to be hidden entirely that shouldn't.

See PR #48

dtaylor113:
Fixes #52


src/services/catalog.service.ts
```diff
@@ -145,26 +145,56 @@ export class ImageItem implements IServiceItem {
   public description: string;
   public longDescription: string;
   public resource: any;
-  public isBuilder: boolean;
-  public isHidden: boolean;
+  private validSpecTagName: any;
```
spadgett: "Valid" seems like the wrong name. Non-builder tags are still valid. Maybe `builderSpecTagName`.

src/services/catalog.service.ts
```diff
@@ -145,26 +145,56 @@ export class ImageItem implements IServiceItem {
...
-    this.isBuilder = _.includes(this.tags, 'builder');
-    this.isHidden = _.includes(this.tags, 'hidden');
+    this.validSpecTagName = this.getValidSpecTagName();
+    if (this.validSpecTagName) {
+      this.tags = this.getTags();
+      this.iconClass = this.getIcon();
+      this.name = this.getName();
+      this.description = this.getDescription();
+      this.longDescription = this.getLongDescription();
+      this.catsBySubCats = this.getCategoriesBySubCategories();
```
spadgett: Potentially the icon, description, and tags can be different for each spec tag. I'm not sure the ImageItem model we've made is correct.

This is OK for the catalog tile itself, but the dialog will need to do something else to support selecting versions.

I suggest when possible we work with the API objects directly. Kubernetes and OpenShift have well-defined, well-documented APIs. As it stands, we have to backtrack through the code when looking at a view to see what a property really is. And things are getting lost in translation. Here we're dropped data (icons and descriptions for other tags).

src/services/catalog.service.ts
dtaylor113: Ok, changed to 'builderSpecTagName'

src/services/catalog.service.ts
dtaylor113: > I suggest when possible we work with the API objects directly

Hi @spadgett , I'm not sure what you are suggesting here?  Are you saying we shouldn't use the ServiceItem and ImageItem classes?

spadgett:
Thanks @dtaylor113. I think we can look at the other issue I raised as part of #29

src/services/catalog.service.ts
spadgett: I'm making a general statement that we should try to use API objects directly when possible. This is the approach we took in origin-web-console, and it worked well.

ServiceItem and ImageItem makes sense here since we're showing different types as catalog tiles with a generic interface. But I think when we pass the objects to the wizard, we should pass the API object instead. We'll probably need different components anyway for the two flows (builder and service class) since they're so different, so the abstraction is less helpful there.

dtaylor113: Ok, I was thinking the same.  IServiceItem.resource (the API object) should be passed to the ordering wizard.

dtaylor113:
Hi @spadgett , waiting on your approval in order to merge.  -thanks!

src/services/catalog.service.ts
```diff
@@ -145,26 +145,52 @@ export class ImageItem implements IServiceItem {
...
+    let validSpecTag: any;
+
+    if (!this.resource.status) {
+      return null;
+    }
+
+    if (this.resource.spec && this.resource.spec.tags) {
+      validSpecTag = _.find(this.resource.spec.tags, (specTag: any) => {
+        let specTagTags: any = _.get(specTag, 'annotations.tags');
+        if (specTagTags && _.includes(specTagTags, 'builder') && !_.includes(specTagTags, 'hidden')) {
```
spadgett: I just noticed you're doing a substring match, which isn't technically right. It would also match tag `notabuilder`.

Really we should be splitting the tags string, e.g. `specTagTags.split(/\s*,\s*/)`, and checking if builder and hidden are in the array that's returned.

dtaylor113: Fixed.  I had that in there earlier, think I accidentally removed it when I cleaned up debugging calls

PR #59, opened 2017-03-30
Show service class tags in order service wizard

### LINKED ISSUE #54: Order wizard should show service class tags
See https://github.com/openshift/origin-web-catalog/pull/50#issuecomment-290298152

cc @serenamarie125

spadgett:
screen shot 2017-03-30 at 12 58 45 pm

Fixes #54

cc @serenamarie125


src/styles/order-service.less
```diff
@@ -21,24 +23,32 @@
   margin-top: 20px;
 
   .order-service-deails-top {
-    height: 70%;
-    .service-title {
-      display: flex;
-      align-items: center;
-      .icon {
+    .icon {
```
jwforres: i noticed this earlier but didnt want to touch it, there is a typo in class name order-service-details-top

src/styles/order-service.less
```diff
@@ -52,17 +62,13 @@
...
-      color: @color-pf-black;
-      background-color: @color-pf-white;
-      border-color: @color-pf-blue-300;
-    }
+  .order-service-description {
+    clear: both;
+    margin-top: 15px;
+  }
+  .tag {
+    margin-right: 5px;
```
jwforres: should this be on a .tag + .tag selector as a margin-left so we dont have a hanging 5px margin on the last one

src/styles/order-service.less
spadgett: I can fix since I'm chaning it anyway

src/styles/order-service.less
spadgett: But if the tags wrap then you will have extra left margin on the second row

spadgett: I could add negative margin to the block and then left margin to all tags

jwforres: good point

jwforres: um i would say leave it as is unless we find its causing an issue, lets not use negative margin unless we really need to

spadgett: I have it working and it seems to work well, but I can take it back out. FWIW I think Bootstrap uses the same technique in its grid

jwforres:
I'll let you decide

PR #61, opened 2017-03-30
First pass of sidebar styles

rhamilto:
screen shot 2017-03-31 at 1 54 18 pm
screen shot 2017-03-31 at 1 58 43 pm
screen shot 2017-03-31 at 2 02 22 pm
screen shot 2017-03-31 at 1 54 40 pm
screen shot 2017-03-31 at 1 54 52 pm


spadgett:
Not new or for this PR, but we should probably remove `class="input-lg"` from the inputs in the create/edit project dialogs. Unfortunately this is set in the template in the common repo and used by origin-web-console for its standalone pages.

dtaylor113:
LGTM except the bright blue text.  That seems really bright to me, like I might get a headache if I had to work with it daily.

src/styles/projects-summary.less
```diff
@@ -1,156 +1,200 @@
...
+
+    .btn-link {
+      color: @color-pf-white;
+      margin-left: 0;
+    }
+
+    .dropdown-menu {
+      &:after {
+        border: 7px solid transparent;
+        border-bottom-color: #fff;
```
jeff-phillips-18: @color-pf-white

src/styles/projects-summary.less
```diff
@@ -1,156 +1,200 @@
...
     content: '*';
     position: absolute;
-    left: 5px;
+    left: 12px;
   }
 }
 
 .catalog-project-summary-list {
-  margin-top: 20px;
+  margin-bottom: 40px;
```
jeff-phillips-18: @grid-gutter-width

rhamilto:
> LGTM except the bright blue text. That seems really bright to me, like I might get a headache if I had to work with it daily.

^ @jennyhaines, @serenamarie125 

> LGTM, just a couple nits.

Good catches, @jeff-phillips-18.  Thanks!

serenamarie125:
Looking good !  Couple questions for ya @rhamilto 
- Once project name is added, does the Create button turns blue?
- What are you doing for hover state for links?  (View All & Project Names)

serenamarie125:
In reference to the headache comment @dtaylor113 the blue text is only used for "View All" and maybe up to 5 project names.  I agree if it was everywhere on the page, it may be overwhelming but when in context looking at the entire page, I don't find it to be a problem.  Thoughts @rhamilto @beanh66 ??

rhamilto:
> Looking good ! Couple questions for ya @rhamilto

Thanks!

> * Once project name is added, does the Create button turns blue?

Yep.  That's just the empty state of the form.

> What are you doing for hover state for links? (View All & Project Names)

The default PatternFly hover for links (underline and darker blue).  This should probably be tweaked.

screen shot 2017-03-31 at 10 26 20 am

> In reference to the headache comment @dtaylor113 the blue text is only used for "View All" and maybe up to 5 project names. I agree if it was everywhere on the page, it may be overwhelming but when in context looking at the entire page, I don't find it to be a problem. Thoughts @rhamilto @beanh66 ??

I don't think it would hurt to lighten the blue in order to increase the contrast.

spadgett:
At some point it's an accessibility issue if there's not enough contrast

jennyhaines:
@rhamilto - in the PSD all of the blue links are (or at least should be!) #00a4e4.  This pass looks like a darker blue than the links in the PSD. Changing to the PSD color would help with this contrast issue and might help with the headache issue a bit. Also, on hover, how about rather than darkening the links and underlining, we just lighten the links to #48c6f6 ?

serenamarie125:
@jennyhaines typically we are trying to reference colors from the PF color palette https://www.patternfly.org/styles/color-palette/ those values don't seem to be there ?

serenamarie125:
@rhamilto is it possible to have the title and "x" vertically middle aligned on the Create Project & Edit Project overlays?

rhamilto:
> @rhamilto is it possible to have the title and "x" vertically middle aligned on the Create Project & Edit Project overlays?

It is, but I intentionally didn't do that as the "x" gets a bit lost when the title is long.  To me it makes the "x" feel like part of the title and not a control for the "window".  Agree/disagree?

screen shot 2017-03-31 at 11 18 00 am

rhamilto:
@serenamarie125, @beanh66 Something else I just noticed:  in the console, the "+ Create Project" button is labeled "New Project".  Presumably these should have the same label since they perform the same function?  I don't have an opinion which one is better, but I do think they should be consistent.

screen shot 2017-03-31 at 11 21 29 am

beanh66:
@rhamilto +1 to the comment @serenamarie125 made regarding the create project title and the x being aligned. Can we just have the title be "Create Project" or "Edit Project" rather than appending the project name? I would also add that we should have a light grey (#f5f5f5) background behind that top title bar, similar to what we do for the guided tour. Is that possible?

beanh66:
@rhamilto Agreed! In the design documentation (http://openshift.github.io/openshift-origin-design/web-console/2-projects-page/overview) I have two changes to that full projects page: (1) Change "Projects" to say "My Projects" and (2) change "New Project" button to say " + Create Project" so we are consistent with both. Would that be you making those changes or someone else?

spadgett:
FWIW, we chose "New Project" to match what's used by the CLI, `oc new-project`

serenamarie125:
@spadgett from a UX/PF perspective, we like to have actions include a verb, thus the "+ Create"

jennyhaines:
@serenamarie125 @rhamilto -

So that we are both PF compliant and color contrast compliant with the link color, it seems we have two options:

We continue to use the same PF blue link color, but change the panel background color to #131B1F
(this noticeably darkens the look of the the panel, maybe a bit too much)
option 1
-OR-
we use the slightly lighter PF Blue in our palette (pf-blue-300 #39a5dc), over a #2C343D panel
(I prefer the look of this option)
option 2

Thoughts?

beanh66:
@jennyhaines I like option 2 as well!

serenamarie125:
Option 2 looks great and contrast looks good as well.  Thanks @jennyhaines !
@rhamilto do you agree?  Could you make those changes?

rhamilto:
[Visual tweaks incorporated](https://github.com/openshift/origin-web-catalog/pull/61#issue-218300449).  I think this one is ready for merge.

rhamilto:
> @rhamilto Agreed! In the design documentation (http://openshift.github.io/openshift-origin-design/web-console/2-projects-page/overview) I have two changes to that full projects page: (1) Change "Projects" to say "My Projects" and (2) change "New Project" button to say " + Create Project" so we are consistent with both. Would that be you making those changes or someone else?

That could be anyone on the openshift UI team.  @jwforres, @spadgett:  You good with me opening an issue in origin-web-console for this?

spadgett:
@rhamilto If we change the new project button, don't forget to change the header on the new project form.

screen shot 2017-03-31 at 2 42 09 pm

I'm fine with the change. My only hesitation is that we don't use `+` on the buttons anywhere else in the web console :/

PR #65, opened 2017-03-31
Add visual styles for saas offerings

cshinn:
screen shot 2017-04-04 at 4 18 52 pm
screen shot 2017-04-04 at 4 19 23 pm


dtaylor113:
Travis failure.  Think you need to run 'npm run build' and then push the dist files.

dtaylor113:
When not hovering over a SaaS tile I cannot see a border for the tiles, or a change in color which separates the tiles from the background.  Not sure if that is intentional or not.

cshinn:
@dtaylor113 the hi-fi visuals do not include a border when the cards are not hovered.

src/styles/variables.less
```diff
@@ -4,3 +4,6 @@
 
 // copied from origin-web-console
 @sidebar-os-bg: #30363d;
+
+@saas-text: @color-pf-black-800; //to be replaced with @color-pf-white when the background image is in place
+@saas-hover: fade(#09405e, 80%);
```
jeff-phillips-18: We should stick to the Patternfly color palette. @color-pf-blue-600 is pretty close.

src/styles/saas-list.less
```diff
@@ -1,53 +1,93 @@
...
+    flex-direction: column;
+    padding: 0;
+    margin: 0 (@grid-gutter-width / 2);
 
-  .card-view-pf .card-content {
-    background-color: @color-pf-white;
-    cursor: pointer;
-    margin: 0;
+    .card-content {
+      cursor: pointer;
```
spadgett: This is often the sign of an accessibility problem. Not something for you to change @cshinn, but if the cards aren't links, you won't be able to focus them with a keyboard or screenreader. We should look at this as a follow on.

src/styles/saas-list.less
```diff
@@ -1,53 +1,93 @@
...
-    text-align: inherit;
-    margin-bottom: 8px;
-
-    @media (min-width: @screen-sm-min) {
-      height: 200px;
-      width: 260px;
-      margin-bottom: 20px;
+    .card {
+      background: none;
+      box-shadow: none;
```
spadgett: It doesn't feel like we should need to set this. Is this coming from `card-view-pf`? Can we just leave that class off since they're no longer styled as cards?

src/styles/saas-list.less
```diff
@@ -1,53 +1,93 @@
...
+      margin: 0;
+
+      .card-content {
+        background: none;
+        text-align: center;
+        cursor: pointer;
+      }
+
+      .card {
+        border: none;
```
spadgett: Same question here: should we just remove the `card-view-pf` class and style the cards ourselves?

src/styles/saas-list.less
```diff
@@ -56,21 +96,45 @@
...
+        display: block;
+        text-align: center;
+        margin-bottom: 10px;
+
+        .card-icon {
+          font-size: 108px;
+          width: 100%;
+          margin-bottom: 20px;
+
+          &.fa.fa-mobile {
```
spadgett: We shouldn't be styling specific icon classes as these are customizable.

cshinn: @spadgett will these be font icons at all? I was under the impression that they would probably be images at some point in the future.

src/styles/saas-list.less
cshinn: Definitely, I'll get rid of that.

src/styles/saas-list.less
spadgett: I'm not sure. We generally try to use font icons when we can.

cc @jwforres

serenamarie125: @spadgett @cshinn I can't imagine that these will be font icons.  My guess is that it will be some custom icons or product imagery, but I will check on our side.  I think we will document that we want them to be monochrome though so that the color can still be changed?

rhamilto:
@cshinn, we try to follow [PatternFly's code guide](http://codeguide.patternfly.org/) for html and css.  Of importance here is [css declaration order](http://codeguide.patternfly.org/#css-declaration-order).  Mind making those adjustments and squashing?  It's much easier to review all the changes at once rather than multiple commits?

src/styles/saas-list.less
jwforres: Yep, my suggestion would be that the name on the tile be an `<a>` tag.  Then if we want to style it to not look like an `<a>` we can. I'm opening https://github.com/openshift/origin-web-catalog/issues/67 since this is an issue with more than just the SaaS tiles.

src/styles/saas-list.less
rhamilto: My recommendation is to use SVG images as I believe the icons in the mocks are part of the [standard brand icons](http://brand.redhat.com/elements/icons/), which are in SVG format, but not available as a font (which is really no longer needed for our purposes since our supported browsers support SVG).

cshinn:
@spadgett I've stripped out references to pf-card-view and re-styled. Much simpler now.

src/components/saas-list/saas-list.html
```diff
@@ -1,13 +1,11 @@
 <span ng-if="$ctrl.hasSaasOfferings()" class="saas-offerings-container">
   <h1 ng-if="$ctrl.saasTitle">{{$ctrl.saasTitle}}</h1>
-  <div pf-card-view config="$ctrl.cardViewConfig" items="$ctrl.saasOfferings">
-    <div class="card-container">
-      <div class="card-header">
+  <div class="saas-list" items="$ctrl.saasOfferings">
+    <div class="card" ng-repeat="item in $ctrl.saasOfferings" ng-class="{ 'last': $ctrl.saasOfferings.length == ($index+1)}">
```
spadgett:
```
ng-class="{ last: $last }"
```

spadgett:
We'll need to update the tests since the markup changed, cc @dtaylor113

spadgett:
Even with my browser fullscreen on a 15" laptop, the cards take the entire page.

@serenamarie125 How should this look at this screen width?

screen shot 2017-03-31 at 1 02 51 pm

cshinn:
@serenamarie125 @spadgett I've reduced the padding, icon size, and title size to get it to fit into a single row on my smaller screen, but without reducing the base font size we're looking at something like this. Seems a bit cramped to me.
screen shot 2017-03-31 at 1 17 03 pm 2

spadgett:
The number of items isn't fixed either. With three, the current stylings are fine. With five, even the smaller fonts and margins would wrap :/

src/styles/saas-list.less
```diff
@@ -1,76 +1,104 @@
...
-  .card-view-pf {
-    padding-left: 21px;
-  }
-
-  .card-view-pf .card-content {
-    background-color: @color-pf-white;
-    cursor: pointer;
-    margin: 0;
+  .saas-list {
+    display: flex;
```
rhamilto: Since you don't need flex to achieve your layout for mobile, it's a bit simpler to only do the flex-ing inside of the media query below (line 26).  That way, there is no need to set and reset flex-direction.  In other words, move lines 20, 22, 23 to line 27.  Delete line 21 and 27.

src/styles/saas-list.less
```diff
@@ -1,76 +1,104 @@
+.build-applications-view {
+
+  h1 {
+    color: @saas-text;
+    margin: 0;
+    padding: 20px;
+    padding-left: (@grid-gutter-width / 2);
```
rhamilto: No need to set padding-left to 20px a second time.  :)

src/styles/saas-list.less
```diff
@@ -1,76 +1,104 @@
+.build-applications-view {
+
+  h1 {
+    color: @saas-text;
+    margin: 0;
+    padding: 20px;
```
rhamilto: Use (@grid-gutter-width / 2) instead?

src/styles/saas-list.less
```diff
@@ -1,76 +1,104 @@
...
+    padding-left: (@grid-gutter-width / 2);
+    width: 100%;
```
rhamilto: Not necessary.  Block level elements assuming a width of 100% by default.

src/styles/saas-list.less
```diff
@@ -1,76 +1,104 @@
...
+        height: 300px;
+        padding: 30px;
+        width: 280px;
+      }
     }
-  }
 
-  .card-container {
-    padding: 6px;
+    .last {
```
rhamilto: Use &[:last-child](https://developer.mozilla.org/en-US/docs/Web/CSS/:last-child) on line 44 instead and remove the class="last" from the markup.

src/styles/saas-list.less
```diff
@@ -1,76 +1,104 @@
...
-  }
 
-  .card-view-pf .card {
-    height: 100%;
-    width: 100%;
-    cursor: pointer;
-    text-align: inherit;
-    margin-bottom: 8px;
+    .card {
+      align-items: center;
```
rhamilto: Not needed.

src/styles/saas-list.less
```diff
@@ -1,76 +1,104 @@
...
+      border-bottom: solid 1px @saas-text;
+      color: @saas-text;
+      cursor: pointer;
+      display: flex;
+      height: 5em;
+      width: 100%;
```
rhamilto: Not needed.

src/styles/saas-list.less
```diff
@@ -1,76 +1,104 @@
...
-      margin-bottom: 20px;
+
+      &:hover {
+        background-color: @saas-hover;
+      }
+
+      @media (min-width: @screen-sm-min) {
+        border-bottom: none;
+        flex: 0 1 auto;
+        height: 300px;
```
rhamilto: One of the benefits of flexbox is that siblings with automatically assuming the same height as one another.  So it's probably best if you don't set a fixed height.

src/styles/saas-list.less
```diff
@@ -1,76 +1,104 @@
...
+            font-size: 108px;
+          }
+        }
       }
+
       .card-title {
-        vertical-align: middle;
-        display: table-cell;
-        padding-left: 8px;
+        font-size: @font-size-base + 4;
```
rhamilto: Perhaps this should be scaled down for resolutions of maybe 480 and less so the text isn't so crowded?

screen shot 2017-03-31 at 3 24 17 pm

src/styles/saas-list.less
```diff
@@ -1,76 +1,104 @@
...
+        font-weight: bold;
```
rhamilto: Use a value of [700](https://github.com/patternfly/patternfly/blob/master/src/less/fonts.less#L80) instead of the keyword.

src/styles/saas-list.less
```diff
@@ -1,76 +1,104 @@
...
+      &:hover {
+        background-color: @saas-hover;
+      }
+
+      @media (min-width: @screen-sm-min) {
+        border-bottom: none;
+        flex: 0 1 auto;
+        height: 300px;
+        padding: 30px;
+        width: 280px;
```
rhamilto: This fixed width could be problematic when resizing the viewport.  It'd be better (and less fragile) if you made the width relative (percentage-based).

serenamarie125:
I like the updates @cshinn 
For Tech Preview, i think the max we will have is 3.  @spadgett are you saying it looks ok with < 5?

src/styles/saas-list.less
```diff
@@ -1,76 +1,104 @@
...
-    .card-header {
-      display: table;
+    .card-content {
+      align-items: center;
+      cursor: pointer;
+      display: flex;
 
       @media (min-width: @screen-sm-min) {
-        height: 70px;
+        display: block;
```
rhamilto: Use flex-direction: column instead so you keep things flexed.

cshinn:
@rhamilto I appreciate your taking the time to help me improve this stuff! I'll get right on it

src/styles/saas-list.less
dtaylor113: Hi @rhamilto, just fyi that I initially created this .less file which is why it probably needs some cleaning up :-)  @cshinn is not to blame!!! :-)

cshinn:
@spadgett I've updated based on @rhamilto 's feedback and removed the saas header per @dtaylor113 's instructions. Between the visual tweaks and the removal of the header and the scrolling buttons, the catalog section is now visible even when the saas items wrap to a second line.

cshinn:
@rhamilto @jwforres I've updated the PR to include the correct text color and added the background placeholder color

cshinn:
@rhamilto I've added in the background color as making your suggested changes

jeff-phillips-18:
@rhamilto Any further issues with this?

src/styles/variables.less
```diff
@@ -7,3 +7,6 @@
 @landing-side-bar-width-sm:                           325px;
 @landing-side-bar-width-md:                           375px;
 @landing-side-bar-width-lg:                           450px;
+@landing-main-bg: #3ca2c9;
```
rhamilto: It would be great if you alphabetized these.

rhamilto:
There are a few responsive display issues I think should be addressed.

At small viewport sizes, the card contents are not vertically centered:

screen shot 2017-04-06 at 8 07 27 am

At medium-ish viewports, 3 over 1 cards feels a bit odd:

screen shot 2017-04-06 at 8 07 43 am

Given we're likely to launch with just 2 cards, I eliminated 2 and noticed the hover state on the second item reveals uneven padding.

screen shot 2017-04-06 at 8 08 24 am

cshinn:
@rhamilto I just checked and was unable to repro the alignment and hover issues in any browser. 
screen shot 2017-04-06 at 8 17 53 am
screen shot 2017-04-06 at 8 19 21 am

As for centering the bottom item, I checked with a few people and we seemed to prefer it over left alignment (especially with fewer items) but if people feel strongly, I'm happy to change these to left-justified

rhamilto:
Yup.  You're correct, @cshinn.  Out of date dependencies was to blame for the responsive issues I was seeing.

spadgett:
We should make it look OK without descriptions since these are customizable, but OK as a follow on issue.

PR #66, opened 2017-03-31
Making panel animation sharper, faster

rhamilto:
The panel animation felt a little slow, so I compared what we had to what Bootstrap's modal does and essentially aligned with Bootstrap.  The result is a much crisper and responsive animation.

These screen recordings are a bit clunky because of the slow capture rate, but it should give you a sense of the before and after.

Before:
http://recordit.co/wDWaDmvd6d

After:
http://recordit.co/20cPrC4gZ7

cc: @spadgett


spadgett:
I prefer the faster animation.

PR #68, opened 2017-03-31
Prevent body scrolling behind open overlay panel

### LINKED ISSUE #49: Page should not scroll behind ordering panel
Using the mouse wheel or trackpad gesture for scrolling causes the entire page behind the overlay pane to scroll. This is distracting. When the modal is open, it should ideally lock the page scroll and only scroll contents inside the panel when needed.

spadgett:
Fixes #49


spadgett:
This is the technique Bootstrap uses for its modal

PR #71, opened 2017-03-31
[WIP] Fix to include ng-included html

jeff-phillips-18:


package.json
```diff
@@ -87,6 +87,7 @@
     "lodash": "3.10.1",
     "ng-annotate-loader": "^0.0.10",
     "ng-annotate-webpack-plugin": "^0.1.2",
+    "ngtemplate-loader": "^1.3.1",
```
dtaylor113: Where is this used?

dtaylor113: Is this a [WIP] because a follow up PR will use ngTemplate-loader?

jeff-phillips-18:
the template loader is somehting else I need to try. This doesn't work, but I hoped to avoid webpack hell.

PR #72, opened 2017-03-31
Catalog Service View: Do not show sub-category tiles in certain situations

dtaylor113:
**Do not show sub-category tiles when 'All' or 'Other' main category is selected**
image

**Do not show 'All' sub-category tile when there is only one sub-category tile**
image


spadgett:
It seems like we should also hide the Jenkins subcategory tile if it's the only one and just show the items.

dtaylor113:
> It seems like we should also hide the Jenkins subcategory tile if it's the only one and just show the items.

I'd be concerned that the items wouldn't convey the sub-category.  What happens if the catalog card said something like 'github jenkins webhook' or something not clearly 'Jenkins'.

spadgett:
The subcategories are just for organization. Hopefully the service classes have meaningful titles or it's a problem with the service class

dtaylor113:
> It seems like we should also hide the Jenkins subcategory tile if it's the only one and just show the items.
> I'd be concerned that the items wouldn't convey the sub-category. What happens if the catalog card said something like 'github jenkins webhook' or something not clearly 'Jenkins'.

@serenamarie125 ?

serenamarie125:
Hmmm, interesting question.  Although it would be cleaner to "hide" it, I think we should continue to show it @spadgett @dtaylor113.  For the same reason Dave mentioned as well as if individual brokers are trying to highlight categories of items

dtaylor113:
Hi @spadgett, please merge if you approve, thanks!

src/styles/services-view.less
```diff
@@ -105,6 +105,17 @@
...
+  .card-expansion {
+    overflow-y: auto;
+    .card-view-pf {
+      max-height: 410px;
+    }
+
+    &.no-sub-cats {
+      margin-top: 12px;
+    }
+  }
```
spadgett: Remove the `overflow-y` and `.card-view-pf max-height` styles from above (line 86) since they're already set here. Shouldn't need to declare them twice.

app/mockServices/mockData/openshift-images.ts
```diff
@@ -21,7 +21,7 @@ export const imagesData = {
                     'description': 'Provides a Jenkins 1.X server on CentOS 7. For more information about using this container image, including OpenShift considerations, see https://github.com/openshift/jenkins/blob/master/README.md.',
                     'iconClass': 'icon-jenkins',
                     'openshift.io/display-name': 'Jenkins 1.X',
-                    'tags': 'jenkins',
+                    'tags': 'builder,jenkins',
```
dtaylor113: There are several other non-builder images in the mock data which do not show up.

dtaylor113: I added 'builder' to Jenkins so it would show up as the only item/sub-category, so I could test the not showing 'All' sub-category when there is only one sub-category.

src/styles/services-view.less
dtaylor113: Thanks, I was trying that the other-way-round and it wasn't working.  Fixed.

app/mockServices/mockData/openshift-images.ts
dtaylor113: Locally, I removed the 'builder' tag from Jenkins and added a Jenkins serviceClass instead.  Let me know if you prefer that setup/mock data instead. -thanks

spadgett: > Locally, I removed the 'builder' tag from Jenkins and added a Jenkins serviceClass instead.

That makes more sense to me

dtaylor113: Ok, pushed those changes.  Should be able to merge once Travis finishes -thanks

PR #73, opened 2017-03-31
Fix to include ng-included html

jeff-phillips-18:
Unfortunately, each of the html files need to be listed separately.


spadgett:
Verified the fix.

PR #74, opened 2017-03-31
Add Guided Tour component and showcase usage

jeff-phillips-18:
The real tour step descriptions are still being worked.

This also adds the button to start the tour to the summary panel as well as some logic fix up to show the getting started area when 1 or less projects exist.

### DIFF
--- a/app/pages/home/homePageController.ts
+++ b/app/pages/home/homePageController.ts
@@ -1,16 +1,24 @@
+import * as jQuery from 'jquery';
+
 export class HomePageController {
-  static $inject = ['$state', 'AuthService', 'Logger', 'Constants', 'DataService'];
+  static $inject = ['$state', '$element', '$timeout', 'AuthService', 'Logger', 'Constants', 'DataService'];
 
   public ctrl: any = this;
   private $state: any;
+  private $element: any;
+  private $timeout: any;
   private authService: any;
   private logger: any;
   private constants: any;
   private dataService: any;
   private Logger: any;
+  private tourInProgress: boolean = false;
+  private gettingStartedTourConfig: any;
 
-  constructor($state: any, AuthService: any, Logger: any, Constants: any, DataService: any) {
+  constructor($state: any, $element: any, $timeout: any, AuthService: any, Logger: any, Constants: any, DataService: any) {
     this.$state = $state;
+    this.$element = $element;
+    this.$timeout = $timeout;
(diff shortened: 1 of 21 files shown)


app/pages/home/homePageController.ts
```diff
@@ -26,6 +34,61 @@ export class HomePageController {
     this.authService.withUser().then(() => {
       this.update();
     });
+
+    this.gettingStartedTourConfig = {
+      id: "getting-started-tour",
```
dtaylor113: Seems like this should be in the constants.js file.

jeff-phillips-18: I can understand the descriptions being in there. That has yet to be determined.

bower.json
```diff
@@ -18,6 +18,7 @@
     "jquery": "2.1.4",
     "lodash": "3.10.1",
     "origin-web-common": "^0.0.8"
```
dtaylor113: missing comma

dtaylor113:
I like it!   One issue is that Step 6 scrolls the browser down, then Step 7 appears cut off at top of page:

image

When I click on the browser to try to scroll up, Step 7 disappears.

jeff-phillips-18:
Right... nothing is fixed anymore 😞 

I'll remove the fixed settings but this will need to be updated when we re-fix position the components.

app/pages/home/homePageController.ts
```diff
@@ -57,4 +120,26 @@ export class HomePageController {
...
+    this.tourInProgress = true;
+  };
+
+  public handleTourEnd = () => {
+    this.tourInProgress = false;
+  };
+
+  public showAllCatalogs = () => {
+    var allSelector: any = jQuery(this.$element).find('.services-categories a');
+    allSelector[0].click();
```
dtaylor113: This might change with #72 as the 'All' category will be selected by default and all category cards shown by default.

jeff-phillips-18:
I've removed the fixed settings but the last bubble (pointing to Resources panel) is still pointing to blank space. That will get fixed when we fix the positioning.

app/pages/home/homePageController.ts
jeff-phillips-18: That's find. If the user had selected something else, it will go back to 'All' as planned.

jeff-phillips-18:
Closing to move the projects summary panel changes into a separate PR.

PR #80, opened 2017-04-03
Updates to project summary panel

### LINKED ISSUE #39: Project summary does not differentiate if 2 projects have the same display name

jeff-phillips-18:
Show Getting Started when less than 2 projects exist
Add project name and creator to project list items
Add 'View and Edit Membership" menu item to projects
Add Start Guided Tour button
Add resource links (actual links and text are TBD)

Fixes #39

### DIFF
--- a/src/components/projects-summary/projects-summary.html
+++ b/src/components/projects-summary/projects-summary.html
@@ -3,13 +3,8 @@
     <span class="fa fa-plus"></span>
     <span class="create-button-text">Create Project</span>
   </button>
-  <div ng-if="$ctrl.showGetStarted">
-    <h2 class="primary">Getting Started</h2>
-    <div class="getting-started-panel">
-      {{$ctrl.gettingStartedText}}
-    </div>
-  </div>
-  <h2 class="primary" ng-if="!$ctrl.showGetStarted">My Projects</h2>
+  <h2 class="primary" ng-if="!$ctrl.projects || !$ctrl.projects.length">Getting Started</h2>
+  <h2 class="primary" ng-if="$ctrl.projects && $ctrl.projects.length">My Projects</h2>
   <div class="catalog-modal catalog-modal-create-project" ng-if="$ctrl.showNewProjectPanel">
     <h4 class="catalog-modal-title">
       Create Project
@@ -19,15 +14,13 @@
       <span class="pficon pficon-close"></span>
     </a>
   </div>
-
-
-  <div ng-if="!$ctrl.showGetStarted" class="catalog-project-summary-list">
+  <div ng-if="$ctrl.projects && $ctrl.projects.length" class="catalog-project-summary-list">
     <div ng-if="$ctrl.totalProjects > $ctrl.maxDisplayProjects" class="projects-count">
       <strong>{{$ctrl.maxDisplayProjects}}</strong>
       of
       <strong>{{$ctrl.totalProjects}}</strong>
       Projects
-      <a href="" class="projects-view-all" ng-click="$ctrl.showAllProjects()">View All</a>
+      <a href="{{$ctrl.projectsUrl}}" class="projects-view-all">View All</a>
     </div>
     <div id="catalog-projects-summary-list">
       <div ng-repeat="project in $ctrl.projects track by (project | uid)" class="project-tile">
@@ -36,10 +29,12 @@
             <span class="fa fa-ellipsis-v"></span>
           </button>
           <ul class="uib-dropdown-menu dropdown-menu-right" aria-labelledby="dropdownKebab">
-            <li><a href="" ng-click="$ctrl.editProject(project)">Edit</a></li>
+            <li><a href="" ng-click="$ctrl.onViewMemebership(project)">View Membership</a></li>
+            <li><a href="" ng-click="$ctrl.editProject(project)">Edit Project</a></li>
             <li>
               <delete-link
                   kind="Project"
+                  label="Delete Project"
                   resource-name="{{project.metadata.name}}"
                   project-name="{{project.metadata.name}}"
                   display-name="{{(project | displayName)}}"
@@ -52,13 +47,16 @@
         </div>
         <h3 class="project-tile-header">
           <span ng-if="project.statusIconClass" class="project-status {{project.statusIconClass}}"></span>
-          <a href="" ng-click="$ctrl.handleProjectClick(project)" class="project-title">{{project | displayName}}</a>
+          <a href="{{project | projectUrl : $ctrl.baseProjectUrl}}" class="project-title">{{project | displayName}}</a>
         </h3>
         <p class="project-date">
-          {{project.metadata.creationTimestamp | date: 'mediumDate'}}
+          <span ng-if="project | displayName : true"><span ng-bind-html="project.metadata.name"></span> &ndash;</span>
+          created
+          <span ng-if="project | annotation : 'openshift.io/requester'">by <span ng-bind-html="project | annotation : 'openshift.io/requester'"></span></span>
+          <span am-time-ago="project.metadata.creationTimestamp"></span>
         </p>
         <div class="project-description" ng-if="project | description">
-          {{project | description}}
+          <truncate-long-text content="project | description" use-word-boundary="true" limit="120"></truncate-long-text>
         </div>
         <div class="catalog-modal catalog-modal-edit-project" ng-if="$ctrl.showEditProjectPanel && $ctrl.edittingProject === project">
           <h4 class="catalog-modal-title">
@@ -72,15 +70,24 @@
       </div>
     </div>
   </div>
-  <h2 class="secondary">Resources</h2>
-  <span>
-    {{$ctrl.resourcesText}}
-  </span>
-  <div class="resources-panel">
-    <ul>
-      <li ng-repeat="resource in $ctrl.resources">
-        {{resource.text}}
-      </li>
-    </ul>
+  <div ng-if="$ctrl.showGetStarted">
+    <div class="getting-started-panel">
+      <h2 class="secondary" ng-if="$ctrl.projects && $ctrl.projects.length > 0">Getting Started</h2>
+      <button class="getting-started-button btn btn-default" ng-class="{'with-projects': $ctrl.projects && $ctrl.projects.length}" ng-click="$ctrl.handleGettingStartedClick()">
+        <span class="fa fa-info-circle fa-2"></span>
+        <span class="getting-started-button-text">Start Guided Tour</span>
+      </button>
+    </div>
+    <div class="resources-panel">
+      <h2 class="secondary">Resources</h2>
+      <p>
+        {{$ctrl.resourceDescription}}
+      </p>
+      <ul>
+        <li ng-repeat="resource in $ctrl.resourceLinks">
+          <a href="{{resource.href}}" target="_blank" title="{{resource.href}}">{{resource.title}}</a>
+        </li>
+      </ul>
+    </div>
   </div>
 </div>
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -1,5 +1,15 @@
 import * as _ from 'lodash';
 
+_.set(window, 'OPENSHIFT_CONSTANTS.HELP_BASE_URL', "https://docs.openshift.org/latest/");
+var HELP: any = {
+  "new_app":                 "dev_guide/application_lifecycle/new_app.html",
+  "application_health":      "dev_guide/application_health.html",
+  "authorization":           "architecture/additional_concepts/authorization.html",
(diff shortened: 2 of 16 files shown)


dtaylor113:
Does this fix #39?

jeff-phillips-18:
Yes, Updated the description. Thanks @dtaylor113

test/projects-summary.spec.ts
```diff
@@ -313,6 +366,24 @@ describe('Projects Summary Panel', () => {
...
+    expect(showTourCount).toBe(0);
+
+    componentTest.eventFire(starTourButton[0], 'click');
+    componentTest.scope.$digest();
+
+    expect(showTourCount).toBe(1);
+  });
+
   it('should call the select project callback when the project name is clicked', () => {
     createProjectSummary();
```
dtaylor113: Hi, I noticed in your unit test:


```
it('should hide the create button when not allowed', () => {
    expectedCanCreate = false;
    createProjectSummary();

    var element = componentTest.rawElement;

    $timeout.flush();

    var createButton = jQuery(element).find('.create-button');
    expect(createButton.length).toBe(1);
  });
```
Shouldn't that last expect be '...toBe(0)' when the create button is hidden?

dtaylor113:
Hi, when I click on the 'Start Guided Tour' button, nothing happens?  No error msgs in console.

jeff-phillips-18:
Correct, the guided tour is coming soon, I didn't want this PR to be too large.

test/projects-summary.spec.ts
jeff-phillips-18: Yes, bad test. canCreate should reject the promise rather than returning false in the promise resolve. I have updated the test.

src/components/projects-summary/projects-summary.html
```diff
@@ -72,15 +69,24 @@ <h4 class="catalog-modal-title">
...
+      </button>
+    </div>
+    <div class="resources-panel">
+      <h2 class="secondary">Resources</h2>
+      <p>
+        {{$ctrl.resourceDescription}}
+      </p>
+      <ul>
+        <li ng-repeat="resource in $ctrl.resourceLinks">
+          <a href="" ng-click="$ctrl.onResourceClick(resource)">{{resource.title}}</a>
```
spadgett: We're using this pattern a lot for links, but it breaks things like right-click -> open new window / copy link. It also prevents being able to hover over the link and see the URL. Is there a way we can set a real href?

I have the same concern for the project links, which are all callbacks.

src/components/projects-summary/projects-summary.html
```diff
@@ -36,6 +29,7 @@ <h4 class="catalog-modal-title">
             <span class="fa fa-ellipsis-v"></span>
           </button>
           <ul class="uib-dropdown-menu dropdown-menu-right" aria-labelledby="dropdownKebab">
+            <li><a href="" ng-click="$ctrl.onViewMemebership(project)">View and Edit Membership</a></li>
```
spadgett: This is a long label :/ Maybe it should just be "View Membership"?

@jwforres any suggestions?

src/components/projects-summary/projects-summary.html
```diff
@@ -55,10 +49,13 @@ <h3 class="project-tile-header">
           <a href="" ng-click="$ctrl.handleProjectClick(project)" class="project-title">{{project | displayName}}</a>
         </h3>
         <p class="project-date">
+          <span ng-if="project | displayName : true"><span ng-bind-html="project.metadata.name | highlightKeywords : keywords"></span> &ndash;</span>
+          created
+          <span ng-if="project | annotation : 'openshift.io/requester'">by <span ng-bind-html="project | annotation : 'openshift.io/requester' | highlightKeywords : keywords"></span></span>
           {{project.metadata.creationTimestamp | date: 'mediumDate'}}
```
spadgett: This isn't new, but was it intentional to switch from relative timestamps? The project list page currently uses strings like "2 days ago"

src/components/projects-summary/projects-summary.html
```diff
@@ -3,13 +3,8 @@
...
   </button>
-  <div ng-if="$ctrl.showGetStarted">
-    <h2 class="primary">Getting Started</h2>
-    <div class="getting-started-panel">
-      {{$ctrl.gettingStartedText}}
-    </div>
-  </div>
-  <h2 class="primary" ng-if="!$ctrl.showGetStarted">My Projects</h2>
+  <h2 class="primary" ng-if="!$ctrl.projects || $ctrl.projects.length < 1">Getting Started</h2>
+  <h2 class="primary" ng-if="$ctrl.projects && $ctrl.projects.length > 0">My Projects</h2>
```
spadgett: This could be `!$ctrl.projects.length` and `$ctrl.projects.length`

src/components/projects-summary/projects-summary.html
```diff
@@ -19,9 +14,7 @@ <h4 class="catalog-modal-title">
       <span class="pficon pficon-close"></span>
     </a>
   </div>
-
-
-  <div ng-if="!$ctrl.showGetStarted" class="catalog-project-summary-list">
+  <div ng-if="$ctrl.projects && $ctrl.projects.length > 0" class="catalog-project-summary-list">
```
```
ng-if="$ctrl.projects.length"
```

src/components/projects-summary/projects-summary.html
```diff
@@ -55,10 +49,13 @@ <h3 class="project-tile-header">
           <a href="" ng-click="$ctrl.handleProjectClick(project)" class="project-title">{{project | displayName}}</a>
         </h3>
         <p class="project-date">
+          <span ng-if="project | displayName : true"><span ng-bind-html="project.metadata.name | highlightKeywords : keywords"></span> &ndash;</span>
```
spadgett: Remove use of `highlightKeywords` since we have no filtering on this page (here are below).

src/components/projects-summary/projects-summary.html
```diff
@@ -72,15 +69,24 @@ <h4 class="catalog-modal-title">
...
-  </span>
-  <div class="resources-panel">
-    <ul>
-      <li ng-repeat="resource in $ctrl.resources">
-        {{resource.text}}
-      </li>
-    </ul>
+  <div ng-if="$ctrl.showGetStarted">
+    <div class="getting-started-panel">
+      <h2 class="secondary" ng-if="$ctrl.projects && $ctrl.projects.length > 0">Getting Started</h2>
```
```
ng-if="$ctrl.projects.length"
```

src/components/projects-summary/projects-summary.html
```diff
@@ -72,15 +69,24 @@ <h4 class="catalog-modal-title">
...
+      <button class="getting-started-button btn btn-default" ng-class="{'with-projects': $ctrl.projects && $ctrl.projects.length > 0}" ng-click="$ctrl.handleGettingStartedClick()">
```
```
ng-class="{'with-projects': $ctrl.projects.length }"
```

src/constants.ts
```diff
@@ -1,5 +1,55 @@
 import * as _ from 'lodash';
 
+_.set(window, 'OPENSHIFT_CONSTANTS.HELP_BASE_URL', "https://docs.openshift.org/latest/");
+var HELP: any = {
```
spadgett: I'm concerned about copying all of these into this repository. How are we going to keep this in sync across repos?

src/components/projects-summary/projects-summary.html
jeff-phillips-18: Yes. @beanh66 asked to display the date rather than the momentAgo.

src/constants.ts
jeff-phillips-18: Agreed. I wasn't sure of a good way to do this. These are not important to be the real links as what is set in console will override these. I can leave them out but it was a good way to test from within the showcase app.

src/components/projects-summary/projects-summary.html
spadgett: I'm not necessarily against this, but everywhere else in the web console we use relative times. I can't think of anywhere we don't. Also the full project list shows relative times, so it would be odd to be different here.

cc @serenamarie125 @jwforres

src/components/projects-summary/projects-summary.html
jeff-phillips-18: I've updated it for the resource links here. Adding this for project link would require a bit of rework since the href is not known here. We could pass in a base url rather than a callback if this is desired. Thoughts?

src/constants.ts
spadgett: Can we just copy the few we want to show?

jeff-phillips-18: Removed the items not being used.

src/components/projects-summary/projects-summary.html
beanh66: @spadgett I agree with you on the oddity of only making the change in one place. Customers have asked to see explicit dates rather than relative dates, but we can wait to make that change (post MVP) once we can change the full projects list as well. @jeff-phillips-18 Does that work for you?

spadgett: It's not hard to change the full project list if we decide it's better.

src/components/projects-summary/projects-summary.html
```diff
@@ -3,13 +3,8 @@
...
   </button>
-  <div ng-if="$ctrl.showGetStarted">
-    <h2 class="primary">Getting Started</h2>
-    <div class="getting-started-panel">
-      {{$ctrl.gettingStartedText}}
-    </div>
-  </div>
-  <h2 class="primary" ng-if="!$ctrl.showGetStarted">My Projects</h2>
+  <h2 class="primary" ng-if="!$ctrl.projects || !$ctrl.projects.length">Getting Started</h2>
+  <h2 class="primary" ng-if="$ctrl.projects && $ctrl.projects.length">My Projects</h2>
```
spadgett: Nit: Things like `ng-if="$ctrl.projects && $ctrl.projects.length"'` could just be `ng-if="$ctrl.project.length"`, but this is fine.

src/components/projects-summary/projects-summary.html
```diff
@@ -52,13 +46,16 @@ <h4 class="catalog-modal-title">
         </div>
         <h3 class="project-tile-header">
           <span ng-if="project.statusIconClass" class="project-status {{project.statusIconClass}}"></span>
-          <a href="" ng-click="$ctrl.handleProjectClick(project)" class="project-title">{{project | displayName}}</a>
+          <a href="{{$ctrl.baseProjectUrl + '/' + project.metadata.name}}" class="project-title">{{project | displayName}}</a>
```
spadgett: Should we try to check if base project URL already ends with `/` ?

src/constants.ts
```diff
@@ -43,3 +53,36 @@ export const allSaasOfferings = saasOfferings;
...
+      help: 'application_health'
+    },
+    {
+      title: 'Visit us on Facebook',
+      href: 'http://facebook.com/RedHatInc'
+    }
+  ]
+};
+
+_.set(window, 'OPENSHIFT_CONSTANTS.RESOURCES', resources);
```
spadgett: Should have commented on first review, but we should really give a name that isn't so generic. Maybe `CATALOG_HELP_RESOURCES` or similar.

src/components/projects-summary/projects-summary.html
jeff-phillips-18: Yeah, I like to not rely on angular being ok with referencing null objects' fields. Habit.

spadgett: OK :)

src/styles/projects-summary.less
```diff
@@ -24,6 +24,28 @@
       font-size: @font-size-base + 6;
     }
   }
+
+  .getting-started-panel {
```
rhamilto: Nit:  alphabetization

src/styles/projects-summary.less
```diff
@@ -24,6 +24,28 @@
...
+    margin-top: @grid-gutter-width;
+  }
+
+  .getting-started-button {
+    width: 100%;
```
rhamilto: Nit: alphabetization

src/components/projects-summary/projects-summary.html
jeff-phillips-18: OK, added a filter to do this.

src/constants.ts
jeff-phillips-18: +1  Fixed

src/styles/projects-summary.less
jeff-phillips-18: Fixed

src/styles/projects-summary.less
jeff-phillips-18: Fixed

src/components/projects-summary/projects-summary.html
spadgett: Works for me, thanks.

jeff-phillips-18:
@rhamilto Any further issues?

PR #81, opened 2017-04-03
Add search to catalog

spadgett:
Basic search support, with little styling.

screen shot 2017-04-03 at 4 20 12 pm

TODO:

- [ ] Style the search results (@cshinn)
- [x] Support image URLs, not just font icons
- [x] Show tags for builder images
- [x] Use KeywordsService (currently broken with mock data)
- [x] Defer search when user starts typing before items are loaded
- [ ] See if we can have one overlay-panel instead of having to duplicate the logic. Currently it's tied to the services view.
- [ ] Add unit tests

### DIFF
--- a/src/components/catalog-search/catalog-search.controller.ts
+++ b/src/components/catalog-search/catalog-search.controller.ts
@@ -1,0 +1,88 @@
+import * as _ from 'lodash';
+
+export class CatalogSearchController implements angular.IController {
+  static $inject = ['$q', 'Catalog', 'KeywordService'];
+
+  public ctrl: any = this;
+
+  private Catalog: any;
+  private KeywordService: any;
+  private allItems: any[];
+  private $q: any;
+  private loaded: boolean = false;
+  // Used when the user starts typing before the items have loaded.
+  private searchDeferred: ng.IDeferred<any[]>;
+
+  constructor($q: any, Catalog: any, KeywordService: any) {
+    this.$q = $q;
+    this.Catalog = Catalog;
+    this.KeywordService = KeywordService;
+  }
+
+  public $onInit() {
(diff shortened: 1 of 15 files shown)


spadgett:
I hate having multiple WIP in flight, but I know @cshinn was going to look at styling the search control.

spadgett:
@dtaylor113 I checked GitHub markdown syntax, and I didn't see any easy way to make the "not ready for code review" text red :)

spadgett:
cc @serenamarie125

pmorie:
ship it

spadgett:
@jwforres @dtaylor113 This is probably far enough along for a review.

@jwforres I didn't debounce the search because it didn't feel as good. There's a `typeahead-wait-ms` setting on the typeahead control if we want it.

src/components/catalog-search/catalog-search.controller.ts
```diff
@@ -0,0 +1,59 @@
...
+      items = items.concat(_.map(this.ctrl.imageStreams, (imageStream) => {
+        return this.Catalog.getImageItem(imageStream);
+      }));
+
+      // Remove null items (non-builder images).
+      items = _.reject(items, (item) => {
+        return !item;
+      });
+
+      this.ctrl.allItems = _.sortBy(items, 'name');
```
jwforres: is it possible for someone to start typing into the filter while the service class or image streams are still loading? do you need to run filter here

spadgett:
> is it possible for someone to start typing into the filter while the service class or image streams are still loading? do you need to run filter here

Added a second commit that defers the search until items load.

spadgett:
Squashed and rebased

jwforres:
can merge once travis passes

PR #83, opened 2017-04-04
Fixing the position of the sidebar at >= 768px wide viewports

rhamilto:
and excluding the navbar from the scrollable page area (adds inner page scroll).

In the console, all we should need to do is override @landing-side-bar-top-offset when consuming the less partials.  Note:  we will not want to add main.less itself in the console since it includes variables.less, and variables.less imports dependency variables.


app/styles/variables.less
```diff
@@ -1,4 +1,4 @@
 // copied from origin-web-console
 @panel-shaded: #f2f2f2;
 
-@navbar-height: 69px;
+@navbar-height: 72px;
```
jeff-phillips-18: I'd rather use flex box so that there is no need to keep track of the navbar height.

src/styles/landing-page.less
```diff
@@ -44,14 +44,19 @@
   background-color: @landing-side-bar-bg;
   color: @color-pf-white;
   @media(min-width: @screen-sm-min) {
-    flex: 0 1 auto;
-    width: 325px;
+    bottom: 0;
+    overflow: hidden;
+    overflow-y: auto;
+    position: fixed;
+    right: 0;
```
jeff-phillips-18: This covers the scroll bar.

rhamilto: There are a number of variables that determine how scroll bars are painted:  OS, OS scrollbar preferences, and browser.  For example, Chrome 56 and Firefox 52 on macOS Sierra with the default macOS scroll bar setting (Show scroll bar: Automatically based on mouse or trackpad) correctly paint the scrollbar on top of the sidebar when the content area is scrolled.  Oddly, Safari 10.1 on the same OS and setting does not.  As best I can tell, this is because of the addition of the inner scroll.   I am thinking it's prudent to eliminate the inner scroll for now so that we can keep things moving forward as there is much additional styling work to be done.   Any objections?  cc:  @jwforres, @spadgett

app/styles/variables.less
rhamilto: Can we punt on this?  Given this only affects the demo app, I'd rather spend cycles getting unstyled parts styled.

src/styles/landing-page.less
jeff-phillips-18: Setting the right position to 15px shows the scrollbar.

spadgett: > Setting the right position to 15px shows the scrollbar.

When there is no scrollbar, is there a right gutter? Are different widths needed for different browsers and platforms since they style scrollbars differently?

jeff-phillips-18: Yes, it leaves a gutter if no scroll bar is shown. But, to me it is worse to see cutoff content and no scroll bar.

spadgett: @jeff-phillips-18 I agree there's a problem here. Do you mind if we make it a follow-on? The right fix isn't obvious to me. It feels like a problem because two scrollbars are competing for the same place on the screen, and potentially we need to test many different browsers and platforms.

jeff-phillips-18: I'm ok with filing an issue and moving forward for sure.

There is only one scrollbar, it just gets covered by the side bar 😉

spadgett: There are two, one for the main content and one for the sidebar content :/

spadgett: Thanks @jeff-phillips-18, @rhamilto can you open an issue?

PR #85, opened 2017-04-05
Add visual styles to search

cshinn:
Visual updates to landing page search bar. Does not include css to make the search bar sticky.

screen shot 2017-04-05 at 2 52 16 pm

### DIFF
--- a/dist/origin-web-catalogs.css
+++ b/dist/origin-web-catalogs.css
@@ -37,6 +37,9 @@
 .landing-main-area {
   background-color: #3ca2c9;
 }
+.landing-main-area h1 {
+  margin-left: 20px;
+}
 @media (min-width: 768px) {
   .landing-main-area {
     margin-right: 325px;
@@ -52,6 +55,15 @@
     margin-right: 450px;
   }
 }
+.landing-search-area {
+  border-bottom: solid 1px #7dc3e8;
+  padding: 10px;
+}
+@media (min-width: 768px) {
+  .landing-search-area {
+    padding: 13.33333333px;
+  }
+}
 .landing-search-form {
   margin: 0 auto;
   width: 100%;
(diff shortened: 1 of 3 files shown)


src/styles/landing-page.less
```diff
@@ -11,6 +10,14 @@
   }
 }
 
+.landing-search-area {
+  border-bottom: solid 1px @color-pf-blue-200;
+  padding: (@grid-gutter-width / 4);
+  @media (min-width: @screen-sm-min) {
+      padding: (@grid-gutter-width / 3);
```
spadgett: nit: indentation

src/styles/landing-page.less
```diff
@@ -19,39 +26,48 @@
...
   }
   .dropdown-menu {
+    margin-top: 0;
     width: 100%;
     // Styles need to be specific enough to override defaults.
     > li > a.catalog-search-match {
       padding-bottom: 5px;
       padding-top: 5px;
     }
+    // important necessary to override defaults
```
spadgett: I'd be more specific: "important needed to override styles declared important in Patternfly"

dist/origin-web-catalogs.css
```diff
@@ -34,9 +34,6 @@
     opacity: 0;
   }
 }
-.landing-main-area {
-  padding: 20px;
```
spadgett: Wonder if we shouldn't give a temporary margin to the headings if we're removing the padding on the entire page.

spadgett:
I noticed that the individual tags highlight when you mouse over them. It makes it look like the tag does something special when clicked.

cshinn:
@spadgett I've fixed the hover issue. It won't let me squash this commit for some reason though...

spadgett:
We should add the hourglass icon to the search input, either in this PR or a follow-on

spadgett:
@cshinn Conflict now that I merged #65 :/

jeff-phillips-18:
I agree, the search component seems small to me.

spadgett:
@chsinn is also looking at fixing the search to the top. Suggest we use this PR for styling the search result and a follow on for search input styling (position, size, and hourglass icon)

PR #78, opened 2017-04-01
Create from builder image

spadgett:
**Not ready for code review**, but basic function is working.

screen shot 2017-04-01 at 4 37 55 pm

TODO

- [x] Link out to advanced options
- [ ] Test if objects exist before trying to create them
- [x] Show which objects failed on errors
- [ ] Handle context dir and git ref when set for sample repos
- [x] Don't show builders with missing status tags in version dropdown
- [x] Show tags that reference other tags in the version dropdown (e.g. `4 -- latest`)
- [x] Maintain server (semver) ordering
- [ ] Quota checks
- [x] Prevent returning to previous steps after the form is submitted
- [x] Let users create new projects from the builder dialog
- [x] Handle builders from search results
- [ ] Unit tests

@jwforres FYI

PR #86, opened 2017-04-05
select-project component

dtaylor113:
Created a `<select-project>` component which allows one to select or create a new project.  Integrated into the serviceClass order service:

image
The most recently created project is selected by default.

**Project keyword search:**
image

**Create Project with validation:**
image
Duplicate project name validation also implemented.

### DIFF
--- a/src/components/order-service/order-service.controller.ts
+++ b/src/components/order-service/order-service.controller.ts
@@ -27,9 +27,11 @@ export class OrderServiceController implements angular.IController {
     this.ctrl.description = this.ctrl.serviceClass.description;
     this.ctrl.longDescription = this.ctrl.serviceClass.longDescription;
     this.ctrl.plans = _.get(this, 'ctrl.serviceClass.resource.plans', []);
+    this.ctrl.forms = {};
 
     // Preselect the first plan. If there's only one plan, skip the wizard step.
     this.ctrl.selectedPlan = _.first(this.ctrl.plans);
+    this.ctrl.selectedProject = {};
     this.ctrl.planIndex = 0;
     this.ctrl.steps = [{
       id: 'plans',
@@ -49,7 +51,7 @@ export class OrderServiceController implements angular.IController {
       this.ctrl.steps.shift();
     }
     this.gotoStep(this.ctrl.steps[0]);
-    this.listProjects();
+    this.ctrl.nameTaken = false;
     this.ctrl.wizardReady = true;
   }
 
@@ -94,7 +96,44 @@ export class OrderServiceController implements angular.IController {
     this.ctrl.selectedPlan = plan;
   }
 
+  public invalidNewProject() {
+    // test whether selectedProject is actually a newProject obj. and if passed name validation
+    return this.isNewProject() && !this.ctrl.selectedProject.validName;
+  }
+
   public provisionService() {
+    if (this.isNewProject()) {
+      // the selectedProject is actually a newProject object
+      let newProjName = this.ctrl.selectedProject.metadata.name;
+      let newProjDisplayName = this.ctrl.selectedProject.metadata.annotations['new-display-name'];
+      let newProjDesc = this.$filter('description')(this.ctrl.selectedProject);
+      let projReqObj: any = {
+        apiVersion: "v1",
+        kind: "ProjectRequest",
+        metadata: {
+          name: newProjName
+        },
+        displayName: newProjDisplayName,
+        description: newProjDesc
+      };
+      this.DataService
+        .create('projectrequests', null, projReqObj, this.$scope)
+        .then( (data: any) => {
+          this.createService();
+        }, (result: any) => {
+          var data = result.data || {};
+          if (data.reason === 'AlreadyExists') {
+            this.ctrl.nameTaken = true;
+          } else {
+            this.ctrl.error = data.message || 'An error occurred creating the project.';
+          }
+        });
+    } else {
+      this.createService();
+    }
+  }
+
+  public createService() {
     let serviceInstance = this.makeServiceInstance();
     this.DataService.create({
       group: 'servicecatalog.k8s.io',
@@ -111,40 +150,26 @@ export class OrderServiceController implements angular.IController {
     });
   }
 
-  public $onChanges(onChangesObj: angular.IOnChangesObject) {
-    return;
-  }
-
-  public $doCheck() {
-    return;
-  }
-
   public closePanel() {
     if (angular.isFunction(this.ctrl.handleClose)) {
       this.ctrl.handleClose();
     }
   }
 
-  private listProjects() {
-    this.DataService.list('projects', this.$scope).then((response: any) => {
-      this.ctrl.projects = _.sortBy(response.by('metadata.name'), this.$filter('displayName'));
-      this.ctrl.selectedProject = _.first(this.ctrl.projects);
-    });
-  }
-
   private makeServiceInstance() {
     let serviceClassName = _.get(this, 'ctrl.serviceClass.resource.metadata.name');
+
     let serviceInstance = {
       kind: 'Instance',
       apiVersion: 'servicecatalog.k8s.io/v1alpha1',
       metadata: {
         namespace: this.ctrl.selectedProject.metadata.name,
         generateName: serviceClassName + '-'
-      },
-      spec: {
-        serviceClassName: serviceClassName,
-        planName: this.ctrl.selectedPlan.name
-      }
+       },
+       spec: {
+         serviceClassName: serviceClassName,
+         planName: this.ctrl.selectedPlan.name
+       }
     };
 
     return serviceInstance;
@@ -153,5 +178,9 @@ export class OrderServiceController implements angular.IController {
   private gotoStepID(id: string) {
     let step = _.find(this.ctrl.steps, { id: id });
     this.gotoStep(step);
+  }
+
+  private isNewProject(): boolean {
+    return this.ctrl.selectedProject && !_.has(this.ctrl.selectedProject, 'metadata.uid');
   }
 }
--- a/src/components/order-service/order-service.html
+++ b/src/components/order-service/order-service.html
@@ -75,7 +75,7 @@
         type="button"
         class="btn btn-primary wizard-pf-next"
         ng-if="$ctrl.currentStep.id === 'configure'"
-        ng-disabled="!$ctrl.selectedProject || $ctrl.currentStep.id === 'results'"
+        ng-disabled="!$ctrl.selectedProject || $ctrl.currentStep.id === 'results' || $ctrl.forms.orderConfigureForm.$invalid"
         ng-click="$ctrl.provisionService()">
       Create
       <span class="fa fa-angle-right" aria-hidden="true"></span>
--- a/src/components/select-project/select-project.controller.ts
+++ b/src/components/select-project/select-project.controller.ts
@@ -1,0 +1,82 @@
+import * as angular from 'angular';
+import * as _ from 'lodash';
+
+export class SelectProjectController implements angular.IController {
+  static $inject = ['$scope', '$filter', 'DataService', 'ProjectsService', 'Logger', 'AuthService'];
+
+  public ctrl: any = this;
+  public $scope: any;
+
+  private $filter: any;
+  private DataService: any;
+  private ProjectsService: any;
+  private Logger: any;
+  private AuthService: any;
+
+  constructor($scope: any, $filter: any, DataService: any, ProjectsService: any, Logger: any, AuthService: any) {
+    this.$scope = $scope;
+    this.$filter = $filter;
+    this.DataService = DataService;
+    this.AuthService = AuthService;
+    this.ProjectsService = ProjectsService;
+    this.Logger = Logger;
+  }
+
+  public $onInit() {
+    this.ctrl.nameTaken = false;
+
+    this.ProjectsService.canCreate().then(() => {
+      this.ctrl.canCreate = true;
+    }, (result) => {
+      this.ctrl.canCreate = false;
+      // 403 Forbidden indicates the user doesn't have authority.
+      // Any other failure status is an unexpected error.
+      if (result.status !== 403) {
+        var msg = 'Failed to determine create project permission';
+        if (result.status !== 0) {
+          msg += " (" + result.status + ")";
+        }
+        this.Logger.warn(msg);
+      }
+    }).finally(() => {
+      this.listProjects();
+    });
+  }
+
+  public $onChanges(onChangesObj: angular.IOnChangesObject) {
+    if (onChangesObj.nameTaken && !onChangesObj.nameTaken.isFirstChange()) {
+      this.ctrl.forms.createProjectForm.name.$setValidity('nameTaken', !this.ctrl.nameTaken);
+    }
+  }
+
+  public onNewProjectNameChange() {
+    this.ctrl.nameTaken = false;
+    this.ctrl.forms.createProjectForm.name.$setValidity('nameTaken', !this.ctrl.nameTaken);
+  }
+
+  public isNewProject(): boolean {
+    return this.ctrl.selectedProject && !_.has(this.ctrl.selectedProject, 'metadata.uid');
+  }
+
+  private listProjects() {
+    this.DataService.list('projects', this.$scope).then((response: any) => {
+      // 'Create Project' placeholder obj in the dropdown
+      let createProject = {
+        "metadata": {
+          "annotations": {
+            "openshift.io/display-name": "Create Project",
+            "new-display-name": ""
+          }
+        }
+      };
+      this.ctrl.projects = _.sortBy(response.by('metadata.name'), this.$filter('displayName'));
+
+      // get most recently created
+      this.ctrl.selectedProject = this.$filter('mostRecent')(this.ctrl.projects);
+
+      if (this.ctrl.canCreate) {
+        this.ctrl.projects.unshift(createProject);
+      }
+    });
+  }
+}
--- a/src/components/select-project/selectProject.html
+++ b/src/components/select-project/selectProject.html
@@ -1,0 +1,91 @@
+<ng-form>
+    <div class="form-group">
+        <label class="col-sm-4 control-label" for="project">Add to Project</label>
+      <div class="col-sm-8">
+        <ui-select ng-model="$ctrl.selectedProject">
+          <ui-select-match placeholder="Project name">
+            {{$select.selected | displayName}}
+          </ui-select-match>
+          <ui-select-choices repeat="project in $ctrl.projects | searchProjects : $select.search">
+            <span ng-bind-html="project | displayName | highlightKeywords : $select.search"></span>
+                <span ng-if="project | displayName : true" class="small text-muted">
+                  <span ng-if="project.metadata.name">&ndash;</span>
+                  <span ng-bind-html="project.metadata.name | highlightKeywords : $select.search"></span>
+                </span>
+          </ui-select-choices>
+        </ui-select>
+      </div>
+    </div>
+</ng-form>
+
+<ng-form name="$ctrl.forms.createProjectForm"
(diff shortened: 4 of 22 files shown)


spadgett:
Suggest we should remove the help text from below the project name field since we're not adding it to other fields (at least not yet).

Screenshots look great!

dtaylor113:
> Suggest we should remove the help text from below the project name field since we're not adding it to other fields (at least not yet).

That is the standard which the Project Summary uses and `<create-project>` dialog uses.

dtaylor113:
Also pls note that I haven't used the `<select-project>` component nor the serviceClass order service against real data.  I plan on setting up my system tomorrow to do this.

dtaylor113:
I'm also seeing a highlight issue when selecting a project:

image

I believe we are getting this because we are using a `<ui-select>` within a `<form>` I spent some time trying to override it in less/css but couldn't find a solution.  @rhamilto I will create an issue if I can't find a solution.  -thanks

src/components/order-service/order-service.controller.ts
```diff
@@ -154,4 +180,8 @@ export class OrderServiceController implements angular.IController {
     let step = _.find(this.ctrl.steps, { id: id });
     this.gotoStep(step);
   }
+
+  private isNewProject(): boolean {
+    return this.ctrl.selectedProject && this.ctrl.selectedProject.metadata ? this.ctrl.selectedProject.metadata.uid === 'create_project' : false;
```
spadgett: I'd rather leave `uid` undefined instead of setting it to something else. Then a project without a `uid` is new.

I know this will never actually happen, but technically `create_project` is a valid UID for a real project :)

spadgett: Maybe

```js
return this.ctrl.selectedProject && !_.has(this.ctrl.selectedProject, 'metadata.uid');
```

src/components/order-service/order-service.html
```diff
@@ -75,7 +75,7 @@
         type="button"
         class="btn btn-primary wizard-pf-next"
         ng-if="$ctrl.currentStep.id === 'configure'"
-        ng-disabled="!$ctrl.selectedProject || $ctrl.currentStep.id === 'results'"
+        ng-disabled="!$ctrl.selectedProject || $ctrl.currentStep.id === 'results' || $ctrl.invalidNewProject()"
```
spadgett: Better to have a nested `ng-form` for new project and make that form invalid. Then we can check form validity similar to 

https://github.com/openshift/origin-web-catalog/pull/78/files#diff-a816c7d825f3b359053f0814dcc271f2R32

src/components/select-project/select-project.controller.ts
```diff
@@ -0,0 +1,93 @@
...
+    this.DataService = DataService;
+    this.AuthService = AuthService;
+    this.ProjectsService = ProjectsService;
+    this.Logger = Logger;
+  }
+
+  public $onInit() {
+    this.ctrl.nameTaken = false;
+    // TODO disableInputs?
+    this.ctrl.disableInputs = false;
```
spadgett: I don't think this is needed, or at least it should be on the entire form, not just the select project control.

src/components/select-project/select-project.controller.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    this.ProjectsService.canCreate().then(() => {
+      this.ctrl.canCreate = true;
+      this.AuthService.withUser().then(this.onVerifyUser);
```
spadgett: `withUser` shouldn't be needed. You should be logged in at this point. You would need to call it before `canCreate()` anyway.

src/components/select-project/select-project.controller.ts
```diff
@@ -0,0 +1,93 @@
...
+      // 403 Forbidden indicates the user doesn't have authority.
+      // Any other failure status is an unexpected error.
+      if (result.status !== 403) {
+        var msg = 'Failed to determine create project permission';
+        if (result.status !== 0) {
+          msg += " (" + result.status + ")";
+        }
+        this.Logger.warn(msg);
+      }
+      this.AuthService.withUser().then(this.onVerifyUser);
```
spadgett: Code shared by both success and failure can use the `finally` method.

src/components/select-project/select-project.controller.ts
```diff
@@ -0,0 +1,93 @@
...
+            "openshift.io/display-name": "Create Project",
+            "new-display-name": ""
+          },
+          validName: false
+        }
+      };
+      this.ctrl.projects = _.sortBy(response.by('metadata.name'), this.$filter('displayName'));
+
+      // get most recently created
+      this.ctrl.selectedProject = _.first(this.$filter('orderObjectsByDate')(this.ctrl.projects, true));
```
spadgett: You can loop through the projects once -- O(n) -- to find most recent rather than sorting -- O(n log(n)). It's possible for admins to see hundreds or thousands of projects, so it can make a difference.

Here is an example:

https://github.com/openshift/origin-web-console/blob/8f4cda159d1382b7e5e2d4c3223b4ac248be6e1f/app/scripts/filters/date.js#L99-L110

src/components/order-service/order-service.controller.ts
spadgett: An alternative is to have a property like `newProjectOption: true`

rhamilto:
>  @rhamilto I will create an issue if I can't find a solution.

That may be a global console bug.  Please open an issue.

spadgett:
@dtaylor113 Do you see the ui-select bug in the console as well? I noticed noticed the caret doesn't look positioned properly in your screenshots.

src/components/select-project/selectProject.html
```diff
@@ -0,0 +1,93 @@
...
+    </div>
+</form>
+
+<form name="$ctrl.createProjectForm" class="form-horizontal"
+      ng-if="$ctrl.isNewProject()">
+  <fieldset ng-disabled="$ctrl.disableInputs">
+    <div class="form-group">
+      <label for="name" class="col-sm-4 control-label required">Project Name</label>
+      <span class="col-sm-8" ng-class="{'has-error': ($ctrl.createProjectForm.name.$error.pattern && $ctrl.createProjectForm.name.$touched) || nameTaken}">
+        <input class="form-control input-lg"
```
spadgett: We shouldn't be using `input-lg`

src/components/select-project/select-project.controller.ts
```diff
@@ -0,0 +1,93 @@
...
+        }
+        this.Logger.warn(msg);
+      }
+      this.AuthService.withUser().then(this.onVerifyUser);
+    });
+  }
+
+  public $onChanges(onChangesObj: angular.IOnChangesObject) {
+    if (onChangesObj.nameTaken && !onChangesObj.nameTaken.isFirstChange()) {
+      this.ctrl.nameTaken = onChangesObj.nameTaken.currentValue;
```
jeff-phillips-18: this is not necessary, it will already have been assigned.

jeff-phillips-18:
I think it might be better to have the existing projects list, then a + button to the left which simply pops up the create project panel (like in the projects summary panel). That would remove the redundancy of creating  a project panel here.

src/components/order-service/order-service.html
dtaylor113: Hi @spadgett, the new project form is in the select project component, not in order service html.  Not sure what you are suggesting here.

spadgett: You can nest forms using `ng-form`.

https://docs.angularjs.org/api/ng/directive/ngForm

I'm suggesting the order service configuration page has a form. The select project component uses ng-form. Then if the select project form is invalid, the outer order service form will be marked invalid. Then we disable the button whenever the form is invalid.

We will need form validation anyway when we add parameters.

src/components/select-project/select-project.controller.ts
dtaylor113: Ok, this is very similar to how the project summary is doing it.  Should I also change it here:  https://github.com/openshift/origin-web-catalog/blob/master/src/components/projects-summary/projects-summary.controller.ts#L36
-thanks

src/components/order-service/order-service.html
spadgett: Here's an example of how you can use ng-form:

https://github.com/openshift/origin-web-console/blob/95b0902ecb3627301fedf2a0a6855cbd5e8c39a2/app/views/directives/select-containers.html

Then we embedded the `select-container` directive in a larger form, which was marked invalid if a container wasn't selected.

src/components/select-project/select-project.controller.ts
spadgett: I think so. You can't properly check `canCreate` without being logged in first.

src/components/select-project/selectProject.html
dtaylor113: For all the input fields or just name?

src/components/order-service/order-service.controller.ts
```diff
@@ -111,40 +149,28 @@ export class OrderServiceController implements angular.IController {
...
   private makeServiceInstance() {
     let serviceClassName = _.get(this, 'ctrl.serviceClass.resource.metadata.name');
+    let projectName = _.get(this, 'ctrl.selectedProject.metadata.name');
+    let planName = _.get(this, 'ctrl.selectedPlan.name');
+
+    let spec: any = {serviceClassName: serviceClassName};
+    spec = planName ? _.assign(spec, {planName: planName}) : spec;
+
+    let metadata: any = {generateName: serviceClassName + '-'};
+    metadata = projectName ? _.assign(metadata, {namespace: projectName}) : metadata;
```
dtaylor113: Put back original code, updated mockData to always have at least one plan.

src/components/order-service/order-service.controller.ts
dtaylor113: Fixed.  Using an undefined `uid` to determine if creating a new project.

src/components/order-service/order-service.html
dtaylor113: Fixed.

src/components/select-project/select-project.controller.ts
dtaylor113: Fixed. Removed

src/components/select-project/select-project.controller.ts
dtaylor113: Fixed in both places.

src/components/select-project/select-project.controller.ts
dtaylor113: Fixed

src/components/select-project/select-project.controller.ts
dtaylor113: Waiting on origin-web-common PR to gain access to those filters.

dtaylor113:
> That may be a global console bug. Please open an issue.

#87

dtaylor113:
> @dtaylor113 Do you see the ui-select bug in the console as well?

@spadgett I'm having trouble viewing the 'create from url' (which has a ui-select) in console.

dtaylor113:
> I think it might be better to have the existing projects list, then a + button to the right which simply pops up the create project panel (like in the projects summary panel). That would remove the redundancy of creating a project panel here.

@jeff-phillips-18 Sorry, I don't follow.  Are you just agreeing with the design here?

src/components/select-project/select-project.controller.ts
dtaylor113: Fixed

src/components/select-project/selectProject.html
spadgett: For all inputs. Let's use consistent input sizes on this form.

spadgett:
> That is the standard which the Project Summary uses and <create-project> dialog uses. I can remove it, but it won't be consistent with other usages.

It's more important to be consistent with fields on the same form. They don't have the `input-lg` class, placeholder text, or field help currently. The new project form will show up in the dialog from #78 and also alongside service instance parameters when we add them.

If we decide we want help and placeholders, I think we need to go through and add them everywhere.

jeff-phillips-18:
I'm kind of disagreeing with the design. I'm suggesting take the 'Create new project' out of the dropdown choices. Add a + button to the right of the combo box which when clicks the + button we simply show the create project bubble to allow the user to create a project (just like the Create New Project button in the sidebar). Once created, auto-select the project in the list.

spadgett:
I personally like that it all feels like one action to the user. Having multiple forms and dialogs on top of dialogs doesn't seem as clean to me, even if implementation is harder this way.

spadgett:
@dtaylor113 Thanks for your patience with all my comments. I really like the functionality.

jeff-phillips-18:
Maybe I'd like it better if the form didn't show up only after selecting "Create Project". I feel that without the form the page is very empty.

spadgett:
> I feel that without the form the page is very empty.

Yeah, I agree, and I mentioned this to @serenamarie125 as well. Eventually we'll have parameters here, too, but only for service classes that use them.

dtaylor113:
> It's more important to be consistent with fields on the same form. They don't have the input-lg class, placeholder text, or field help currently.

@spadgett Agree to remove ` input-lg ` so all form fields look the same.  I'd rather not remove the placeholder or help text as I believe some thought went into making the form easy to use.  I vote for adding placeholders and help text (if appropriate) to the other forms fields.

src/components/order-service/order-service.controller.ts
```diff
@@ -94,7 +96,44 @@ export class OrderServiceController implements angular.IController {
...
+        description: newProjDesc
+      };
+      this.DataService
+        .create('projectrequests', null, projReqObj, this.$scope)
+        .then( (data: any) => {
+          this.createService();
+        }, (result: any) => {
+          var data = result.data || {};
+          if (data.reason === 'AlreadyExists') {
+            this.ctrl.nameTaken = true;
```
spadgett: Hey @dtaylor113 I was thinking since we already have the list of projects, can we check name taken before trying to create the project? We had a directive that does something like this in origin-web-console.

https://github.com/openshift/origin-web-console/blob/e7a0c0a8f703d5429f70b78223abb31856a66670/app/scripts/directives/oscUnique.js

Doesn't have to be in this PR

src/components/select-project/select-project.controller.ts
dtaylor113: Fixed.

src/components/order-service/order-service.controller.ts
spadgett: You'll still need to check the return status as well since you probably can't see all projects.

dtaylor113: Sounds good.  #88 so we don't loose track of it.

spadgett:
@dtaylor113 I'm seeing the same visual problems as you. I think you might be missing the ui-select CSS.

dtaylor113:
I think I've addressed all review issues except the css/styling one, which I'm going to look into some more.   There might be some lingering design considerations, which we should file issues for.  -thanks

src/components/select-project/selectProject.html
```diff
@@ -0,0 +1,93 @@
+<form class="form-horizontal config-form">
```
spadgett: Needs to be `ng-form` since you can't nest forms in HTML.

```html
<ng-form class=form-horizontal config-form">
```

src/components/select-project/selectProject.html
```diff
@@ -0,0 +1,93 @@
...
+                  <span ng-if="project.metadata.name">&ndash;</span>
+                  <span ng-bind-html="project.metadata.name | highlightKeywords : $select.search"></span>
+                </span>
+          </ui-select-choices>
+        </ui-select>
+      </div>
+    </div>
+</form>
+
+<ng-form name="$ctrl.forms.createProjectForm" class="form-horizontal"
```
spadgett: Just have one `ng-form` for the whole component instead of two.

src/components/select-project/selectProject.html
```diff
@@ -0,0 +1,93 @@
...
+      ng-if="$ctrl.isNewProject()">
+  <fieldset>
```
spadgett: `fieldset` shouldn't be needed since we removed `disableInputs`. That's why it was there.

src/components/order-service/order-service.controller.ts
```diff
@@ -94,7 +96,44 @@ export class OrderServiceController implements angular.IController {
...
+  public invalidNewProject() {
+    // test whether selectedProject is actually a newProject obj. and if passed name validation
+    return this.isNewProject() && !this.ctrl.selectedProject.validName;
+  }
+
   public provisionService() {
+    if (this.isNewProject()) {
+      // the selectedProject is actually a newProject object
+      let newProjName = this.ctrl.selectedProject.metadata.name;
+      let newProjDisplayName = this.ctrl.selectedProject.metadata.annotations['new-display-name'];
```
spadgett: Can we just use the normal annotation for display name `openshift.io/display-name`? It looks like you use the normal description annotation already.

src/components/order-service/order-service.controller.ts
```diff
@@ -111,40 +150,25 @@ export class OrderServiceController implements angular.IController {
...
-      this.ctrl.selectedProject = _.first(this.ctrl.projects);
-    });
-  }
-
   private makeServiceInstance() {
     let serviceClassName = _.get(this, 'ctrl.serviceClass.resource.metadata.name');
+
     let serviceInstance = {
       kind: 'Instance',
-      apiVersion: 'servicecatalog.k8s.io/v1alpha1',
```
spadgett: Why did you take out the api version? That's required

dtaylor113:
> @dtaylor113 I'm seeing the same visual problems as you. I think you might be missing the ui-select CSS.

When I inspect the ui-select dropdown I do see `ui-select` classes:

```
.ui-select-bootstrap .ui-select-match-text span {
    display: inline-block;
    width: 100%;
    overflow: hidden;
}

catalogs-app.css:1611
* {
    box-sizing: border-box;
}
```

spadgett:
> When I inspect the ui-select dropdown I do see ui-select classes

I think it will look OK in origin-web-console with these styles:

https://github.com/openshift/origin-web-console/blob/master/app/styles/_select.less

src/components/select-project/selectProject.html
dtaylor113: The top form is selectProject whereas the bottom form is createProject.  Did change both to `ng-form`

src/components/order-service/order-service.controller.ts
dtaylor113: In the case where a user is creating a new project, the 'selected project' is the createNewProjObj used in the dropdown which has it's `openshift.io/display-name` set to 'Create Project'.  'new-display-name' contains user input for the new proj's displayName.

src/components/select-project/selectProject.html
dtaylor113: Fixed

src/components/select-project/selectProject.html
dtaylor113: Fixed

src/components/order-service/order-service.controller.ts
dtaylor113: whoops! sorry, restored

spadgett: Thanks :)

spadgett:
@dtaylor113 As a follow on, can you pull _select.less into common so the ui-select looks correct outside origin-web-console?

spadgett:
Yup, thanks :)

src/styles/ui-select.less
```diff
@@ -0,0 +1,75 @@
+/* Fix some styles with the angular-ui-select bootstrap theme that isn't right when combined with all of our styles */
+.ui-select-bootstrap {
```
spadgett: Let's actually put this in origin-web-common so we don't have copies of it.

dtaylor113: Ok, I removed it for now and will follow up with a PR after moving it into common.

spadgett: Thanks!

PR #93, opened 2017-04-07
Fixes #87 - The <select-project> component shows offset highlight when selecting a project

### LINKED ISSUE #87: The <select-project> component shows offset highlight when selecting a project
In the <select-project> component seeing a highlight issue when selecting a project:

image

It may be a global console bug.   Root cause maybe because we are using a `<ui-select>` within a `<form>`.

dtaylor113:
Changed semantic versioning to retrieve all origin-web-common 0.0 minor versions
This picks up origin-web-common@0.0.11 which contains the ui-select styles for the `<select-project>` component.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -18,7 +18,7 @@
     "patternfly": "3.21.0",
     "jquery": "2.1.4",
     "lodash": "3.10.1",
-    "origin-web-common": "^0.0.10"
+    "origin-web-common": "0.x"
   },
   "devDependencies": {
     "ace-builds": "1.2.2",
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "bootstrap": "^3.3.6",
     "font-awesome": "^4.6.1",
     "jquery": "~2.1.4",
-    "origin-web-common": "^0.0.10",
+    "origin-web-common": "0.x",
     "paper-css": "^0.1.2",
     "patternfly": "3.20.0",
     "urijs": "1.18.0"


bower.json
```diff
@@ -18,7 +18,7 @@
     "patternfly": "3.21.0",
     "jquery": "2.1.4",
     "lodash": "3.10.1",
-    "origin-web-common": "^0.0.10"
+    "origin-web-common": "~0.x"
```
spadgett: Should this just be

```
"origin-web-common": "0.0.x"
```

jeff-phillips-18: Based on the comment of the PR yes, but maybe we should just pickup any minor version changes.

spadgett: OK, I think it could just be `0.x` then. Shouldn't need the tilde.

dtaylor113: > Should this just be
> 
> "origin-web-common": "0.0.x"

I initially tried that,  it doesn't seem to work.  Maybe something with '0.0'? '1.0.x' works as expected.  I've tried it here: https://semver.npmjs.com/

dtaylor113:
> OK, I think it could just be 0.x then. Shouldn't need the tilde.

'0.x' works, I'll change to that.

bower.json
dtaylor113: > OK, I think it could just be 0.x then. Shouldn't need the tilde.

'0.x' works, I'll change to that.

jeff-phillips-18:
there could be a snafu with this. Currently we change common and don't worry about backwards compatibility. With this we could be broken if a change is made to common that is not backwards compatible (such as https://github.com/openshift/origin-web-common/pull/29).

Are we OK with this issue? Typically when we change common we make the change to catalog pretty quickly (I have a fix in hand for the aforementioned issue).

spadgett:
In practice, it's only an issue we start making catalog releases that will run outside of the console. The console locks its dependencies to known good versions. We could start being more strict at that time.

We're also fibbing because the catalog won't work with common 0.0.1, but technically 0.x matches that.

jeff-phillips-18:
Agreed, I just wanted to make sure folks were aware.

spadgett:
@jwforres OK with this change?

jwforres:
thats fine for now

PR #94, opened 2017-04-07
Add watcher to determine service provision completion.

jeff-phillips-18:


src/components/order-service/order-service.controller.ts
```diff
@@ -183,4 +190,18 @@ export class OrderServiceController implements angular.IController {
...
+    this.watches.push(this.DataService.watchObject(resource, data.metadata.name, context, (instanceData: any, action: any) => {
+      var conditions: any = _.get(instanceData, 'status.conditions');
+
+      this.ctrl.orderComplete = _.find(conditions, {type: "Ready"});
+      var errorCondition: any = _.find(conditions, {status: "False"});
+
+      this.ctrl.error = errorCondition ? errorCondition.data : _.find(conditions, {type: "ProvisionFailed"});
+
+      this.ctrl.error = {
+        Message: "This is a fake failure."
```
spadgett: Looks like you left some debugging code here

jeff-phillips-18: Bah!  My bad...

src/components/order-service/order-service-review.html
```diff
@@ -1,44 +1,36 @@
...
-        View Project
-      </a>
+    <div ng-if="$ctrl.orderComplete">
+      <h3 class="mar-top-none">
+        <strong>{{$ctrl.serviceName}}</strong> has been added to <strong>{{$ctrl.selectedProject | displayName}}</strong> successfully
+      </h3>
+      <div class="alert alert-info mar-top-xxl mar-bottom-xl">
+        <span class="pficon pficon-info" aria-hidden="true"></span>
+        <span class="sr-only">Info</span>
+        Continue to your project to bind this service to your application. Binding this service creates a sectret containing the information necessary for your application to use the service.
```
jwforres: typo, secret

src/components/order-service/order-service.controller.ts
```diff
@@ -183,4 +190,14 @@ export class OrderServiceController implements angular.IController {
   private isNewProject(): boolean {
     return this.ctrl.selectedProject && !_.has(this.ctrl.selectedProject, 'metadata.uid');
   }
+  private watchResults = (resource: any, data: any, context: any) => {
+    this.watches.push(this.DataService.watchObject(resource, data.metadata.name, context, (instanceData: any, action: any) => {
+      var conditions: any = _.get(instanceData, 'status.conditions');
+
+      this.ctrl.orderComplete = _.find(conditions, {type: "Ready"});
+      var errorCondition: any = _.find(conditions, {status: "False"});
```
jwforres: the existence of the Ready condition doesn't mean the order is complete.  There will be a point in time where Ready will have status False, not because it has failed, but because it is still provisioning.  I would definitely not assume the existence of a False condition means that there is an error.

jeff-phillips-18: So, is the only real error condition when we get a type of "ProvisionFailed" ?

jwforres: @pmorie what's the current "perm failed" for provision, assuming thats the condition of type ProvisionedFailed that jeff is referencing?

@jeff-phillips-18 I doubt we have anything else to check for right now as far as conditions.  What you could have is an immediate failure when you make the provision request, but that would be in the error handler of the DataService.create

PR #96, opened 2017-04-07
Add background image

cshinn:
Will not work in console. Temporary solution

screen shot 2017-04-07 at 1 09 55 pm


src/styles/variables.less
```diff
@@ -3,10 +3,11 @@
 @import '~patternfly/dist/less/variables.less';
 
 @landing-main-bg:                                     #3ca2c9;
+@landing-image:                                       './assets/upstream_bg.png';
```
spadgett: Suggest a name other than "upstream_bg.png". Maybe "landing_page_bg.png"

spadgett:
@jeff-phillips-18 Do you know what needs to happen to include this in the dist?

jeff-phillips-18:
We could copy the png to the dist directory pretty easily, but then console needs to do something with that. Is that enough?

spadgett:
@rhamilto suggested we base64 encode the image so we can include it directly in the less

rhamilto:
> @rhamilto suggested we base64 encode the image so we can include it directly in the less

Ideally as a less var so it can be easily changed.

rhamilto:
@cshinn, Let's use this JPG instead (171 KB instead of 1.6 MB)
landing_page_bg

rhamilto:
PR looks good, but I suggest changing https://github.com/cshinn/origin-web-catalog/blob/fa81cfe3ed7628eb037339cd360fc8df9bd9d644/src/styles/landing-page.less#L17 to #043d67

screen shot 2017-04-07 at 2 25 20 pm

rhamilto:
And please use a new variable so we can easily change that color.

cshinn:
@rhamilto border color's been updated

PR #97, opened 2017-04-07
Fix search bar at top of page

cshinn:
Fixes search bar to the top of the page, also addresses small visual bug in search dropdown. Width: inherit does not work in this case because of the  `<catalog-search>` item

screen shot 2017-04-07 at 2 56 13 pm 2


rhamilto:
@cshinn, .landing-search-area needs a background color and a z-index: @zindex-navbar-fixed to avoid the scrolling contents changing behind it/over it.

screen shot 2017-04-07 at 1 34 30 pm

cshinn:
@rhamilto added the z-index, but I'm not sure how to best deal with the background image. Should I add a solid color for now? Should probably get visuals input

rhamilto:
@cshinn, you're going to need to duplicate the bg-color and image from #96 in that area (they'll overlap one another).

cshinn:
Search bar area now uses background image to cover page content

PR #99, opened 2017-04-07
Change use of deleteLink directive to deleteProject directive

jeff-phillips-18:


spadgett:
@jeff-phillips-18 looks like you have some test failures

jeff-phillips-18:
Failure is due to not pulling the latest origin-web-common. I'll have it clean dependencies during travis build

PR #100, opened 2017-04-07
Fix the create project form inputs on mobile

spadgett:
Fixes the following issue where the inputs are cut off.

screen shot 2017-04-06 at 2 10 03 pm


spadgett:
Seeing a large diff in dist that I need to figure out.

spadgett:
Travis seems happy, so I guess it's OK

PR #102, opened 2017-04-07
Initial cut of subcategory area

rhamilto:


src/styles/main.less
```diff
@@ -1,3 +1,6 @@
+@import '~bootstrap/less/variables.less';
+@import '~patternfly/dist/less/color-variables.less';
+@import '~patternfly/dist/less/variables.less';
 @import './variables.less';
```
jeff-phillips-18: Adding these here, we should remove them from variables.less

rhamilto: Good catch, @jeff-phillips-18.  Those slipped back in while rebasing.

spadgett:
@jwforres @jeff-phillips-18 Robb needs to revert my font-icon change, it's causing problems.

PR #106, opened 2017-04-07
Go to project page when view project clicked after ordering a service

jeff-phillips-18:
Also put back specification of origin-web-common version to clear up dependency issues at build time.


jeff-phillips-18:
Yeah, I thought the same for the url.

What about search won't work?

spadgett:
It has a separate order-service panel currently, which won't get the base URL value. We need to refactor that anyway.

PR #122, opened 2017-04-10
Correctly show service class description

### LINKED ISSUE #121: Not showing service class description
@jwforres FYI, looks like we only look in `osbMetadata` for the description.

https://github.com/spadgett/origin-web-catalog/blob/2df69ab49c73dce99a433c2060b3648bf270352d/src/services/catalog.service.ts#L130

screen shot 2017-04-10 at 8 58 19 am

spadgett:
Fixes #121

### DIFF
--- a/src/services/catalog.service.ts
+++ b/src/services/catalog.service.ts
@@ -127,7 +127,7 @@ export class ServiceItem implements IServiceItem {
   }
 
   private getDescription() {
-    return _.get(this.resource, 'osbMetadata.description', '');
+    return _.get(this.resource, 'description', '');
   }
 
   private getLongDescription() {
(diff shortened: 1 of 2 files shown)


src/services/catalog.service.ts
```diff
@@ -127,7 +127,7 @@ export class ServiceItem implements IServiceItem {
   }
 
   private getDescription() {
-    return _.get(this.resource, 'osbMetadata.description', '');
+    return _.get(this.resource, 'description', '');
```
dtaylor113: Thought of this after the merge, should I update the mock data and move description from under osbMetadata?

spadgett: Yeah, I didn't realize the mock data was wrong. It should be a top-level item, not under `osbMetadata`.

PR #126, opened 2017-04-10
Default to `project/` for base project URL

spadgett:
Default to relative URL `project/` instead of `/` for the default base URL. This will work correctly when running in the web console.


src/filters/projectUrl.ts
```diff
@@ -1,7 +1,7 @@
 export function projectUrlFilter () {
```
dtaylor113: Should this be in 'origin-web-common'?

PR #128, opened 2017-04-10
Hide the "advanced options" link for new projects

spadgett:
The link will be broken for new projects since the project hasn't been created yet.


src/components/create-from-builder/create-from-builder-configure.html
```diff
@@ -88,7 +88,8 @@
       </div>
     </div>
 
-    <div ng-if="$ctrl.selectedProject && $ctrl.istag" class="form-group">
+    <!-- Only show the link for existing projects. It will be broken for new projects. -->
+    <div ng-if="$ctrl.selectedProject.metadata.uid && $ctrl.istag" class="form-group">
```
jwforres: yuck :(  ok for now, but we need to come up with a better solution long term

Issue #124, opened 2017-04-10
Large amount of whitespace sometimes appears at bottom of "All" category


spadgett:
![screen shot 2017-04-10 at 9 38 01 am](https://cloud.githubusercontent.com/assets/1167259/24864208/78b95a00-1dd1-11e7-88d8-8204d0569645.png)

cc @rhamilto

Issue #91, opened 2017-04-06
.build-applications-view layout issues in IE11


rhamilto:
<img width="1107" alt="screen shot 2017-04-06 at 4 43 59 pm" src="https://cloud.githubusercontent.com/assets/895728/24774831/6f039a5c-1ae8-11e7-9ee1-e8b62739f9b3.PNG">

I haven't investigated, but I would wager a [flexbug](https://github.com/philipwalton/flexbugs) or two are to blame.

fyi:  @cshinn

PR #129, opened 2017-04-10
Fixing various cosmetic issues

### LINKED ISSUE #84: Ensure scrollbars properly indicate scrollable regions
With #83, the sidebar's positioning was fixed.  As a result of that positioning and the move of the scrollbar to the inner contents of the page, it's not immediately obvious that both the sidebar and the inner page contents are scrollable.  Given that a combination of browser, OS, and OS settings (e.g., macOS) can result in the scrollbars rendering differently, we ideally want to find a solution that normalizes these and clearly communicates independent scrollable regions.

rhamilto:
Fixes #84, #91, #124

Tweaks to saas styles

Improves alignment of fa-cubes, fa-question, and icon-ruby when inside hover circle

As seen in Firefox on Fedora

screen shot 2017-04-10 at 2 20 54 pm


src/styles/main.less
```diff
@@ -1,6 +1,7 @@
 @import '~bootstrap/less/variables.less';
 @import '~patternfly/dist/less/color-variables.less';
 @import '~patternfly/dist/less/variables.less';
+@import "../../app/styles/variables.less";
```
spadgett: We should be moving the variables we need under `src`. We shouldn't import from `app` to `src` since the app directory is meant to only be for the sample app.

src/styles/variables.less
```diff
@@ -4,14 +4,18 @@
...
 @landing-main-bg:                                     #3ca2c9;
+@landing-offset-top:                                  (@landing-search-area-height + @navbar-height);
+@landing-search-area-height:                          52px;
 @landing-side-bar-bg:                                 #2C343D;
 @landing-side-bar-top-offset:                         72px;
 @landing-side-bar-width-lg:                           425px;
 @landing-side-bar-width-md:                           375px;
 @landing-side-bar-width-sm:                           325px;
-@saas-hover:                                          fade(@color-pf-blue-700, 60%);
+@navbar-height: 72px;
```
spadgett: nit: indentation :)

spadgett:
@jwforres @jeff-phillips-18 might be good to cut a release after this goes in since this PR and #127 had downstream breaking changes. Then we can bump origin-web-console

rhamilto:
[test]

PR #132, opened 2017-04-10
Project Summary: Recently Viewed Items

dtaylor113:
image

- Also updated serviceClass's mock data
- Remaining issue, see #131


src/components/projects-summary/projects-summary.controller.ts
```diff
@@ -85,6 +86,10 @@ export class ProjectsSummaryController implements angular.IController {
         nextResource.href = this.Constants.HELP_BASE_URL + this.Constants.HELP[nextResource.help];
       }
     });
+
+    this.$scope.$watch(this.getgetRecentlyViewedItemsFromStorage, (newValue: any) => {
```
spadgett: I'm not sure what the overhead of `localStorage.getItem` is, but `$watch` is evaulated a lot. Can we just update the recent items on init and when the user clicks on one? We could use `$rootScope.$emit` when the services view updates the list and `$rootScope.$on` to handle it here.

dtaylor113:
@serenamarie125

src/components/services-view/services-view.controller.ts
```diff
@@ -180,4 +181,35 @@ export class ServicesViewController implements angular.IController {
...
+    });
+
+    // limit to 3 max.
+    if (recentlyViewed.length >= 3) {
+      recentlyViewed.pop();
+    }
+
+    // convert back to strings for local strorage
+    recentlyViewed = _.map(recentlyViewed, (item: any) => {
+      return JSON.stringify(item);
```
spadgett: I'd rather just store the UID and not the whole item. Then we can pass the service classes to the project summary indexed by UID.

This could result in stale display names or icons if the service class changes. It also could result in errors if a service class was deleted after you looked at it.

src/styles/services-view.less
```diff
@@ -1,6 +1,7 @@
 .services-view {
   .services-item {
     color: @text-color;
+    cursor: pointer;
```
spadgett: You shouldn't need this if you make the element a link. Otherwise it's not keyboard accessible.

test/landing-page.spec.ts
```diff
@@ -57,7 +57,7 @@ describe('landingPage', () => {
     var testSearchArea = jQuery(headerArea).find('#testSearch');
     expect(testSearchArea.length).toBe(1);
 
-    var headerArea = jQuery(element).find('.landing-header-area');
+    headerArea = jQuery(element).find('.landing-header-area');
```
spadgett: Oops, do you mind fixing `headerArea` above to be `searchArea`? This is my bug.

app/mockServices/mockData/services.ts
```diff
@@ -4,10 +4,10 @@ export const servicesData = {
       "name": "test-serviceclass-java-nodejs",
       "uid": "1",
     },
+    description: 'BUILDS SOURCE CODE',
```
spadgett: Thanks for fixing this.

src/components/projects-summary/projects-summary.html
```diff
@@ -88,4 +101,25 @@ <h2 class="secondary">Resources</h2>
       </ul>
     </div>
   </div>
+  <div ng-if="$ctrl.recentlyViewedItems && $ctrl.recentlyViewedItems.length > 0">
+    <h2 class="primary">Recently Viewed</h2>
+    <div class="services-view">
+      <div class="services-view-container">
+        <div class="services-items-inner">
+          <div class="services-items-inner-shadow-covers">
+            <div class="services-item" ng-repeat="item in $ctrl.recentlyViewedItems track by item.resource.metadata.uid"
```
spadgett: Make this an `<a href="">` element instead of a `div` so that it is keyboard accessible.

src/components/services-view/services-view.controller.ts
```diff
@@ -109,6 +109,7 @@ export class ServicesViewController implements angular.IController {
       this.ctrl.selectedImageStream = null;
       this.ctrl.selectedServiceClass = item;
     }
+    this.setRecentlyViewed();
```
spadgett: This won't catch items viewed from search.

src/components/services-view/services-view.controller.ts
```diff
@@ -180,4 +181,35 @@ export class ServicesViewController implements angular.IController {
   private updateActiveCardStyles() {
     this.$timeout(this.resizeExpansion, 50);
   }
+
+  private setRecentlyViewed() {
```
spadgett: It's probably worth having a recently viewed service for managing this. Then we can add items from other places (like search) and the projects summary can get items without having to know that they're kept in localStorage.

src/components/projects-summary/projects-summary.controller.ts
```diff
@@ -144,10 +149,43 @@ export class ProjectsSummaryController implements angular.IController {
...
+  private getRecentlyViewedItems(recentlyViewed: any) {
+    recentlyViewed = recentlyViewed ? JSON.parse(recentlyViewed) : [];
+    recentlyViewed = _.map(recentlyViewed, (item: any) => {
+      return JSON.parse(item);
+    });
+
+    return recentlyViewed;
+  }
+
+  private getgetRecentlyViewedItemsFromStorage() {
```
dtaylor113: Did I stutter!?  :-)  sorry, I fix that

dtaylor113:
Hi, closing this PR as a number of architectural issues were raised.  I'll open another one when ready.

src/components/services-view/services-view.controller.ts
spadgett: In case it's not clear, I mean an AngularJS service like `RecentlyViewedItemsService`

PR #134, opened 2017-04-11
Add guided tour component and showcase usage

jeff-phillips-18:
Design: http://openshift.github.io/openshift-origin-design/web-console/1-homepage/project-panel#getting-started---guided-tour

Sample of tour bubble:
image

Note: Start Guided Tour button (and 'Getting Started' title) are hidden on mobile. Guided tour will not work on mobile.

@serenamarie125 @beanh66

### DIFF
--- a/app/pages/home/home-page.html
+++ b/app/pages/home/home-page.html
@@ -11,6 +11,6 @@
     <services-view service-classes="$ctrl.serviceClasses" image-streams="$ctrl.imageStreams" base-project-url="{{$ctrl.baseProjectUrl}}"></services-view>
   </landingbody>
   <landingside>
-    <projects-summary base-project-url="{{$ctrl.baseProjectUrl}}" projects-url="{{$ctrl.projectsUrl}}"></projects-summary>
+    <projects-summary start-getting-started-tour="$ctrl.startGuidedTour" base-project-url="{{$ctrl.baseProjectUrl}}" projects-url="{{$ctrl.projectsUrl}}"></projects-summary>
   </landingside>
 </landing-page>
--- a/app/services/landingPageTourService.ts
+++ b/app/services/landingPageTourService.ts
@@ -1,0 +1,119 @@
+import * as angular from 'angular';
+import * as jQuery from 'jquery';
+
+interface ILandingPageTourService {
+  startTour(tourConfig: any);
+}
+
+export class LandingPageTourService implements ILandingPageTourService {
+  public static $inject = ['$document', '$timeout', 'Constants', 'GuidedTourService'];
+
+  private $document: any;
+  private $timeout: any;
+  private Constants: any;
+  private GuidedTourService: any;
+  private tourConfig: any;
+
+  constructor ($document: any, $timeout: any, Constants: any, GuidedTourService: any) {
+    this.$document = $document;
+    this.$timeout = $timeout;
+    this.Constants = Constants;
+    this.GuidedTourService = GuidedTourService;
+
+    this.tourConfig = {
+      id: "getting-started-tour",
+      steps: []
+    };
+
+    let stepConstants = Constants.LANDING_PAGE_TOUR_STEPS || {};
+
+    this.addStep(stepConstants.createProject,
+      {
+        title: "Create Project",
+        content: "Projects allow you to organize and manage your content. Projects require a unique name, and optionally can include a display name, and description",
+        target: ".catalog-projects-summary-panel .create-button",
+        targetScrollElement: '.landing-side-bar',
+        placement: "left"
+      });
+
+    this.addStep(stepConstants.search,
+      {
+        title: "Search Resources",
+        content: "Search by name, description, keyword, or label to quickly locate items in the catalog that you want to add to your project.",
+        target: ".landing-search-area .landing-search-form",
+        placement: "bottom",
+        fixedElement: true
+      });
+
+    this.addStep(stepConstants.saasList,
+      {
+        title: "Build Applications",
+        content: "Build applications from a quickstart or popular framework.",
+        target: ".saas-list .card",
+        targetScrollElement: '.landing',
+        placement: "right",
+      });
+
+    this.addStep(stepConstants.catalogs,
+      {
+        title: "Browse Catalog",
+        content: "If you don’t know exactly what you are looking for, you can browse all available catalog items under the first tab in the catalog.",
+        target: "#category-all",
+        targetScrollElement: '.landing',
+        placement: "top",
+        preShow: this.showAllCatalogs
+      });
+
+    this.addStep(stepConstants.category,
+      {
+        title: "Browse by Category",
+        content: "A secondary level of categorization is available to further narrow your search.",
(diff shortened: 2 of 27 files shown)


app/pages/home/home-page.html
```diff
@@ -11,6 +12,6 @@
     <services-view service-classes="$ctrl.serviceClasses" image-streams="$ctrl.imageStreams" base-project-url="{{$ctrl.baseProjectUrl}}"></services-view>
   </landingbody>
   <landingside>
-    <projects-summary base-project-url="{{$ctrl.baseProjectUrl}}" projects-url="{{$ctrl.projectsUrl}}"></projects-summary>
+    <projects-summary start-getting-started-tour="$ctrl.startGuidedTour" base-project-url="{{$ctrl.baseProjectUrl}}" projects-url="{{$ctrl.projectsUrl}}"></projects-summary>
```
dtaylor113: nit.  probably could just use 'start-tour'

package.json
```diff
@@ -38,6 +38,7 @@
     "font-awesome": "^4.6.1",
     "jquery": "~2.1.4",
     "origin-web-common": "~0.0.12",
+    "hopscotch": "~0.2.7",
```
dtaylor113: Not for this PR, but at some point we should discuss why/if we need to define dependencies in both NPM and Bower.

dtaylor113:
Nit.  Arrow seem to point a little off target:

image

dtaylor113:
LGTM!  Looks great!!

serenamarie125:
screen shot 2017-04-11 at 10 11 34 am

screen shot 2017-04-11 at 10 11 24 am

beanh66:
@serenamarie125 Good point. @jeff-phillips-18 can you replace masthead with "top banner" in step 7?

serenamarie125:
@jeff-phillips-18 ultimately this needs to be configurable by the customer
- title & description should be modifiable on premise
- should be able to disable access on premise

@sspeiche  @jwforres did i miss any other requirements?

spadgett:
If we make the tour JSON a value in constants.js, it could be customized through an extension

serenamarie125:
@jeff-phillips-18 I think this is what @jwforres and @spadgett were referring too.  the Title looks to be a pixel too high :eye_speech_bubble: 
screen shot 2017-04-11 at 4 43 08 pm

serenamarie125:
Does anyone else this that the top corner juts out too far?  I actually think it's an optical illusion :joy: but curious what others think @jeff-phillips-18 

screen shot 2017-04-11 at 10 11 34 am

serenamarie125:
another nit ... #2 says Search Resources, but it should be Search Catalog @jeff-phillips-18

jwforres:
I see it too, it is an illusion but I think it's caused by overlapping
borders between the box border and the caret border creating a darker
angled line 1px thick.

jeff-phillips-18:
This is ready for review. I believe I've addressed previous comments.

jeff-phillips-18:
Update bubbles:
image

image

image

image

beanh66:
@jeff-phillips-18 Thanks for the updates, I think the arrows look way better! The bubble placement changes look good to me as well.

app/services/landingPageTourService.ts
```diff
@@ -0,0 +1,120 @@
...
+    this.$timeout = $timeout;
+    this.Constants = Constants;
+    this.GuidedTourService = GuidedTourService;
+
+    this.tourConfig = {
+      id: "getting-started-tour",
+      steps: []
+    };
+
+    let stepConstants = Constants.LANDING_PAGE_TOUR_STEPS || {};
```
spadgett: It looks like the config options now are only to hide or modify existing steps?

I think I liked it better before where it was simply a JSON object, and then in an extension you could completely change the entire tour if you want. We could add in an `id` attribute to the step object if we want people to be able find an individual step without hard-coding an array index.

src/services/guided-tour.service.ts
```diff
@@ -0,0 +1,141 @@
...
+  private innerConfig: any;
+  private bubbleHeight: number = 175;
+
+  constructor ($document: any) {
+    this.$document = $document;
+  }
+
+  public startTour(tourConfig: any) {
+    var backdropElement = '<div id="guided_tour_backdrop" class="modal-backdrop fade guided-tour-backdrop"></div>';
+    $(this.$document).find('body').append(backdropElement);
```
spadgett: Can't this just be...?

```js
$('body').append(backdropElement);
```

src/services/guided-tour.service.ts
```diff
@@ -0,0 +1,141 @@
...
+import * as hopscotch from 'hopscotch';
+import * as $ from 'jquery';
+import * as _ from 'lodash';
+
+interface IGuidedTourService {
+  startTour(tourConfig: any);
+  cancelTour();
+}
+
+export class GuidedTourService implements IGuidedTourService {
```
spadgett: We might want to eventually push this into common so that we can reuse it for other console pages.

app/services/landingPageTourService.ts
jeff-phillips-18: There are two pieces that make it difficult for steps to be added. The target field which specifies a jquery selector to find the target component, and any actions that need to be performed prior to showing the step (ie. open the Languages tab in the catalog).

src/services/guided-tour.service.ts
jeff-phillips-18: Agreed. Should I do that at this point?

src/services/guided-tour.service.ts
jeff-phillips-18: Yes.

app/services/landingPageTourService.ts
spadgett: I don't think it's unreasonable asking people to write some code if they want to do anything advanced like clicking tabs, but yeah custom tours can break if we change the page structure or class names. So I'm on the fence.

@jwforres Thoughts?

I have a feeling admins who customize these will want to be able to completely replace the tours.

jeff-phillips-18: How would they add code? Define a function in their config file? They would have to be intimate with the page structure and/or class names for that as well.

spadgett: Yes, an inline function in the tour JSON probably. We already support something like it for left nav with `isValid`.

https://docs.openshift.org/latest/install_config/web_console_customization.html#project-left-navigation

> They would have to be intimate with the page structure and/or class names for that as well.

If they want to completely replace the tour, this is true. I think the biggest problem is that it could break on upgrade if we change things.

jeff-phillips-18: I can completely move the tour config to the constants file to allow it to be completely over-ridden. I can define the functions that are currently used there as well. 

When we add this to console, we could put all of this into the config.js file so there is a starting point for folks to go off of. 

We could also make some defines for selectors of different components in the landing page if we feel that would be helpful.

jeff-phillips-18:
Moving GuidedTourService to the origin-web-common repo. Closing this PR. Will open a new PR with the adjustments.

PR #135, opened 2017-04-11
Adding styling for images in panel overlay, improving overlay markup/css

rhamilto:
font icon
screen shot 2017-04-11 at 9 38 47 am

img
screen shot 2017-04-11 at 9 39 07 am

PR #138, opened 2017-04-11
Improve selected plan display in order-service dialog

spadgett:
Show plan descriptions in the order service left panel for selected plans. Only show the plan name if there's a display name.

This avoids seeing things like "Plan default" with no description for template broker items or cryptic k8s names with no plan description.


spadgett:
I'll rebase on #135 when it merges. That PR should go in first

jeff-phillips-18:
Sorry, I missed that rebase comment.

PR #141, opened 2017-04-11
Enlarge saas icons and titles on large screens and enable images

cshinn:
screen shot 2017-04-11 at 1 44 34 pm
screen shot 2017-04-11 at 1 44 39 pm


src/components/saas-list/saas-list.html
```diff
@@ -3,7 +3,7 @@ <h1 ng-if="$ctrl.saasTitle">{{$ctrl.saasTitle}}</h1>
   <div class="saas-list" items="$ctrl.saasOfferings">
     <div class="card" ng-repeat="item in $ctrl.saasOfferings">
       <a href="" class="card-content" ng-click="$ctrl.handleClick(item)">
-        <span class="card-icon {{item.icon}}"></span>
+        <span class="card-icon {{item.icon}}"><img ng-if="item.image" src="{{item.image}}"></img></span>
```
spadgett: Needs to be `ng-src=` instead of `src=`

spadgett: `img` doesn't need a closing tag in HTML5

rhamilto:
IMO, fa-mobile could be bigger in the first screenshot.  It reads smaller than the others.

serenamarie125:
@rhamilto we will only have 2 items displayed by default, and mobile isn't one of them.  I don't think we need to worry about that here.  Question will be where do we check in actual content for Online versus the others.

spadgett:
We should just remove the special style we have for `fa-mobile` since it's only for the placeholder items

https://github.com/openshift/origin-web-catalog/blob/0d94b6f6f0c2d66cd23d588be61cb91c3aca7bfc/src/styles/saas-list.less#L82-L90

cshinn:
@spadgett I was thinking the same thing.

cshinn:
@rhamilto is this travis failure similar to what you've been running into? I've been checking in the built files to this PR.

jeff-phillips-18:
@cshinn Yeah, once #143 goes in, rebasing with that should clear your travis failure.

rhamilto:
My $0.02?  Use a different icon that doesn't require the override. No point in having a visual defect for the sake of it.

cshinn:
I swapped it out in the most recent set of changes. I'll squash now that this seems to have been resolved.

screen shot 2017-04-11 at 2 37 09 pm

src/styles/saas-list.less
```diff
@@ -67,26 +66,28 @@
       }
 
       .card-icon {
+        align-items: center;
+        display: flex;
         font-size: 24px;
         text-align: center;
 
+        img{
```
rhamilto: Nit:  add space between "img" and "{".

Would be great to put this rule after the media queries rather than before since the media queries are for .card-icon and this is for a separate node.

src/styles/saas-list.less
```diff
@@ -49,10 +48,10 @@
       color: @saas-text;
       display: flex;
       flex: 1 1 0%; // IE 11
+      padding: (@grid-gutter-width / 2);
```
rhamilto: #142 also makes this change, but adds the removal of line 24.  Maybe add the removal of line 24 and we can close #142?

cshinn: Done and done

spadgett:
@cshinn Merge conflict now #135 merged

cshinn:
@rhamilto could you take another look at this?

PR #143, opened 2017-04-11
Removing 'node_modules' from travis cache so builds run properly

rhamilto:


spadgett:
It looks like `npm install` only takes 2-3 minutes, so I'm fine removing it from the cache

screen shot 2017-04-11 at 2 08 33 pm

spadgett:
@jwforres OK with this change?

PR #146, opened 2017-04-12
Project Summary: Recently Viewed Items

dtaylor113:
image
- Opened Issue #131 
- Consolidated `<ordering-panel>`'s
- Updated mockData for easier unit testing


src/components/landing-page/landing-page.html
```diff
@@ -1,4 +1,18 @@
 <div class="landing">
+  <overlay-panel show-panel="$ctrl.orderingPanelVisible" show-close="true" handle-close="$ctrl.closeOrderingPanel">
+    <order-service
+        ng-if="$ctrl.selectedServiceClass"
+        base-project-url="{{$ctrl.baseProjectUrl}}"
```
dtaylor113: Not sure if this is correct.  Previously, `base-project-url `was only in `<catalog-search>`, but not set in `<catalog-search>`'s controller (...at least not in the catalog app?).  I assume this is set in the console.   If so, will it affect` <ordering-panel> `when launched from `<service-view>` and` <project-summary>` in the console app?

spadgett: Yeah it should always be set (and I believe is set in the console).

src/services/recently-viewed.service.ts
```diff
@@ -0,0 +1,40 @@
+export class RecentlyViewedService {
```
jeff-phillips-18: Can we name this a bit more specifically? There may come a time where there are other recently viewed item types.

src/components/projects-summary/projects-summary.controller.ts
```diff
@@ -2,34 +2,44 @@ import * as angular from 'angular';
 import * as _ from 'lodash';
 
 export class ProjectsSummaryController implements angular.IController {
-  static $inject = ['$element', '$scope', '$filter', 'ProjectsService', 'Logger', 'AuthService', 'DataService', 'Constants', 'AlertMessageService'];
+  static $inject = ['$element', '$rootScope', '$scope', '$filter', 'ProjectsService', 'Logger', 'AuthService',
```
spadgett: This list is getting long enough it's probably good to put one item on each line and sort them alphabetically (same for constructor below). I'm worried otherwise it will be too easy to get out of sync with the constructor args.

src/components/projects-summary/projects-summary.controller.ts
```diff
@@ -144,10 +184,33 @@ export class ProjectsSummaryController implements angular.IController {
...
     }
   }
+
+  private getRecentlyViewedItems() {
+    // recentItems is an array of uids
+    let recentItems: any = this.RecentlyViewed.getItems();
+
+    // replace uids with IServiceItems
+    let items: any = _.map(recentItems, (recentItem: any) => {
+      return _.find(this.allItems, (item: any) => {
```
spadgett: Shouldn't need `_.find` since you already indexed the items by UID.

```ts
let items = _.map(recentItems, (uid: string) => {
  return this.allItems[uid];
};
```

src/components/projects-summary/projects-summary.html
```diff
@@ -88,4 +88,25 @@ <h2 class="secondary">Resources</h2>
       </ul>
     </div>
   </div>
+  <div ng-if="$ctrl.recentlyViewedItems && $ctrl.recentlyViewedItems.length > 0">
```
spadgett: This could be just

```html
<div ng-if="$ctrl.recentlyViewedItems.length">
```

src/services/recently-viewed.service.ts
dtaylor113: How about `ReventlyViewedCatalogItems`?

src/components/projects-summary/projects-summary.html
```diff
@@ -88,4 +88,25 @@ <h2 class="secondary">Resources</h2>
...
+    <h2 class="primary">Recently Viewed</h2>
+    <div class="services-view">
+      <div class="services-view-container">
+        <div class="services-items-inner">
+          <div class="services-items-inner-shadow-covers">
+            <a href="" class="services-item" ng-repeat="item in $ctrl.recentlyViewedItems track by item.resource.metadata.uid"
```
spadgett: Better to use the UID filter

```
track by (item.resource | uid)
```

It avoids duplicates in a repeater errors if UID is not set for some reason.

src/services/recently-viewed.service.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  public getItems() {
+    let recentlyViewed: any = localStorage.getItem('catalog-recently-viewed-services');
+    return recentlyViewed ? JSON.parse(recentlyViewed) : [];
+  }
+
+  public addItem(uid: any) {
+    let recentlyViewed: any = this.getItems();
+
+    // if previously viewed, remove from list
```
spadgett: :+1:

src/services/recently-viewed.service.ts
dtaylor113: Fixed.  Changed to `RecentlyViewedServiceItems`

src/components/projects-summary/projects-summary.html
```diff
@@ -88,4 +88,25 @@ <h2 class="secondary">Resources</h2>
...
+                 ng-click="$ctrl.orderService(item)">
+              <div ng-if="!item.imageUrl" class="services-item-icon">
+                <span class="{{item.iconClass}}"></span>
```
```html
<span class="{{item.iconClass}}" aria-hidden="true"></span>
```

src/services/recently-viewed-service-items.service.ts
```diff
@@ -0,0 +1,40 @@
+export class RecentlyViewedServiceItems {
```
spadgett: :+1: Seems much cleaner as a separate service, thanks

src/services/recently-viewed-service-items.service.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  public addItem(uid: any) {
+    let recentlyViewed: any = this.getItems();
+
+    // if previously viewed, remove from list
+    recentlyViewed.map((rvUID, index, array) => {
+      if (rvUID === uid) {
+        array.splice(index, 1);
+      }
+    });
```
spadgett: Easier to use `_.remove`

```ts
_.remove(recentlyViewed, uid);
```

src/services/recently-viewed-service-items.service.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    // limit to 3 max.
+    if (recentlyViewed.length >= 3) {
+      recentlyViewed.pop();
+    }
```
spadgett: Could also be

```ts
recentlyViewed.unshift(uid);
// limit to 3 items
recentlyViewed = _.take(recentViewed, 3);
```

src/services/recently-viewed-service-items.service.ts
spadgett: Sorry I think this actually needs to be

```ts
_.remove(recentViewed, (rvUID) => {
  return rvUID === uid;
});
```

dtaylor113: So three lines instead of one?  IMO splice is better :-)

spadgett: 3 lines instead of 5. We shouldn't be splicing the array in the middle of a map

src/components/projects-summary/projects-summary.controller.ts
dtaylor113: Fixed

src/components/projects-summary/projects-summary.html
dtaylor113: Fixed

src/components/projects-summary/projects-summary.html
dtaylor113: Fixed

src/components/projects-summary/projects-summary.html
dtaylor113: Fixed

src/services/recently-viewed-service-items.service.ts
dtaylor113: Fixed

src/services/recently-viewed-service-items.service.ts
dtaylor113: Ok, Fixed

dtaylor113:
@spadgett , @jeff-phillips-18 - All review issues addressed, thanks

dtaylor113:
Hi @spadgett, when you get a moment, this is waiting on your approval/merge.  -thanks

src/components/projects-summary/projects-summary.controller.ts
```diff
@@ -167,4 +236,20 @@ export class ProjectsSummaryController implements angular.IController {
     $('.catalog-projects-summary-modal-backrop').remove();
   }
 
+  private getRecentlyViewedItems() {
+    // recentItems is an array of uids
+    let recentItems: any = this.RecentlyViewed.getItems();
+
+    // replace uids with IServiceItems
+    let items: any = _.map(recentItems, (uid: any) => {
+      return this.allItems[uid];
```
spadgett: Missed this on previous review, but we might want a null check here because this will be a runtime error if you haven't passed in the items. Do we want to require that for this component?

dtaylor113: Like this:  return `this.allItems && this.allItems[uid];`

dtaylor113: Do you mean like this?:   `return this.allItems ? this.allItems[uid] : null;`

dtaylor113: This method only gets called after serviceClasses and imageStreams have been passed in.

spadgett: It gets called on recently viewed events, and the listener is set up `$onInit`. If I've never passed in image streams or service classes (or they haven't loaded yet), won't this be a runtime error?

I know we expect them to be passed in, but might be better to add a null guard just in case something doesn't load since it's easy to do.

src/components/projects-summary/projects-summary.html
```diff
@@ -88,4 +88,25 @@ <h2 class="secondary">Resources</h2>
       </ul>
     </div>
   </div>
+  <div ng-if="$ctrl.recentlyViewedItems.length">
+    <h2 class="primary">Recently Viewed</h2>
```
jeff-phillips-18: this should be secondary rather than primary

dtaylor113: In the design doc it looks like the same size (perhaps even bigger) than "My Projects" http://openshift.github.io/openshift-origin-design/web-console/1-homepage/project-panel#recently-viewed
Pls see picture under 'Projects List (More than 2)"  @serenamarie125 ?  -thanks

dtaylor113: Ok, it now has:

```
    let items: any = _.map(recentItems, (uid: any) => {
      return this.allItems ? this.allItems[uid] : null;
    });
```

dtaylor113: Actually, going to change this to do the check initially, before iterating on anything:

```
   if (!this.allItems) {
      return;
    }
```

src/components/projects-summary/projects-summary.controller.ts
dtaylor113: Ok, added this to the beginning of the method:
```
   if (!this.allItems) {
      return;
    }
```

src/components/projects-summary/projects-summary.html
```diff
@@ -99,4 +99,25 @@ <h2 class="secondary" ng-if="$ctrl.projects && $ctrl.projects.length > 0">Gettin
       </ul>
     </div>
   </div>
+  <div ng-if="$ctrl.recentlyViewedItems.length">
+    <h2 class="secondary">Recently Viewed</h2>
+    <div class="services-view">
+      <div class="services-view-container">
+        <div class="services-items-inner">
+          <div class="services-items-inner-shadow-covers">
+            <a href="" class="services-item" ng-repeat="item in $ctrl.recentlyViewedItems track by (item.resource | uid)"
```
spadgett: I was talking to @jwforres, and we were thinking it might be good to sort these alphabetically by name here (but still store them by most recently clicked). Then the items don't reorder when you click on one in recently viewed.

dtaylor113: Ok, that is a design question best answered by @beanh66 and @serenamarie125.  I think the design specifically wanted the most recently clicked item to appear left-most in the list.  Pretty sure I specifically asked @beanh66 that exact question.

spadgett: It's really odd to see them shuffle when I click on one of the recently-viewed items.

I think we should delay updating recently-viewed until after the panel opens as well because seeing them move is distracting. But I think that can be a follow-on if @sg00dwin helps with the styling.

spadgett: For example, when I click Python here, Python and WildFly swap places. The musical chairs dance is strange.

screen shot 2017-04-20 at 2 09 07 pm

dtaylor113: idk, it's kinda what I expect.  The most recently viewed item is always the left-most in the list.

serenamarie125: hmmm ... regardless if it's alphabetical or not, you will still see things move :) won't you ? Maybe less frequently, but it'll still happen.  So i'm not sure that alphabetizing it "fixes" the issue.  @beanh66 thoughts?

spadgett: It fixes the issue where the exact icon I just clicked moved :)

spadgett: We can look at this as a follow-on if not everyone's on board. I do think we should at least soften the transition though since it draws your eyes away from the overlay panel animation. Maybe a delay or fade in / fade out.

dtaylor113: I suggest opening an issue for this which we can follow up on.  -thanks

dtaylor113: FWIW, it's a removal of one line in Project Summary to not update the recently viewed items when clicked on from within Project Summary.

serenamarie125: @dtaylor113 when does it get updated?  Maybe we should update Recently Viewed once the overlay panel is "closed"?

dtaylor113: Recently Viewed Items get updated when clicked upon from Search, Service View, and Project Summary.

beanh66: Yeah after interacting with it I agree @serenamarie125! Having it update after the overlay panel closes would be less distracting

dtaylor113: Ok, created issue #168 to track this

PR #149, opened 2017-04-12
Refactoring panel overlay layout so items don't get clipped

### LINKED ISSUE #116: View project button cut off at some screen widths
screen shot 2017-04-08 at 10 10 51 am

rhamilto:
Fixes #116 

This change allows the inner content to expand, but only up to a max-height, at which point it scrolls.  As a result, I've removed catalogs-overlay-panel-full-height.  I am unclear if we still need 
catalogs-overlay-panel-single-column.  Given the nature of these changes and my lack of knowledge on the history of the design decisions, I imagine this may violate some of those decisions.

Please ignore the shadow on the left side of the latter screenshots.  That's macOS's dock casting that shadow (it's not part of the implementation).

screen shot 2017-04-12 at 6 35 09 pm
screen shot 2017-04-12 at 6 35 18 pm
screen shot 2017-04-12 at 6 35 48 pm
screen shot 2017-04-12 at 6 35 57 pm
screen shot 2017-04-12 at 6 36 36 pm
screen shot 2017-04-12 at 6 36 21 pm
screen shot 2017-04-12 at 6 37 13 pm


serenamarie125:
@rhamilto I think that the single-column could be being used for import yaml or deploy image, but I could be wrong.  Not sure exactly what you changed, but from a design perspective, having the scrollbar attached to the content between the steps panel & button bar is what we expected!

spadgett:
From my testing, the dialog works much better, and this fixes the scrollbar problems we were seeing. The CSS is also cleaner.

@serenamarie125 The single column layout should still work with these changes.

@rhamilto I believe the only difference with single-column is `max-width: 800px`, which we use in the bind dialog for the console. (I see you left that in.)

beanh66:
This looks great @rhamilto!

PR #151, opened 2017-04-13
Change subcategory background color

cshinn:
An alternate coloration of the subcategory background

screen shot 2017-04-13 at 9 26 49 am

screen shot 2017-04-13 at 9 26 54 am

screen shot 2017-04-13 at 9 25 57 am

### DIFF
--- a/dist/origin-web-catalogs.css
+++ b/dist/origin-web-catalogs.css
@@ -60,6 +60,7 @@
 }
 @media (min-width: 768px) {
   .landing {
+    background-color: #fff;
     bottom: 0;
     overflow-y: auto;
     position: absolute;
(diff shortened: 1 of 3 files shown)


jeff-phillips-18:
Might look better if there was some margin around the catalog items panel to show a touch of the background image.

rhamilto:
To my eye, the primary nav is getting lost now.  Reducing the opacity on the h1 (.5) and the primary nav (.8) and increasing it on the tab bg (.95) improves things.

screen shot 2017-04-13 at 10 45 47 am

serenamarie125:
@rhamilto imo that brings in too many lines/stripes, I'd prefer the original.  I particularly like the fact that you can see the image in the area next to the subcategories.

spadgett:
This isn't a new problem, but it's odd to me that there's a max height and scrollbar here when there's a lot of space below. Changing the color to white makes it feel even more like I should be able to use the whole space.

screen shot 2017-04-13 at 12 08 55 pm

beanh66:
@spadgett I agree with you it is a little weird since we have the space. @rhamilto I agree with @serenamarie125 on the stripes, adding too many horizontal lines may be distracting.

jeff-phillips-18:
A bit more space looks good to me:

image

serenamarie125:
I think this already got merged, but @jeff-phillips-18 we should take a look at that with the blue background and see if it looks good when resized.  I definitely like the direction.

beanh66:
@jeff-phillips-18 @serenamarie125 I agree, this looks great with the margins.

PR #152, opened 2017-04-13
Updates to projects summary panel

### LINKED ISSUE #40: Projects summary does not show Getting Started area/text

### LINKED ISSUE #144: "My Projects" heading should link to the full projects list
@jeff-phillips-18 The "My Projects" heading on the homepage projects panel should link to the full projects list (navigates to the same place as "view all"). This should follow the same hover treatment as the other links on the right panel. 

The top heading on the projects panel is larger than the lower heading. My Projects is larger than Getting Started when both are displayed. They should be the same size (smaller of the two).

@serenamarie125 FYI ^

jeff-phillips-18:
Adjust font-size for My Projects / Getting Started title
Add link from ‘My Projects’ to go to the projects page
Add message when user is not allowed to create a project
Add loading spinner to avoid flicker from Getting Started to My Projects

Fixes #40 
Fixes #144 

image

image


src/constants.ts
```diff
@@ -86,3 +86,5 @@ var resources: any = {
 };
 
 _.set(window, 'OPENSHIFT_CONSTANTS.CATALOG_HELP_RESOURCES', resources);
+
+_.set(window, 'OPENSHIFT_CONSTANTS.NEW_PROJECT_MESSAGE', "Contact your system administrator at projectname@example.com to request a project.");
```
spadgett: @jeff-phillips-18 This is customized in OpenShift master-config.yaml. We should use that message.

https://github.com/openshift/origin-web-console/blob/master/app/scripts/controllers/projects.js#L154-L163

jwforres:
In an environment like online free tier it is quite possible that we will be pre-provisioning a project for a user, so they will come in for the first time with a project already, but they wouldn't know what to do next. Probably not something to fix in this PR but @beanh66 might want to think about a design for that?

app/config.js
```diff
@@ -41,4 +41,4 @@ window.OPENSHIFT_VERSION = {
   kubernetes: "dev-mode"
 };
 
-window.OPENSHIFT_CONSOLE_BASE_URL = "https://localhost:9000/dev-console";
+window.OPENSHIFT_CONSOLE_BASE_URL = "https://localhost:" + (window.DEV_SERVER_PORT || '9001');
```
spadgett: This should be the web console URL, not the catalog URL. It's used for project links.

jeff-phillips-18: This is in the showcase app only. It breaks in there unless you happen to have the console running.

beanh66:
@jeff-phillips-18 I was thinking the My Projects heading should still be white so it stands out and is consistent with the other headers, then on hover it would be blue and underlined. I know this is slightly inconsistent with the way other links work, but @cshinn @serenamarie125 @ncameronbritt what do you think?

src/constants.ts
jeff-phillips-18: It does use that, but defaults to what is in constants just in case. I can remove the constant if desired.

https://github.com/openshift/origin-web-catalog/blob/master/src/components/projects-summary/projects-summary.controller.ts#L56-L68

spadgett: In that case, we should make the default message something real because the custom message is not required.

app/config.js
spadgett: But it's broken anyway since the project links will never work against the showcase app URL. We went from sometimes working to never working. You can customize it to point to another base URL if you need.

jeff-phillips-18: there is a projects section (though very rough) and the link should work. The sometimes working take you out of the showcase application. That said, I can put it back since it is customizable. It just seemed bad to me when I clicked the link and left the current app.

src/constants.ts
jwforres: That is what we do in the console, if no custom message is defined we fall back to a generic message defined in the source.  I would not put it in constants mainly because we don't want to give people two ways to customize that message and constants are supposed to be used for things people may want to override.

app/config.js
spadgett: > The sometimes working take you out of the showcase application.

But that's what's going to happen when the catalog is running outside of the console. Why not make the showcase app work like it will really work? And then we can test the base URL customization.

I'm OK building the URL form the API server config, but I'd rather not change it to something we know is broken.

src/constants.ts
jeff-phillips-18: I see, I found that code and will incorporate.

app/config.js
jeff-phillips-18: Ok, I'll revert this change.

jeff-phillips-18:
@beanh66 @jwforres We don't currently show the not on empty projects. That would mean retrieving each projects individually to gather that information. If this is necessary, I can add it.

spadgett:
This could be a lot of requests depending on what types we check. I guess it depends on how we define "empty" since even new project have a few things like service accounts and secrets.

jeff-phillips-18:
Default message:

image

src/components/projects-summary/projects-summary.html
```diff
@@ -1,10 +1,25 @@
...
+  </h2>
+  <div ng-if="!$ctrl.canCreate">
+    <span ng-if="!$ctrl.newProjectMessage">
+      A cluster admin can create a project for you by running the command:
+      <div class="code-block">
+        <code class="projects-instructions-link ">oadm new-project &lt;projectname&gt;
+   --admin={{$ctrl.user.metadata.name || '&lt;YourUsername&gt;'}}</code>
+      </div>
+    </span>
+    <span ng-if="newProjectMessage" ng-bind-html="newProjectMessage | linky : '_blank'"></span>
```
```html
<span ng-if="$ctrl.newProjectMessage" ng-bind-html="$ctrl.newProjectMessage | linky : '_blank'"></span>
```

jeff-phillips-18: yes, thank you.

beanh66:
@jeff-phillips-18 LGTM!

jeff-phillips-18:
Fixed issue, added tests.

jeff-phillips-18:
Sorry for the churn, added two more updated based on UXD review with @beanh66

spadgett:
@jeff-phillips-18 Is this ready to merge?

jeff-phillips-18:
@spadgett yes, thanks.

PR #164, opened 2017-04-18
Consolidated retrieval of Catalog Service Items

dtaylor113:
In `<catalog-search>`, `<services-view>`, and `<projects-summary>` replaced attributes `service-classes` and `image-streams` with a single attribute `catalog-service-items`.  `catalog-items` are retrieved via `CatalogServices` class.  

This PR basically removes a bunch of data normalization functions, which were duplicated in the three components listed above, and centralizes them in CatalogServices.

### DIFF
--- a/app/pages/home/homePageController.ts
+++ b/app/pages/home/homePageController.ts
@@ -1,25 +1,17 @@
 export class HomePageController {
-  static $inject = ['$state', 'AuthService', 'Logger', 'Constants', 'DataService'];
+  static $inject = ['$state', 'AuthService', 'Catalog', 'Constants'];
 
   public ctrl: any = this;
   private $state: any;
   private authService: any;
-  private logger: any;
+  private Catalog: any;
   private constants: any;
-  private dataService: any;
-  private Logger: any;
 
-  constructor($state: any, AuthService: any, Logger: any, Constants: any, DataService: any) {
+  constructor($state: any, AuthService: any, Catalog: any, Constants: any) {
     this.$state = $state;
     this.authService = AuthService;
-    this.logger = Logger;
+    this.Catalog = Catalog;
     this.constants = Constants;
-    this.dataService = DataService;
-    this.Logger = Logger;
-    this.ctrl.applications = [];
-    this.ctrl.categories = [];
-    this.ctrl.serviceClasses = {};
-    this.ctrl.imageStreams = {};
   };
 
   public $onInit() {
@@ -33,21 +25,10 @@ export class HomePageController {
   };
 
   public update() {
-    this.dataService.list({
-      group: 'servicecatalog.k8s.io',
-      resource: 'serviceclasses'
-    }, {}).then( (resources: any) => {
-      this.ctrl.serviceClasses = resources.by("metadata.name");
+    this.Catalog.getCatalogItems().then( (catalogServiceItems: any) => {
+      this.ctrl.catalogItems = catalogServiceItems;
     }, () => {
-      this.logger.log("Error Loading serviceclasses from servicecatalog.k8s.io: ");
-      this.ctrl.serviceClasses = {};
-    });
-
-    this.dataService.list("imagestreams", {namespace: "openshift"}).then( (resources: any) => {
-      this.ctrl.imageStreams = resources.by("metadata.name");
-    }, () => {
-      this.logger.log("Error Loading openshift imagestreams: ");
-      this.ctrl.imageStreams = {};
+      this.ctrl.catalogItems = {};
     });
 
     this.ctrl.saasOfferings = this.constants.SAAS_OFFERINGS;
--- a/src/components/catalog-search/catalog-search.component.ts
+++ b/src/components/catalog-search/catalog-search.component.ts
@@ -3,8 +3,7 @@ import {CatalogSearchController} from './catalog-search.controller';
 export const catalogSearch: angular.IComponentOptions = {
   bindings: {
     baseProjectUrl: '@',
-    imageStreams: '<',
-    serviceClasses: '<'
+    catalogItems: '<'
   },
   controller: CatalogSearchController,
   template: require('./catalog-search.html')
--- a/src/components/catalog-search/catalog-search.controller.ts
+++ b/src/components/catalog-search/catalog-search.controller.ts
@@ -7,7 +7,6 @@ export class CatalogSearchController implements angular.IController {
 
   private Catalog: any;
   private KeywordService: any;
-  private allItems: any[];
   private $scope: any;
   private $q: any;
   private loaded: boolean = false;
@@ -26,27 +25,8 @@ export class CatalogSearchController implements angular.IController {
   }
 
   public $onChanges(onChangesObj: angular.IOnChangesObject) {
-    if (onChangesObj.serviceClasses || onChangesObj.imageStreams) {
-      if (!this.ctrl.serviceClasses && !this.ctrl.imageStreams) {
-        return;
-      }
+    if (onChangesObj.catalogItems && this.ctrl.catalogItems) {
 
-      // Convert service classes to ServiceItem, which is needed for the ordering panel.
-      let items = _.map(this.ctrl.serviceClasses, (serviceClass) => {
-        return this.Catalog.getServiceItem(serviceClass);
-      });
-
-      // Convert builders to ImageItem.
-      items = items.concat(_.map(this.ctrl.imageStreams, (imageStream) => {
-        return this.Catalog.getImageItem(imageStream);
-      }));
-
-      // Remove null items (non-builder images).
-      items = _.reject(items, (item) => {
-        return !item;
-      });
-
-      this.allItems = _.sortBy(items, 'name');
       this.loaded = true;
 
       // Show search results now if the user began typing before the items loaded.
@@ -79,7 +59,7 @@ export class CatalogSearchController implements angular.IController {
 
   private filterForKeywords(searchText: string) {
     let keywords = this.KeywordService.generateKeywords(searchText);
-    let items = this.KeywordService.filterForKeywords(this.allItems, ['name', 'tags', 'resource.osbTags'], keywords);
+    let items = this.KeywordService.filterForKeywords(this.ctrl.catalogItems, ['name', 'tags', 'resource.osbTags'], keywords);
     return _.take(items, 5);
   }
 }
--- a/src/components/projects-summary/projects-summary.component.ts
+++ b/src/components/projects-summary/projects-summary.component.ts
@@ -2,9 +2,8 @@ import {ProjectsSummaryController} from './projects-summary.controller';
 
 export const projectsSummary: angular.IComponentOptions = {
   bindings: {
-    imageStreams: '<',
-    serviceClasses: '<',
     baseProjectUrl: '@',
+    catalogItems: '<',
     projectsUrl: '@',
     viewEditMembership: '&',
     startGettingStartedTour: '&'
--- a/src/components/projects-summary/projects-summary.controller.ts
+++ b/src/components/projects-summary/projects-summary.controller.ts
@@ -135,27 +135,8 @@ export class ProjectsSummaryController implements angular.IController {
   };
 
   public $onChanges(onChangesObj: angular.IOnChangesObject) {
-    if (onChangesObj.serviceClasses || onChangesObj.imageStreams) {
-      if (!this.ctrl.serviceClasses && !this.ctrl.imageStreams) {
-        return;
-      }
-
-      // Convert service classes to ServiceItem, which is needed for the ordering panel.
-      let items = _.map(this.ctrl.serviceClasses, (serviceClass) => {
-        return this.Catalog.getServiceItem(serviceClass);
-      });
-
-      // Convert builders to ImageItem.
-      items = items.concat(_.map(this.ctrl.imageStreams, (imageStream) => {
-        return this.Catalog.getImageItem(imageStream);
-      }));
-
-      // Remove null items (non-builder images).
-      items = _.reject(items, (item) => {
-        return !item;
-      });
-
-      this.allItems = _.indexBy(items, 'resource.metadata.uid');
+    if (onChangesObj.catalogItems && this.ctrl.catalogItems) {
+      this.allItems = _.indexBy(this.ctrl.catalogItems, 'resource.metadata.uid');
       this.ctrl.recentlyViewedItems = this.getRecentlyViewedItems();
     }
   }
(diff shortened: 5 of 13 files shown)


src/components/services-view/services-view.controller.ts
```diff
@@ -21,44 +19,39 @@ export class ServicesViewController implements angular.IController {
...
-      this.imageStreamsLoaded = true;
-      this.updateImageStreams();
+    if (onChangesObj.catalogServiceItems && (!onChangesObj.catalogServiceItems.isFirstChange() ||
+      !_.isEmpty(onChangesObj.catalogServiceItems.currentValue))) {
+      this.ctrl.filteredItems = this.ctrl.catalogServiceItems;
+      this.ctrl.categories = this.catalog.removeEmptyCategories(this.ctrl.filteredItems);
+      this.ctrl.subCategories = this.getSubCategories('all');
+      this.ctrl.isEmpty = _.isEmpty( this.ctrl.catalogServiceItems);
+      this.ctrl.loaded = true;
     }
```
jeff-phillips-18: None of this gets invoked if there are no items.  isEmpty would never get set and the page would continuously show the loading state.

app/pages/home/homePageController.ts
```diff
@@ -1,25 +1,20 @@
 export class HomePageController {
-  static $inject = ['$state', 'AuthService', 'Logger', 'Constants', 'DataService'];
+  static $inject = ['$state', 'AuthService', 'Logger', 'Catalog', 'Constants'];
```
jeff-phillips-18: Logger is no longer used

src/components/catalog-search/catalog-search.controller.ts
```diff
@@ -26,27 +25,9 @@ export class CatalogSearchController implements angular.IController {
...
-      items = items.concat(_.map(this.ctrl.imageStreams, (imageStream) => {
-        return this.Catalog.getImageItem(imageStream);
-      }));
-
-      // Remove null items (non-builder images).
-      items = _.reject(items, (item) => {
-        return !item;
-      });
+    if (onChangesObj.catalogServiceItems && (!onChangesObj.catalogServiceItems.isFirstChange() ||
+      !_.isEmpty(onChangesObj.catalogServiceItems.currentValue))) {
```
spadgett: I think it is better to make `null` or `undefined` mean not loaded, and `[]` mean loaded, but empty.

src/components/catalog-search/catalog-search.component.ts
```diff
@@ -3,8 +3,7 @@ import {CatalogSearchController} from './catalog-search.controller';
 export const catalogSearch: angular.IComponentOptions = {
   bindings: {
     baseProjectUrl: '@',
-    imageStreams: '<',
-    serviceClasses: '<'
+    catalogServiceItems: '<'
```
spadgett: Can we just call these `catalogItems`? Not everything is a service.

src/components/projects-summary/projects-summary.component.ts
```diff
@@ -2,9 +2,8 @@ import {ProjectsSummaryController} from './projects-summary.controller';
 
 export const projectsSummary: angular.IComponentOptions = {
   bindings: {
-    imageStreams: '<',
-    serviceClasses: '<',
     baseProjectUrl: '@',
+    catalogServiceItems: '<',
```
spadgett: `catalogItems`

src/components/projects-summary/projects-summary.controller.ts
```diff
@@ -135,27 +135,9 @@ export class ProjectsSummaryController implements angular.IController {
...
-      }));
-
-      // Remove null items (non-builder images).
-      items = _.reject(items, (item) => {
-        return !item;
-      });
-
-      this.allItems = _.indexBy(items, 'resource.metadata.uid');
+    if (onChangesObj.catalogServiceItems && (!onChangesObj.catalogServiceItems.isFirstChange() ||
+      !_.isEmpty(onChangesObj.catalogServiceItems.currentValue))) {
```
spadgett: Same comment here. `null` or `undefined` if not loaded, and we should be able to remove the `isFirstChange` logic. I don't think we should assume what first change and subsequent changes mean

src/components/services-view/services-view.component.ts
```diff
@@ -3,8 +3,7 @@ import {ServicesViewController} from './services-view.controller';
 export const servicesView: angular.IComponentOptions = {
   bindings: {
     baseProjectUrl: '@',
-    serviceClasses: '<',
-    imageStreams: '<'
+    catalogServiceItems: '<',
```
spadgett: `catalogItems`

src/components/services-view/services-view.controller.ts
```diff
@@ -21,44 +19,39 @@ export class ServicesViewController implements angular.IController {
...
-      this.serviceClassesLoaded = true;
-      this.updateServiceClasses();
-    }
-
-    if (onChangesObj.imageStreams && !onChangesObj.imageStreams.isFirstChange()) {
-      this.ctrl.imageStreams = onChangesObj.imageStreams.currentValue;
-      this.imageStreamsLoaded = true;
-      this.updateImageStreams();
+    if (onChangesObj.catalogServiceItems && (!onChangesObj.catalogServiceItems.isFirstChange() ||
+      !_.isEmpty(onChangesObj.catalogServiceItems.currentValue))) {
```
spadgett: Same comment again :)  `null` or `undefined` for not loaded

src/components/services-view/services-view.html
```diff
@@ -52,6 +52,7 @@
     <div ng-if="$ctrl.currentFilter === 'other' || $ctrl.currentFilter === 'all'" class="services-items">
       <div class="services-items-inner">
         <div class="services-items-inner-shadow-covers">
+          <div ng-if="$ctrl.isEmpty">There are no Catalog Items</div>
```
spadgett: Let's add punctuation and remove the capitalization. We don't capitalize kinds in other parts of the web console.

"There are no catalog items."

src/services/catalog.service.ts
```diff
@@ -2,16 +2,75 @@ import * as angular from 'angular';
...
+    this.logger = Logger;
+  }
+
+  public getCatalogServiceItems() {
+    let deferred: any = this.$q.defer();
+    let serviceClasses: any, imageStreams: any;
+
+    this.getServiceClasses().then((resources: any) => {
+      serviceClasses = resources;
+      this.getImageStreams().then((resources: any) => {
```
spadgett: Better to make these requests asynchronously and use `$q.all()`. It would greatly simplify this logic. I don't think you need to create your own deferred anywhere. Let's chat about this.

src/services/catalog.service.ts
```diff
@@ -79,6 +138,32 @@ export class CatalogService {
 
     return retCategories;
   }
+
+  private getServiceClasses() {
+    let deferred: any = this.$q.defer();
+
+    this.dataService.list({group: 'servicecatalog.k8s.io', resource: 'serviceclasses'}, {}).then((resources: any) => {
```
spadgett: Just return the value from `dataService.list` instead of creating your own deferred.

src/services/catalog.service.ts
```diff
@@ -79,6 +138,32 @@ export class CatalogService {
...
+      deferred.reject();
+    });
+
+    return deferred.promise;
+  }
+
+  private getImageStreams() {
+    let deferred: any = this.$q.defer();
+
+    this.dataService.list("imagestreams", {namespace: "openshift"}).then((resources: any) => {
```
spadgett: Return the value from `dataService.list` instead of creating your own deferred.

src/services/catalog.service.ts
```diff
@@ -2,16 +2,75 @@ import * as angular from 'angular';
...
+  constructor($filter: any, $q: any, constants: any, DataService: any, Logger: any) {
     this.$filter = $filter;
+    this.$q = $q;
+    this.constants = constants;
     this.categories = this.constants.SERVICE_CATALOG_CATEGORIES;
+    this.dataService = DataService;
+    this.logger = Logger;
+  }
+
+  public getCatalogServiceItems() {
```
spadgett: `getCatalogItems`

src/services/catalog.service.ts
dtaylor113: I did initially have $q.all() however that would mean if one call failed, the entire promise would be rejected.  I was trying to have it so that if serviceClasses failed, it would still list imageBuilders.

src/components/catalog-search/catalog-search.controller.ts
dtaylor113: Ok, what about when there is an error and the promise is rejected -should catalogItems be set to [ ] or left as undefined?   When error, do we to indicate the error on the view/page, or say 'There are no catalog items'/

dtaylor113: Fixed, although not sure how/if we want to display an error msg when one occurs.

src/components/catalog-search/catalog-search.component.ts
dtaylor113: Fixed

src/components/projects-summary/projects-summary.component.ts
dtaylor113: Fixed

src/components/projects-summary/projects-summary.controller.ts
dtaylor113: yeah, I was trying to accommodate production code (which is async so $onChanges gets called twice) and unit tests (where $onChanges only is called once).  
I fixed with a much simpler test.

src/components/services-view/services-view.component.ts
dtaylor113: Fixed

src/components/services-view/services-view.controller.ts

src/components/services-view/services-view.html
dtaylor113: Fixed

src/services/catalog.service.ts

src/services/catalog.service.ts
dtaylor113: Changed back to $q.all(), fixed.

spadgett: It seems wrong to me that if I call a method that is supposed to get me both, and it only gives me some, that it returns success. As a caller, I would want to know if something failed.

If we really care about continuing after errors, we should have separate methods for service classes and image streams and then another method to combine and sort them. Then the callers can decide what to do if one fails. I kind of like the current approach though because it will let us turn on and off templates in the catalog in one place without having to change the calling code.

src/services/catalog.service.ts
dtaylor113: Removed since now using $q.all()

src/services/catalog.service.ts
dtaylor113: Removed since now using $q.all()

app/pages/home/homePageController.ts
dtaylor113: Fixed

dtaylor113: Fixed

src/components/catalog-search/catalog-search.controller.ts
dtaylor113: Fixed

src/components/services-view/services-view.controller.ts
dtaylor113: yeah, I was trying to accommodate production code (which is async so $onChanges gets called twice) and unit tests (where $onChanges only is called once).
I fixed with a much simpler test.

dtaylor113:
Hi, I believe I addressed all issues.  Not sure why Sam's comments got repeated again at the end of this PR?

src/services/catalog.service.ts
```diff
@@ -2,16 +2,70 @@ import * as angular from 'angular';
...
+    };
+
+    this.$q.all(promises).then((values) => {
+      serviceClasses = values.serviceClasses.by("metadata.name");
+      imageStreams = values.imageStreams.by("metadata.name");
+      this.catalogServiceItems = this.convertToServiceItems(serviceClasses, imageStreams);
+      deferred.resolve(this.catalogServiceItems);
+    }, () => {
+      this.logger.log("Error Loading Catalog Items");
+      deferred.reject();
```
spadgett: We should try to pass the HTTP error through in deferred.reject so we can show the cause in the error message, although this might not be necessary if you use `return $q.all...` as suggested above.

src/services/catalog.service.ts
```diff
@@ -2,16 +2,70 @@ import * as angular from 'angular';
...
+
+    let promises = {
+      serviceClasses: this.dataService.list({
+        group: 'servicecatalog.k8s.io',
+        resource: 'serviceclasses'
+      }, {}),
+      imageStreams: this.dataService.list("imagestreams", {namespace: "openshift"})
+    };
+
+    this.$q.all(promises).then((values) => {
```
spadgett: Thanks, @dtaylor113. This looks a lot cleaner.

You could probably just `return this.$q.all(...).then(...)` here and not have to create a deferred yourself.

dtaylor113: If I do that, how do I call `deferred.resolve/reject` in the `this.$q.all(...).then(...)`?

dtaylor113: I tried:
```
public getCatalogItems() {
    return this.$q.all({
      serviceClasses: this.dataService.list({
        group: 'servicecatalog.k8s.io',
        resource: 'serviceclasses'
      }, {}),
      imageStreams: this.dataService.list("imagestreams", {namespace: "openshift"})
    }).then((values) => {
      let serviceClasses: any = values.serviceClasses.by("metadata.name");
      let imageStreams: any = values.imageStreams.by("metadata.name");
      this.catalogServiceItems = this.convertToServiceItems(serviceClasses, imageStreams);
      this.$q.resolve(this.catalogServiceItems);
    }, () => {
      this.logger.log("Error Loading Catalog Items");
      this.$q.reject();
    });
  }
```
But that causes the an endless spinner in the service-view, and no 'Recently Viewed' in Project Summary.

spadgett: You don't need to. You just return the new value from `then`. This can probably be....

```js
return this.$q.all((values) => {
  let serviceClasses = values.serviceClasses.by("metadata.name");
  let imageStreams = values.imageStreams.by("metadata.name");
  return this.convertToServiceItems(serviceClasses, imageStreams);
});
```

dtaylor113: Ok, fixed, thanks!

PR #166, opened 2017-04-20
Add toast notifications to the showcase application

jeff-phillips-18:
image

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -12,6 +12,7 @@
   <body class="ng-cloak surface-shaded">
     <navigation>
     </navigation>
+    <toast-notifications></toast-notifications>
     <main ng-view>
       <ui-view></ui-view>
     </main>
--- a/app/mockServices/mockData.service.ts
+++ b/app/mockServices/mockData.service.ts
@@ -103,7 +103,7 @@ export class DataService implements IDataService {
         "annotations": {
           "openshift.io/description": object.description,
           "openshift.io/display-name": object.displayName,
-          "openshift.io/requester": "mock devleoper",
+          "openshift.io/requester": "mock developer",
         },
       },
       "spec": {


dtaylor113:
LGTM.  There are going to be merge conflicts as landingPage and homePage html and controllers changed in the 'Recently Viewed' and 'Consolidate Catalog Items' PRs.

src/components/landing-page/landing-page.controller.ts
```diff
@@ -2,10 +2,39 @@ import * as angular from 'angular';
...
 
   public $onInit() {
     this.ctrl.searchText = '';
   }
+
+  public $postLink() {
+    this.$scope.$watch(() => {
+      return this.ctrl.alerts;
+    }, () => {
+      _.forEach(this.ctrl.alerts, (alert: any) => {
```
jwforres: I don't like that this timer logic would have to be handled by every controller that wants to have toast notifications

spadgett: This code looks fine, but I'm worried it's going to be copied and pasted everywhere we want to use toast notifications. I think this would make sense as a service that shows and hides toast notifications, and there is a property in the alert object saying whether it should be automatically hidden after a delay.

jeff-phillips-18: OK, that makes sense. This was the piece I was most interested in getting a review on. I'm not sure we want every place that add alerts to have to specify how to handle each alert though. What if we made the service configurable to hide alerts by type after the delay? We could also allow individual alerts to specify, but configuring the service would at least be application wide.

spadgett: > What if we made the service configurable to hide alerts by type after the delay? We could also allow individual alerts to specify, but configuring the service would at least be application wide.

+1, good idea. Should help us be consistent in how we use toasts throughout the app

src/components/landing-page/landing-page.controller.ts
```diff
@@ -2,10 +2,39 @@ import * as angular from 'angular';
...
+        if (alert.type === 'success' && !alert.hidden) {
+          if (!alert.timerId) {
+            alert.timerId = this.$timeout(function () {
+              alert.timerId = -1;
+              if (!alert.isHover) {
```
spadgett: We might want to force all toasts to have a close button. Otherwise if I hover over it, I won't ever be able to dismiss?

spadgett: Nevermind, I see it will dismiss when I mouse out because it triggers an alert update

jeff-phillips-18:
With the updates to https://github.com/openshift/origin-web-common/pull/39 it is only required to include the toast-notifications directive at the application level (could by page if desired).

jeff-phillips-18:
This PR is no longer WIP. Ready for final reviews.

PR #167, opened 2017-04-20
Add Landing Page Guided Tour

jeff-phillips-18:
Please take a look at constants.ts in particular. This is the quick and dirty way of setting up the tour. It is prone to breaking if any of the selectors change. We could define constants for these selectors that could then be more easily updated, but how many to define would be the question. This along with defining functions used to show certain portions of the page (showAllCatalogs, showCategories) should be discussed and I'm open to suggestions.

### DIFF
--- a/src/components/projects-summary/projects-summary.controller.ts
+++ b/src/components/projects-summary/projects-summary.controller.ts
@@ -21,7 +21,7 @@ export class ProjectsSummaryController implements angular.IController {
 
   public ctrl: any = this;
   public showNewProjectPanel: boolean = false;
-  public showEditwProjectPanel: boolean = false;
+  public showEditProjectPanel: boolean = false;
   public alerts: any = [];
   public projects: any = [];
   private $element: any;
@@ -191,7 +191,7 @@ export class ProjectsSummaryController implements angular.IController {
   };
 
   public handleGettingStartedClick() {
-    var cb: any = this.ctrl.startGettingStartedTour();
+    var cb: any = this.ctrl.startTour();
     if (cb) {
       cb();
     }
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -82,3 +82,77 @@ var resources: any = {
 };
 
 _.set(window, 'OPENSHIFT_CONSTANTS.CATALOG_HELP_RESOURCES', resources);
+
+var showAllCatalogs = function() {
+  $('body').find('.services-view-container .nav-tabs a')[0].click();
+};
+
+var showCategories = function() {
+  $($('body').find('.services-view-container .nav-tabs li')[1]).find('a')[0].click();
+  setTimeout(function() {
+    $('body').find('.services-sub-category-tab')[1].click();
+  });
+};
+
+var homePageTourSteps = {
+  id: "getting-started-tour",
+  steps: [
+    {
+      title: "Create Project",
+      content: "Projects allow you to organize and manage your content. Projects require a unique name, and optionally can include a display name, and description",
+      target: ".catalog-projects-summary-panel .create-button",
+      targetScrollElement: '.landing-side-bar',
+      placement: "left"
+    },
+    {
+      title: "Search Catalog",
+      content: "Search by name, description, keyword, or label to quickly locate items in the catalog that you want to add to your project.",
+      target: ".landing-search-area .landing-search-form",
+      placement: "bottom",
+      fixedElement: true
+    },
+    {
+      title: "Browse Catalog",
+      content: "If you don’t know exactly what you are looking for, you can browse all available catalog items under the first tab in the catalog.",
+      target: ".services-view-container",
+      targetScrollElement: '.landing',
+      placement: "top",
+      xOffset: 20,
+      preShow: showAllCatalogs
+    },
+    {
+      title: "Browse by Category",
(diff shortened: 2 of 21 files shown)


app/components/oauth/oauth.controller.ts
```diff
@@ -56,15 +62,31 @@ export class OAuthController implements angular.IController {
...
+                _this.$location.replace();
+                _this.$location.url(destination);
+              } else {
+                localStorage.setItem(loggedInKey, 'true');
+                _this.$location.replace();
+                _this.$location.url('./');
+                this.$timeout(function() {
+                  let tourConfig = _this.Constants.LANDING_PAGE_TOUR_STEPS;
+                  if (tourConfig && tourConfig.steps) {
+                    _this.GuidedTourService.startTour(tourConfig);
```
jwforres: i know this is the example in the catalog app, but thinking about this from the context of how it will behave in the console I have a few concerns

1) a customer may not want their users forced to go through a guided tour, so we either need to let them turn off the tour completely and hide the button (possibly by null-ing the LANDING_PAGE_TOUR_STEPS, or setting it to an empty array), or we have to let them disable automatically sending a user through the guided tour

2) If someone sends me a link to a particular page like logs for a pod, and I'm logging in for the first time, then I still want that to take me to the log page. So we can't always start the guided tour for someone who logs in for the first time.  We need to check if the "then" is set or not.

3) We should consider upgraded clusters.  Users may have been logged in before on the cluster, but the token would never have been set in localStorage.  As an existing user on a cluster would I be annoyed by a guided tour popping up, even if its clearly for a drastically new experience?  I'm on the fence on this one.  cc @serenamarie125

src/components/projects-summary/projects-summary.controller.ts
```diff
@@ -191,7 +191,7 @@ export class ProjectsSummaryController implements angular.IController {
   };
 
   public handleGettingStartedClick() {
-    var cb: any = this.ctrl.startGettingStartedTour();
+    var cb: any = this.ctrl.startTour();
```
jwforres: should we be making sure startTour has been passed into the directive, and hiding the button if startTour is undefined / not a function.  Really a question of whether the guided tour is optional.

src/constants.ts
```diff
@@ -85,3 +85,78 @@ var resources: any = {
...
+      content: "Additional resources can be found here or you can always access the help icon in the top banner for more information.",
+      target: ".resources-panel",
+      targetScrollElement: '.landing-side-bar',
+      placement: "left",
+      yOffset: 50,
+    }
+  ]
+};
+
+_.set(window, 'OPENSHIFT_CONSTANTS.LANDING_PAGE_TOUR_STEPS', homePageTourSteps);
```
jwforres: I think since we know we will have multiple tours down the road, having a higher level OPENSHIFT_CONSTANTS.GUIDED_TOURS map makes sense, and underneath that map they can have unique names like "landing_page_tour".  Otherwise this config is what you are directly passing to hopscotch right?  I think thats the right thing.

app/components/oauth/oauth.controller.ts
```diff
@@ -103,6 +125,7 @@ export class OAuthController implements angular.IController {
   }
 
   public completeLogin() {
+    console.log("Logged in");
```
jeff-phillips-18: Just keeping you honest 😉

app/components/oauth/oauth.controller.ts
jeff-phillips-18: What if we:

- Allow shutting of the auto-tour via the config
- Don't add the menu item or the button if the constant does not exist
- Auto start the tour when the landing page is navigated to and the user has not yet seen it.

As for the upgrade, this was discussed with @beanh66 during development and the idea is that this is new content and it is valid to offer a tour of it.

src/constants.ts
jeff-phillips-18: It is what is being passed to hopscotch. I will adjust the structure as advised.

src/components/projects-summary/projects-summary.controller.ts
jeff-phillips-18: I have hidden the button and menu item when there is no tour configured.

jeff-phillips-18:
Removed WIP status, ready for final reviews.

src/constants.ts
```diff
@@ -85,3 +85,85 @@ var resources: any = {
...
+    },
+    {
+      title: "Search Catalog",
+      content: "Search by name, description, keyword, or label to quickly locate items in the catalog that you want to add to your project.",
+      target: ".landing-search-area .landing-search-form",
+      placement: "bottom",
+      fixedElement: true
+    },
+    {
+      title: "Build Applications",
```
jwforres: This section is going to be empty OOTB, so having a step for it won't make sense.  I'd rather not override the default catalog tour in the console constants. I would prefer to just add this step in the environments where we plan to have SaaS offerings configured (like Online).

jeff-phillips-18:
Addressed @jwforres concerns. Rebased and squashed.

jwforres:
@jeff-phillips-18 ready to merge?

jeff-phillips-18:
@jwforres I believe so.

jwforres:
should make sure we get a catalog release cut tomorrow and bump the web console before sprint devcut

PR #170, opened 2017-04-21
Styling of recently view items in sidebar and some minor fine tuning

### LINKED ISSUE #131: Project Summary: Recently Viewed Items - Need black icon on white circle
Currently, Recently Viewed item looks like this:

image

Design Doc calls for white circle:

image

http://openshift.github.io/openshift-origin-design/web-console/1-homepage/project-panel#recently-viewed

I was able to add a white color to `.services-item-icon:after` but it covers up the service item icon.

sg00dwin:
Had to restructure some rules to prevent conflicts between `.landing-side-bar` and `.services-items-inner` being on different backgrounds

... hover state shown

screen shot 2017-04-21 at 3 19 40 pm

**Other fine tuning**
- Crisper modal animation
- remove underline and switch to lighter blue from project heading link:hover state

Fixes #131


sg00dwin:
@jwforres @spadgett 
cc @dtaylor113 @rhamilto

src/styles/animations.less
```diff
@@ -16,21 +16,25 @@
 }
 @keyframes modalSlideDown {
   0% {
-    margin-top: -25%;
-    opacity: .5;
+    // margin-top: -25%;
```
rhamilto: should probably go ahead and remove these?

src/styles/projects-summary.less
```diff
@@ -76,17 +76,35 @@
...
       }
     }
   }
 
   .services-view {
     .services-item {
-      color: @color-pf-white;
       flex: 0 1 33%;
       padding-left: 0px;
+      padding-right: 0px;
```
rhamilto: Can drop "px" since 0 is always 0.  Get the one on line 88, too, please.

sg00dwin:
Updated pr for review

spadgett:
@serenamarie125 @beanh66 Do the recently viewed items styled as in the design seem too strong? My eye goes right to them rather than the catalog content. Opinion?

screen shot 2017-04-24 at 12 52 28 pm

dtaylor113:
> @serenamarie125 @beanh66 Do the recently viewed items styled as in the design seem too strong? My eye goes right to them rather than the catalog content. Opinion?

I kinda liked the black background with blue hover circle, seemed more subtle.   If we do want a circle background, maybe a light grey so it's not so bright/strong?

sg00dwin:
My first thought was the same, that the white on dark appeared heavy. Something more subtle, along the lines of this would be my preference.
screen shot 2017-04-24 at 3 32 52 pm

beanh66:
@spadgett @dtaylor113 @sg00dwin Regardless of the circle color, I think the text should be white so these don't look disabled. I do agree the grey looks nice but the white circles don't bother me at all because we do want users to notice these and use them!  The other thing to note is that we will have real images instead of placeholder icons, which may look better on the white. I wonder if making the recently viewed items smaller and putting 4 across would make them less prominent. What do you guys think? I will be testing this at Summit as well so whatever is decided, we can get additional feedback. @serenamarie125 @cshinn any thoughts?

serenamarie125:
We need to consider the contrast with the icon.  Currently in these mocks, font icons are being used, but that's not going to be the case.

If they are too prominent, I'd rather see the icon size decreased, but use the white background.

serenamarie125:
total Fireworks hack ... but I mean something similar to this
screen shot 2017-04-25 at 1 24 03 pm
@beanh66 @sg00dwin @spadgett @cshinn @dtaylor113

jwforres:
I definitely like the smaller size better, but five of them feels like a
lot.  Maybe 3 or 4 centered and/or spaced out a little bit more.

beanh66:
@dtaylor113 Could you try this with 4 recently viewed items across and with smaller circles, icons, and font size?

spadgett:
> @dtaylor113 Could you try this with 4 recently viewed items across and with smaller circles, icons, and font size?

@sg00dwin since you own the PR, do you mind trying? You can make it 4 items by changing this:

https://github.com/openshift/origin-web-catalog/blob/master/src/services/recently-viewed-service-items.service.ts#L26-L27

beanh66:
Oops sorry @spadgett, I didn't realize this was yours @sg00dwin!

sg00dwin:
Here are versions with 4 items, smaller circles, icons, font-size.

**Width >992**
screen shot 2017-04-26 at 3 34 58 pm


**Width <992**
screen shot 2017-04-26 at 3 35 20 pm

dtaylor113:
My vote is for the Width <992 :+1:

beanh66:
@sg00dwin This looks good to me with the smaller recently viewed items 👍

sg00dwin:
As I mentioned in scrum today, my concern with displaying four can be seen when they have names that are longer than 3-4 words, especially at the smaller screens (>1200px). Showing 4 at at these widths can cause the titles to break the name awkwardly and grow quite tall. (screenshots and alternatives further down)






Currently the right sidebar has 3 variable widths based on the browser width.  325px, 375px, & 425px. 
I explored some options of varying the number of recent items based on sidebar widths.  These are the options that I think work well and reviewed with @jwforres. At 325px show 2, at 425px show 3, and at 475px show 4 (last expands sidebar by 50px). Thoughts?

**325px**



**425px**



**475px**


Here are links to full page screenshots at responsive browser widths.
[800px](https://drive.google.com/a/redhat.com/file/d/0B3F2EMy0q8RjVWIzeGlJc1hHZXM/view?usp=sharing)
[1200px](https://drive.google.com/a/redhat.com/file/d/0B3F2EMy0q8RjZzUzcmQtd1hva3c/view?usp=sharing)
[1600px](https://drive.google.com/a/redhat.com/file/d/0B3F2EMy0q8RjVGQyYjIyX1FGQVE/view?usp=sharing)

beanh66:
@sg00dwin I like that idea. The 4 look good at the max width, but due to long names, I'm with you on showing less for the smaller screens.

dtaylor113:
@sg00dwin :+1:  I like the variable items at diff screen sizes

dtaylor113:
Fixes #131

serenamarie125:
Hmmm @sg00dwin i'm wondering if we chop the name after a certain amount of characters?  Do we know what the max length is ?

sg00dwin:
@jwforres @spadgett  max character length?  The long name in my example is 65, for reference, if we want to truncate.

spadgett:
@serenamarie125 It's unstructured metadata, so the display name could theoretically be any length.

Looks like it has a `max-height:2.8em` set and will get cutoff after that, so we do handle it in CSS. But I don't think that rule is correctly applied here. @sg00dwin can you take a look?

spadgett:
Opened #189 for long display names in other places

spadgett:
@sg00dwin Since the long names appear to be a problem everywhere, I suggest we fix that once in follow on (#189). It's not specific to this PR.

sg00dwin:
@spadgett PTAL

screen shot 2017-05-01 at 4 32 39 pm
screen shot 2017-05-01 at 4 33 22 pm


screen shot 2017-05-01 at 4 26 03 pm
screen shot 2017-05-01 at 4 26 41 pm
screen shot 2017-05-01 at 4 29 17 pm

spadgett:
@sg00dwin You'll need to `npm run build` again

src/styles/landing-page.less
```diff
@@ -34,10 +34,10 @@
     top: @landing-offset-top;
     width: calc(100% ~"-" @landing-side-bar-width-sm);
   }
-  @media(min-width: @screen-md-min) {
+  @media(min-width: @screen-lg-min) {
     width: calc(100% ~"-" @landing-side-bar-width-md);
```
spadgett: Is this changing the sidebar width? We should name the variables so that `landing-side-bar-width-lg` is for `screen-lg-min`

cc @rhamilto since he originally set these values

src/styles/mixins.less
```diff
@@ -0,0 +1,17 @@
+// TODO: word-break mixins can be removed once the web console mixin is migrated to commons so these are accessible.
```
spadgett: These need to be moved from origin-web-console into origin-web-common instead of here. I'd rather do that in a follow on than have them defined twice.

src/styles/projects-summary.less
```diff
@@ -87,11 +87,88 @@
...
       &:hover {
-        color: @link-color;
-        text-decoration: underline;
+        color: @landing-side-bar-link-color-hover;
+        text-decoration: none;
       }
     }
   }
+  .services-view {
+    .services-item {
```
spadgett: Can we add these styles to the existing `.services-view` declaration above?

src/styles/projects-summary.less
```diff
@@ -87,11 +87,88 @@
...
+      }
+      .fa-cubes, .icon-wildfly {
+        position: relative;
+        right: -1px;
+      }
+      .icon-nodejs {
+        position: relative;
+        top: 1px;
+      }
+      .icon-ruby, .fa-question {
```
spadgett: I'm unclear why we need special positioning of these icons here and not in the main services area.

src/styles/services-view.less
```diff
@@ -33,19 +34,37 @@
...
+        box-shadow: 0 0 0 2px @link-color;
+      }
+
+      .services-item-name {
+        color: @link-color;
+      }
+    }
+
+    .services-item-icon {
+      width: 90px;
```
rhamilto: Alpha, please.  :)

src/styles/services-view.less
```diff
@@ -33,19 +34,37 @@
...
+      height: 90px;
+      border-radius: 50%;
+      &:after {
+        box-shadow: 0 0 0 2px transparent;
         border-radius: 100px;
         content: "";
         height: 90px;
         margin: -45px 0 0 -45px;
         position: absolute;
         left: 50%;
```
rhamilto: Fix my mis-alphabetization, please?  :)

src/styles/projects-summary.less
```diff
@@ -87,11 +87,88 @@
...
+          display: none; // display first 3
+        }
+      }
+      @media (min-width: @screen-xl-min) {
+        &:nth-child(n+3) {
+          display: flex; // display first 4
+        }
+      }
+      .fa-cubes, .icon-wildfly {
+        position: relative;
```
rhamilto: Can the position: relative be shared amongst the rules that need it so it isn't duplicated?

src/styles/variables.less
```diff
@@ -11,16 +11,22 @@
...
+@landing-side-bar-width-lg:                           475px;
+@landing-side-bar-width-md:                           425px;
 @landing-side-bar-width-sm:                           325px;
 @navbar-height:                                       72px;
 @saas-hover:                                          fade(@color-pf-blue-700, 35%);
 @saas-border-bottom-color:                            fade(@color-pf-blue-700, 60%);
 @saas-text:                                           @color-pf-white;
 @search-area-border:                                  fade(@color-pf-blue-700, 60%);
-@screen-xl-min:                                       1800px;
+@screen-lg-max:                                       (@screen-xl-min - 1);
```
rhamilto: Alpha

rhamilto: Since [these same vars exist in console](https://github.com/openshift/origin-web-console/blob/master/app/styles/_variables.less#L104-L108), should these be moved to common so they're not declared twice?

src/styles/projects-summary.less
```diff
@@ -87,11 +87,88 @@
...
+        }
+      }
+      .services-item-name {
+        color: @color-pf-white;
+        font-size: 12px;
+        font-weight: 300;
+        padding: 0 5px;
+      }
+    }
+    .services-items-inner {
```
rhamilto: Please double check your rebase.  I removed the .services-items-inner and .services-items-inner-shadow-covers rules in [#188](https://github.com/openshift/origin-web-catalog/pull/188/files#diff-d4de2c7595ae630bb7ebb283c569efddL96).

src/styles/mixins.less
sg00dwin: Spoke with @rhamilto and created an issue to track items that will be shared. 
https://github.com/openshift/origin-web-console/issues/1500

I'll remove this mixin file.

src/styles/projects-summary.less
sg00dwin: The sidebar icons are smaller and with white circle some looked slightly off center. The  ? and ruby looked slightly overcorrected so I adjusted those 1px. 
screen shot 2017-05-02 at 1 46 06 pm

Wildfly is actually off center in both places so I moved it's rule to services-view.less
screen shot 2017-05-02 at 1 25 28 pm
screen shot 2017-05-02 at 1 54 36 pm

**Adjusted**
screen shot 2017-05-02 at 1 56 37 pm
screen shot 2017-05-02 at 1 56 25 pm

sg00dwin: I removed my rules for nodejs and cubes after looking more closely, they where fine.

sg00dwin:
@spadgett @rhamilto pr updated

PR #175, opened 2017-04-25
[WIP] Change saas offerings to horizontal scrolling

cshinn:
Changes mobile version of SaaS offerings from a list to horizontally scrolling tiles. Discoverability of >2 items is currently low and requires additional thoughts on design

untitled

### DIFF
--- a/dist/less/saas-list.less
+++ b/dist/less/saas-list.less
@@ -9,24 +9,24 @@
 
 .saas-offerings-container {
 
-  .saas-list {
-    justify-content: space-around;
+  .saas-list{
+    display: flex;
+    overflow-x: scroll;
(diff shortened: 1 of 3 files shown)


serenamarie125:
Looking good @cshinn !  

@rhamilto @jwforres @spadgett are you guys doing something similar anywhere else in OpenShift that we can use?

spadgett:
We don't have anything similar currently, but there is the Bootstrap carousel we could use if we wanted.

https://angular-ui.github.io/bootstrap/#!#carousel

cshinn:
There is some turmoil about the design of this section, so I'm going to close this until that gets sorted out.

PR #181, opened 2017-04-27
Fix svg display bug

### LINKED ISSUE #180: Font Icons are not showing up in SaaS area in some versions of Chrome & Firefox
Looks like newer versions of Chrome & Firefox cause issues with the font icons.

Here is Chrome:
screen shot 2017-04-27 at 12 01 30 pm

Here is Safari running the same build:
screen shot 2017-04-27 at 12 01 43 pm

cshinn:
SVGs now display properly in most recent version of Chrome and Firefox

Closes #180

### DIFF
--- a/dist/less/saas-list.less
+++ b/dist/less/saas-list.less
@@ -85,7 +85,7 @@
         }
 
         img {
-          max-height: 1em;
+          height: 1em;
           max-width: 100%;
         }
       }
--- a/dist/origin-web-catalogs.css
+++ b/dist/origin-web-catalogs.css
@@ -756,7 +756,7 @@ body.overlay-open {
   }
 }
 .saas-offerings-container .saas-list .card-content .card-icon img {
-  max-height: 1em;
+  height: 1em;
   max-width: 100%;
 }
 .saas-offerings-container .saas-list .card-content .card-title {
--- a/src/styles/saas-list.less
+++ b/src/styles/saas-list.less
@@ -85,7 +85,7 @@
         }
 
         img {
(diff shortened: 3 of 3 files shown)


rhamilto:
[test]

cshinn:
@rhamilto I've just noticed that this causes some weird effects in IE11. SVGs appear slightly flattened. I think it's a flexbug

rhamilto:
> @rhamilto I've just noticed that this causes some weird effects in IE11. SVGs appear slightly flattened. I think it's a flexbug

@cshinn for today, I think it's ok to roll with the IE11 bug.  We'll want to circle back to that in the near future.  Please open an issue for it?  Ideally we'd be following the advice in https://css-tricks.com/scale-svg/ and not using width and height attributes to scale the SVGs...

cshinn:
This could theoretically be any kind of image file though, couldn't it?

rhamilto:
> This could theoretically be any kind of image file though, couldn't it?

Good point.  So we'll need to figure out the bug (but not necessarily now since this seems to resolve the immediate issue).  I'm in the process of staging the data as pictured so I can better see what's happening.

cshinn:
@rhamilto just an FYI, This fix works on the most recent chrome and ff in fedora as well as on edge. Looks like the only one that still looks weird is IE 11, which looked fine before the fix.

dtaylor113:
Hi @cshinn, after you rebase that Travis error should go away :-)

rhamilto:
Closing in favor in #187

PR #186, opened 2017-04-28
Fixes 'SaaS offerings should not be using a click handler'

### LINKED ISSUE #185: SaaS offerings should not be using a click handler to open the URL
We are only planning to support opening a URL right now, we should not be using click handlers for that, we should be using appropriate links with hrefs so that right-click -> open in new tab behaviors are not broken.

dtaylor113:
Fixes #185


src/components/saas-list/saas-list.html
```diff
@@ -2,7 +2,7 @@
   <h1 ng-if="$ctrl.saasTitle">{{$ctrl.saasTitle}}</h1>
   <div class="saas-list" items="$ctrl.saasOfferings">
     <div class="card" ng-repeat="item in $ctrl.saasOfferings">
-      <a href="" class="card-content" ng-click="$ctrl.handleClick(item)">
+      <a href="{{item.url}}" class="card-content">
```
spadgett:
```html
<a ng-href="{{item.url}}" target="_blank" class="card-content">
```

dtaylor113: Fixed, thanks

spadgett:
Still missing `target="_blank"` :)

dtaylor113:
> Still missing target="_blank" :)

arrggg...thanks, fixed :-)

Issue #174, opened 2017-04-24
When sorting Catalog Items, use localeCompare for Display Name


dtaylor113:
Here is how we are sorting category items:

https://github.com/openshift/origin-web-catalog/blob/master/src/services/catalog.service.ts#L58

Fix the display name to pass a compare fn that uses localeCompare. Unique name and kind don't necessarily need localeCompare since they're only Latin characters, but display name does.

spadgett:
We should also use `{ sensitivity: 'base' }` with `localeCompare` rather than `toLowerCase`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare

@jwforres We should start using `localeCompare` for display names generally in the web console. It doesn't look like lodash uses it in `sortBy`.

cdcabrera:
Closed with #191

PR #190, opened 2017-04-28
WIP Locale Comparison Issue #174

cdcabrera:
Case insensitive Locale comparison across all fields for simplicity. Addresses issue #174 . If locale needs to be more focused, can do


src/services/catalog.service.ts
```diff
@@ -55,7 +55,21 @@ export class CatalogService {
 
     // Perform a case-insensitive sort on display name, falling back to kind
     // and metadata.name for a stable sort when items have the same display name.
-    return _.sortByAll(items, [(item: any) => item.name.toLowerCase(), 'resource.kind', 'resource.metadata.name']);
+    items = items.sort((item1: any, item2: any) => {
+      let comparison = item1.name.localeCompare(item2.name, undefined, {sensitivity: 'base'});
+
+      if (comparison === 0) {
+        comparison = item1.resource.kind.localeCompare(item2.resource.kind, undefined, {sensitivity: 'base'});
```
spadgett: I'd feel more comfortable using `_.get` here and below to avoid an error if the property doesn't exist.

```js
_.get(item 'resource.kind', '')
```

https://lodash.com/docs/3.10.1#get

cdcabrera: can do

spadgett: Thanks, probably should do it for `item.name` too just in case

spadgett:
@cdcabrera Thanks, can you squash your commits and `npm run build` to update the dist folder?

cdcabrera:
almost there, resolving conflicts

spadgett:
If the conflict is in dist, should be able to rebase and `npm run build` again

cdcabrera:
reopened under PR#191

PR #193, opened 2017-05-01
Optionally show templates in the service catalog

spadgett:
https://trello.com/c/2A7x5Pbc

Optionally adds templates to the service catalog when the template broker is not running. The template dialog itself is handled in the web console by https://github.com/openshift/origin-web-console/pull/1492

### DIFF
--- a/app/pages/home/homePageController.ts
+++ b/app/pages/home/homePageController.ts
@@ -25,20 +25,12 @@ export class HomePageController {
   };
 
   public update() {
-    this.Catalog.getCatalogItems().then( (catalogServiceItems: any) => {
+    this.Catalog.getCatalogItems(false).then( (catalogServiceItems: any) => {
       this.ctrl.catalogItems = catalogServiceItems;
     }, () => {
       this.ctrl.catalogItems = {};
     });
 
     this.ctrl.saasOfferings = this.constants.SAAS_OFFERINGS;
-  };
-
-  public navToProject = (project: any) => {
-    this.$state.go('projects/' + project.metadata.name);
-  };
-
-  public navToProjects = () => {
-    this.$state.go('projects');
   };
 }
--- a/src/components/landing-page/landing-page.controller.ts
+++ b/src/components/landing-page/landing-page.controller.ts
@@ -17,14 +17,15 @@ export class LandingPageController implements angular.IController {
     this.ctrl.orderingPanelVisible = false;
 
     this.$scope.$on('open-overlay-panel', (event: any, item: any) => {
-      let kind = _.get(item, 'resource.kind');
-      if (kind === 'ImageStream') {
-        this.ctrl.selectedImageStream = item;
-        this.ctrl.selectedServiceClass = null;
-      } else {
-        this.ctrl.selectedImageStream = null;
-        this.ctrl.selectedServiceClass = item;
+      if (item.resource.kind === 'Template') {
+        let cb = this.ctrl.onTemplateSelected();
+        if (cb) {
+          cb(item.resource);
+        }
+        return;
       }
+
+      this.ctrl.selectedItem = item;
       this.ctrl.orderingPanelVisible = true;
     });
   }
@@ -36,8 +37,7 @@ export class LandingPageController implements angular.IController {
   }
 
   public closeOrderingPanel = () => {
-    let item: any = this.ctrl.selectedImageStream || this.ctrl.selectedServiceClass;
-    this.RecentlyViewed.addItem(item.resource.metadata.uid);
+    this.RecentlyViewed.addItem(this.ctrl.selectedItem.resource.metadata.uid);
     this.ctrl.orderingPanelVisible = false;
-  };
+  }
 }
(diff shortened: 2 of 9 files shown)


spadgett:
@dtaylor113 I've added tags to `IServiceItem`

app/pages/home/homePageController.ts
```diff
@@ -25,20 +25,12 @@ export class HomePageController {
...
   };
-
-  public navToProject = (project: any) => {
-    this.$state.go('projects/' + project.metadata.name);
-  };
-
-  public navToProjects = () => {
-    this.$state.go('projects');
-  };
 }
```
dtaylor113: Was navToProject and navToProjects not used?

spadgett: No longer used

src/components/landing-page/landing-page.controller.ts
```diff
@@ -17,14 +17,12 @@ export class LandingPageController implements angular.IController {
...
     this.$scope.$on('open-overlay-panel', (event: any, item: any) => {
-      let kind = _.get(item, 'resource.kind');
-      if (kind === 'ImageStream') {
-        this.ctrl.selectedImageStream = item;
-        this.ctrl.selectedServiceClass = null;
-      } else {
-        this.ctrl.selectedImageStream = null;
-        this.ctrl.selectedServiceClass = item;
+      if (item.resource.kind === 'Template') {
+        this.ctrl.onTemplateSelected({ template: item.resource });
```
dtaylor113: Do we need a null test to see if 'onTemplateSelected' was defined?

app/pages/home/homePageController.ts
```diff
@@ -25,20 +25,12 @@ export class HomePageController {
   };
 
   public update() {
-    this.Catalog.getCatalogItems().then( (catalogServiceItems: any) => {
+    this.Catalog.getCatalogItems(false).then( (catalogServiceItems: any) => {
```
dtaylor113: I do not see where 'this.Catalog.getCatalogItems(true)' is ever called, I assume it's done in the console?

spadgett: It's called in https://github.com/openshift/origin-web-console/pull/1492

spadgett: https://github.com/openshift/origin-web-console/pull/1492/files#diff-9b83bfc0c5e34ad58889288622257b70R30

src/components/landing-page/landing-page.controller.ts
spadgett: It's always defined even if there is no `on-template-selected` attribute.

dtaylor113: Where is it defined as a function if no `on-template-selected` attribute is passed in?  -thanks

spadgett: It's how `&` bindings work in Angular. They will always be defined.

src/services/catalog.service.ts
```diff
@@ -21,23 +21,30 @@ export class CatalogService {
...
+    }
+
+    return this.$q.all(promises).then((values) => {
       let serviceClasses: any = values.serviceClasses.by("metadata.name");
       let imageStreams: any = values.imageStreams.by("metadata.name");
-      return this.convertToServiceItems(serviceClasses, imageStreams);
+      let templates: any = values.templates ? values.templates.by("metadata.name") : {};
+      return this.convertToServiceItems(serviceClasses, imageStreams, templates);
     }, () => {
       this.logger.log("Error Loading Catalog Items");
```
jwforres: I feel like we should not be bombing out if one thing fails to load but the other two succeed, seems like we could at least show what was able to load and then show a message that not everything loaded successfully.  But we can open a separate issue to address this. Note that $q.all fails out at the first failure, so you'd need a different way of handling this.

Just considering the fact that the svc cat api server is separate and could potentially be down even though the master is available.

src/components/landing-page/landing-page.controller.ts
spadgett: @dtaylor113 I'll change it to use the other callback style that @jeff-phillips-18 uses for consistency

src/services/catalog.service.ts
spadgett: I'm OK with that, but I think we need to break this into separate methods. It also feels wrong to return a partial result with no indication that something failed.

spadgett: Maybe one call for each kind of item and then one call to concat and sort them?

jwforres:
the promise can be rejected with the data you were able to retrieve as the
argument, or as part of the rejection response in combination with more
error details if necessary

jwforres:
or this, whatever ends up being cleaner

spadgett:
@dtaylor113 I changed the callback to a function reference instead of an object literal

src/services/catalog.service.ts
spadgett: Opened #194 for this

PR #200, opened 2017-05-02
Utilizing origin-web-common variables and mixins

rhamilto:
Once merged, I'll need a new release cut so I can make use of these changes in console.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -39,7 +39,7 @@
     "patternfly": "3.21.0",
     "jquery": "2.1.4",
     "lodash": "3.10.1",
-    "origin-web-common": "~0.0.16"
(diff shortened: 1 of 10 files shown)


bower.json
```diff
@@ -39,7 +39,7 @@
     "patternfly": "3.21.0",
     "jquery": "2.1.4",
     "lodash": "3.10.1",
-    "origin-web-common": "~0.0.16"
+    "origin-web-common": "~0.0.19"
```
spadgett: Need to bump this in package.json too

rhamilto: Fixed.  Thanks.  I should have noticed that.

PR #205, opened 2017-05-03
New Categorization of Catalog Items

### LINKED ISSUE #29: Need to review subcategory matching logic

### LINKED ISSUE #192: Services matching subcategories appear when parent category doesn't match
When a subcategory matches, it's displayed even, if there's no match for the parent category:

For Example:
Given categories:
```javascript
var categories: any = [
  {id: 'amazon', label: 'Amazon' , subCategories: [
    {id: 'database', label: 'Database'}
  ]},
  {id: 'other', label: 'Other' , subCategories: [
    {id: 'other', label: 'Other'}
  ]}
];
```
and services with the following (abbreviated) json

```json
{
  "metadata": {
    "name": "amazon-database"
  },
  "osbTags": [
    "amazon",
    "database"
  ]
},
{
  "metadata": {
    "name": "plain-database"
  },
  "osbTags": [
    "database"
  ]
}
```

The plain database, not associated with Amazon, will show up under Amazon -> Databases.  I would have thought the parent tag must also be on the service for it to be included under that main category and sub category.

It may also occur that if the parent tag was different, ie parent-not-amazon, it might also show under Amazon, but I haven't tested it.

dtaylor113:
Please refer to: https://trello.com/c/xbWZVIrz for detailed information about how the categorization algorithm has changed.

Fixes #192 and Fixes #29

### DIFF
--- a/src/services/catalog.service.ts
+++ b/src/services/catalog.service.ts
@@ -16,7 +16,6 @@ export class CatalogService {
     this.$filter = $filter;
     this.$q = $q;
     this.constants = constants;
-    this.categories = this.constants.SERVICE_CATALOG_CATEGORIES;
     this.dataService = DataService;
     this.logger = Logger;
   }
@@ -45,6 +44,8 @@ export class CatalogService {
   }
 
   public convertToServiceItems(serviceClasses: any, imageStreams: any, templates: any) {
+    this.categories = angular.copy(this.constants.SERVICE_CATALOG_CATEGORIES);
+
     // Convert service classes to ServiceItem
     let items: any = _.map(serviceClasses, (serviceClass) => {
       return this.getServiceItem(serviceClass);
@@ -96,30 +97,54 @@ export class CatalogService {
     return new TemplateItem(resource, this);
   }
 
-  public getCategoriesBySubCategories(tags: any) {
+  public getCategoriesBySubCategories(itemTags: any) {
     let catsBySubCats = {};
     let otherId = 'other';
+    let filteredSubCats: any;
+    let otherMainCategory: any = {
+      id: 'other', label: 'Other', subCategories: [
+        {id: 'other', label: 'Other'}
+      ]
+    };
 
-    _.each(tags, (tag) => {
-      _.each(this.categories, (category) => {
-        let subCat: any = _.find(category.subCategories, (subCategory: any) => {
-              return subCategory.id === tag || _.includes(subCategory.categoryAliases, tag);
-            }
-        );
-        if (subCat) {
-          catsBySubCats[subCat.id] = category.id;
-          return false;
+    _.each(this.categories, (category) => {
+      if (category.tags) {
+        if (!_.isEmpty(this.getMatchingTags(category.tags, itemTags))) {
+          filteredSubCats = this.filterSubCatsByTags(category.subCategories, itemTags);
+          if (!_.isEmpty(filteredSubCats)) {
+            _.each(filteredSubCats, (subCat) => {
+              this.categorize(catsBySubCats, category, subCat.id);
+            });
+          } else {
+            this.categorize(catsBySubCats, category, otherId);
+          }
         }
-      });  // .ea category
-    });  // .ea tag
+      } else {
+        filteredSubCats = this.filterSubCatsByTags(category.subCategories, itemTags);
+        if (!_.isEmpty(filteredSubCats)) {
+          _.each(filteredSubCats, (subCat) => {
+            this.categorize(catsBySubCats, category, subCat.id);
+          });
+        }
+      }
+    });  // .ea category
+
     if (_.isEmpty(catsBySubCats)) {
-      catsBySubCats[otherId] = otherId;
+      this.categorize(catsBySubCats, otherMainCategory, otherId);
     }
+
     return catsBySubCats;
   }
 
   public hasCategory(item: any, category: string) {
-    return _.includes(item.catsBySubCats, category);
+    let found: boolean = false;
+
+    _.each(item.catsBySubCats, (cats: any) => {
+      found = _.includes(cats, category);
+      return !found;  // to break out of _.each when found
+    });
+
+    return found;
   }
 
   public hasSubCategory(item: any, subCategory: string) {
@@ -139,7 +164,7 @@ export class CatalogService {
     _.each(categories, (category) => {
       let retSubCategories: any = _.filter(category.subCategories, (subCategory: any) => {
         return _.some(items, (item: any) => {
-          return this.hasSubCategory(item, subCategory.id);
+          return this.hasSubCategory(item, subCategory.id) && this.hasCategory(item, category.id);
         });
       });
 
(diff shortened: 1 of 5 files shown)


src/services/catalog.service.ts
```diff
@@ -152,6 +166,56 @@ export class CatalogService {
...
+  };
+
+  private filterSubCatsByTags(subCats: any, tags: any) {
+    return _.filter(subCats, (subCat: any) => {
+      return !_.isEmpty(this.getMatchingTags(subCat.tags, tags));
+    });
+  }
+
+  private addOtherMainCategory() {
+    let otherCategory: any = _.find(this.categories, (category: any) => {
```
jwforres: you should use the shorthand version of _.find
```
_.find(this.categories, {id: 'other'});
```

src/services/catalog.service.ts
```diff
@@ -152,6 +166,56 @@ export class CatalogService {
...
+          id: 'other', label: 'Other', subCategories: [
+            {id: 'other', label: 'Other'}
+          ]
+        }
+      );
+    }
+  }
+
+  private addOtherSubCategory(category: any) {
+    let otherSubCategory: any = _.find(category.subCategories, (subCat: any) => {
```
jwforres: same, use the shorthand _.find

src/services/catalog.service.ts
```diff
@@ -96,25 +96,39 @@ export class CatalogService {
...
+    _.each(this.categories, (category) => {
+      if (category.tags) {
+        if (!_.isEmpty(this.getMatchingTags(category.tags, itemTags))) {
+          filteredSubCats = this.filterSubCatsByTags(category.subCategories, itemTags);
+          if (!_.isEmpty(filteredSubCats)) {
+            _.each(filteredSubCats, (subCat) => {
+              catsBySubCats[subCat.id] = category.id;
+            });
+          } else {
+            catsBySubCats[otherId] = category.id;
```
jwforres: this is going to be broken if you end up categorizing an item into more than one main category's "Other", i end up in Foo -> Other and Bar -> Other, the second one that matches wins, where we really want it to be in both

I think in general we should not be assuming that subcategory id's are unique across main categories, a subcategory id just needs to be unique within its main category

dtaylor113: Sorry, not following.  Using this data:

Categories =
```
{id: 'mycata', label: 'My Category A', tags: ['mycat-a'], subCategories: [
    {id: 'database', label: 'Database', tags: ['database']}
  ]},
  {id: 'mycatb', label: 'My Category B', tags: ['mycat-b'], subCategories: [
    {id: 'database', label: 'Database', tags: ['database']}
  ]}
```
And these serviceClasses:
```
"mycat-a-other": {
    kind: 'ServiceClass',
    "metadata": {
      "name": "mycat-a-other",
      "uid": "11",
    },
    osbTags: [
      "mycat-a",
      "foobar"
    ]
  },
  "mycat-b-other": {
    kind: 'ServiceClass',
    "metadata": {
      "name": "mycat-b-other",
      "uid": "11",
    },
    osbTags: [
      "mycat-b",
      "foobar"
    ]
  }
```
The items were categorized under` 'My Project A' -> Other->mycat-a-other `and` 'My Project B'->Other->mycat-b-other`.

src/services/catalog.service.ts
```diff
@@ -152,6 +166,56 @@ export class CatalogService {
 
     return retCategories;
   }
+
+  /*
+   * Catalog service dynamically adds 'Other' main and sub categories.
+   * This method removes all 'Other' categories.  Used for unit testing
+   */
+  public removeOtherCategories() {
```
dtaylor113: Fixed, moved function into unit test.  I originally left it here because I think in the future, if we ever get to watching & dynamically updating items & categories, this function would be useful to reset the categories.

src/services/catalog.service.ts

src/services/catalog.service.ts

src/services/catalog.service.ts
jwforres: I don't see how that is possible.  catsBySubCats is a single map and your keys are the subcategory id's, what you are storing as the value is the main category id, not an array of main category ids. So the outer loop is going to initially set `catsBySubCats['other'] = "mycat-a"` and on the the second iteration of the outer loop it is going to set `catsBySubCats['other'] = "mycat-b"` which is overwriting what you set before, not adding to it.

dtaylor113: catsBySubCats is stored per item.

jwforres: sorry yes, your particular example is fine, thats not the example i'm saying is broken. any single item can be categorized into many different main categories and subcategories.  if you instead had:

```
"mycat-a-other": {
    kind: 'ServiceClass',
    "metadata": {
      "name": "mycat-a-other",
      "uid": "11",
    },
    osbTags: [
      "mycat-a",
      "mycat-b",
      "foobar"
    ]
  }
```

Then what should happen is that it appears both in `My Category A -> Other` AND in `My Category B -> Other`  but what will happen is that it will only show up in the second one.

jwforres: Similar case that is broken:

categories:
```
{id: 'mycata', label: 'My Category A', subCategories: [
    {id: 'database', label: 'Database', tags: ['database']}
  ]},
  {id: 'mycatb', label: 'My Category B', subCategories: [
    {id: 'database', label: 'Database', tags: ['database']}
  ]}
```

```
"mycat-a-other": {
    kind: 'ServiceClass',
    "metadata": {
      "name": "mycat-a-other",
      "uid": "11",
    },
    osbTags: [
      "database"
    ]
  }
```

This should show up under both `My Category A -> Database` AND `My Category B -> Database`

test/services-view.spec.ts
```diff
@@ -90,7 +98,68 @@ describe('servicesView', () => {
     expect(jQuery(element).find('.services-item-name').length).toBe(1);
   });
 
+  it("should categorize items correctly when main category has 'tags'", () => {
+    // remove any dynamically added 'Other' categories
+    removeOtherCategories(Constants.SERVICE_CATALOG_CATEGORIES);
```
jwforres: I still question why this is necessary for your test

dtaylor113: The unit test's before() categorizes using mock items and dynamically adds 'Other' categories where needed.  Need to remove those before categorizing using the items specific to this unit test.

jwforres: we shouldn't be directly modifying the Constants.SERVICE_CATALOG_CATEGORIES from source or tests, we should be cloning that object first

src/services/catalog.service.ts
dtaylor113: Ok, fixed.  Added unit test for this.

test/services-view.spec.ts
dtaylor113: Fixed

dtaylor113:
Added #215 as a possible follow up issue.

dtaylor113:
@jwforres I believe I addressed all review issues -thanks!

jwforres:
definitely need to stop re-calculating this every time we switch categories, so if you can start looking at #215 next that would be good

PR #207, opened 2017-05-03
Fix extra whitespace at bottom of form when creating project

### LINKED ISSUE #204: div.form-group.invisible creates empty space at bottom of create from builder form
screen shot 2017-05-03 at 9 15 08 am

@spadgett, based on your comment in the template, I'm guessing you determined this was the lesser of evils; however, it becomes a bit more obvious when only the right column scrolls (which I'm still working on).

spadgett:
Fixes #204


src/components/create-from-builder/create-from-builder-configure.html
```diff
@@ -90,10 +90,11 @@
 
     <!--
       Only show the link for existing projects. It will be broken for new
-      projects.  Make `invisible` instead of using an ng-if so that the dialog
-      doesn't resize when the message appears.
+      projects.  Use class `invisible` when the project list is still loading
+      so the dialog doesn't resize.
     -->
-    <div ng-class="{ invisible: !$ctrl.selectedProject.metadata.uid || !$ctrl.istag }" class="form-group">
+    <div ng-hide="($ctrl.selectedProject && !$ctrl.selectedProject.metadata.uid) || !$ctrl.istag"
```
jwforres: this also changes the behavior from invisible to hidden when `!$ctrl.istag` , just making sure that was intentional

spadgett: It was intentional although maybe it's better to keep it invisible. Really there should always be an istag, though

Let me change it back.

spadgett:
@jwforres updated to use class invisible when no istag

PR #208, opened 2017-05-03
Hiding .landing and .landing-side-bar scrollbars when panel overlay i…

### LINKED ISSUE #120: Scrollbar can appear briefly above overlay panel
I believe this only happens on macOS with autohiding scrollbars.

screen shot 2017-04-07 at 8 50 14 pm

cc @rhamilto

rhamilto:
…s open

Fixes #120 

@jeff-phillips-18, I assume reducing the timeout from 500ms to 100ms is ok?

### DIFF
--- a/src/components/overlay-panel/overlay-panel.controller.ts
+++ b/src/components/overlay-panel/overlay-panel.controller.ts
@@ -2,14 +2,9 @@ import * as angular from 'angular';
 import * as $ from 'jquery';
 
 export class OverlayPanelController implements angular.IController {
-  static $inject = ['$timeout'];
-
   public ctrl: any = this;
-  private $timeout: any;
 
-  constructor ($timeout: any) {
-    this.$timeout = $timeout;
-    this.ctrl.showOverlayPanel = false;
+  constructor () {
     this.ctrl.shown = false;
   }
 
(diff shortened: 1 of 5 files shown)


spadgett:
@jeff-phillips-18 Do you remember why we have the timeout? Should probably add a comment explaining why it's needed

jeff-phillips-18:
@spadgett The timeout was due to a scrollbar effect on the page. That scrollbar is no longer being shown so the timeout can be removed. In fact, both the showDialog and hideDialog can be reduced to simply setting the shown flag and nothing more.

rhamilto:
Thanks, @jeff-phillips-18.  $timeout removed and the bug is full fixed!

src/components/overlay-panel/overlay-panel.controller.ts
```diff
@@ -41,19 +41,11 @@ export class OverlayPanelController implements angular.IController {
 
   private showDialog = () => {
     this.ctrl.shown = true;
-
-    this.$timeout(() => {
-      this.ctrl.showOverlayPanel = true;
```
spadgett: It's not clear to me where this value was used, but looks like the only other reference to it is in the constructor. @rhamilto Can you remove that?

rhamilto: Done.  I've also removed $timeout in its entirety since it's not longer in use.

Issue #201, opened 2017-05-02
service items hover defects


rhamilto:
In Chrome, I'm seeing a weird border on the left-most edge:

![screen shot 2017-05-02 at 3 41 50 pm](https://cloud.githubusercontent.com/assets/895728/25636019/133f899a-2f4e-11e7-9e5d-c9cc75ed03c9.PNG)

In Firefox, the blue circle is getting clipped on the right edge:
![screen shot 2017-05-02 at 3 42 03 pm](https://cloud.githubusercontent.com/assets/895728/25636030/1f041534-2f4e-11e7-83da-3f7790e16a1e.PNG)

If there is only 1 recently viewed item, it centers:
![screen shot 2017-05-02 at 3 43 51 pm](https://cloud.githubusercontent.com/assets/895728/25636059/3621a592-2f4e-11e7-91a6-ea57dfd9eefd.PNG)

PR #209, opened 2017-05-04
Switch icon hover circle to border from box-shadow and make service-i…

sg00dwin:
…tem display:block to enable word-break
Also
- img max height and width needed to be set
- ` .services-item-icon` flex removed and height set. flex was preventing word-break from functioning

Fixes https://github.com/openshift/origin-web-catalog/issues/201


sg00dwin:
@rhamilto PTAL
cc @spadgett

dtaylor113:
It would be great to see a screenshot :-)

src/styles/services-view.less
```diff
@@ -43,19 +43,17 @@
     }
 
     .services-item-icon {
-      border-radius: 50%;
-      height: 90px;
-      width: 90px;
       &:after {
         border-radius: 100px;
-        box-shadow: 0 0 0 2px transparent;
+        border: 2px solid
```
rhamilto: Nit:  alpha

sg00dwin:
@spadgett @rhamilto Updated pr for review

Added gradient fade and overview hidden to handle long names.

screen shot 2017-05-04 at 2 21 32 pm
screen shot 2017-05-04 at 2 21 11 pm

sg00dwin:
pr rebased

beanh66:
@sg00dwin @rhamilto I think we should use an ellipses when text gets cut off. In this case we would not use the fade, we would just show the ... after three lines of text and then show the full name in a tooltip on hover. This would be in line with PF and we should also be consistent with what we decide here: https://github.com/openshift/origin-web-catalog/pull/214 @serenamarie125

spadgett:
@beanh66 We can only do this if we decide to truncate to a certain number of characters instead of cleanly at the end of the 3rd line. There's no good way that I know in CSS to use an ellipsis to truncate a multiline string. @rhamilto keep me honest here

This will be a bigger problem for #214 where we can't truncate by characters since it will break any links that get truncated the middle. We'd have to remove the link highlighting, and it also means we might truncate in the middle of a line.

rhamilto:
> @beanh66 We can only do this if we decide to truncate to a certain number of characters instead of cleanly at the end of the 3rd line. There's no good way that I know in CSS to use an ellipsis to truncate a multiline string. @rhamilto keep me honest here

To my knowledge, there is no cross-browser compatible solution as described, which is why we went with the fade out effect here and in in #214.  :(

beanh66:
@spadgett @rhamilto Can we cut these titles off at 40 characters then show the ellipses? We would still allow users to see the full title on hover. Thoughts @ncameronbritt @serenamarie125?

sg00dwin:
@spadgett updated with overflow gradient removed.

spadgett:
@serenamarie125 @beanh66 @rhamilto I asked @sg00dwin to remove the gradient for now and add the ellipsis in a follow-on. We need to find a good way to do that both here and the overlay panel, but unfortunately it's a little tricky.

spadgett:
@rhamilto Can you take another look?

rhamilto:
[merge]

serenamarie125:
agreed @spadgett I think it's also worth us discussing best practices, and making sure our recommendations are documented appropriately.  I know that we will still need to address, but if we document best practices, hopefully these edge cases will be minimized.

PR #211, opened 2017-05-04
Optimized nameTaken validation in Order Service

### LINKED ISSUE #88: In Order Service optimize nameTaken validation
In `<order-service>` since we already have the list of projects, we check name taken before trying to create the project. We had a directive that does something like this in origin-web-console:

https://github.com/openshift/origin-web-console/blob/e7a0c0a8f703d5429f70b78223abb31856a66670/app/scripts/directives/oscUnique.js

dtaylor113:
Using `<osc-unique>` directive to validate unique project names.

Fixes #88


src/components/select-project/select-project.controller.ts
```diff
@@ -70,6 +70,7 @@ export class SelectProjectController implements angular.IController {
         }
       };
       this.ctrl.projects = _.sortBy(response.by('metadata.name'), this.$filter('displayName'));
+      this.ctrl.existingProjectNames = this.getExistingProjectNames();
```
spadgett: Should be able to use

```js
this.ctrl.existingProjectNames = _.map(this.ctrl.projects, 'metadata.name');
```

dtaylor113: Thanks, fixed

PR #214, opened 2017-05-05
Making only right side panel contents scrollable, left side panel content clip

### LINKED ISSUE #196: Description should not scroll with dialog content
The description in the left panel should not scroll with the rest of the dialog content. See discussion in #177

rhamilto:
Fixes #196 

At a tall, wide desktop resolution, very long descriptions clip if they exceed the max-height of the panel
screen shot 2017-05-09 at 4 51 04 pm

At short, wide desktop resolution, very long descriptions clip and the right side panel scrolls if the contents are taller than the panel
screen shot 2017-05-09 at 4 51 17 pm
screen shot 2017-05-09 at 4 51 27 pm

At mobile, there is no change in the behavior and the two columns collapse to one and there is an inner scroll
screen shot 2017-05-05 at 1 56 06 pm

Whatchyall think?  @jwforres, @spadgett, @sg00dwin, @serenamarie125, @beanh66, etc.

### DIFF
--- a/dist/less/catalog.less
+++ b/dist/less/catalog.less
@@ -2,6 +2,7 @@
 
 @import './animations.less';
 @import './landing-page.less';
+@import './mixins.less';
 @import './order-service.less';
 @import './overlay-panel.less';
 @import './projects-summary';
--- a/dist/less/mixins.less
+++ b/dist/less/mixins.less
@@ -1,0 +1,25 @@
+.scroll-shadows-horizontal(@shadow-width: 75%, @shadow-opacity: 0.2) {
+  background-attachment: scroll;
+  background-image: radial-gradient(ellipse at top, rgba(0, 0, 0, @shadow-opacity) 0%, rgba(0, 0, 0, 0) @shadow-width), radial-gradient(ellipse at bottom, rgba(0, 0, 0, @shadow-opacity) 0%, rgba(0, 0, 0, 0) @shadow-width);
+  background-position: 0 0, 0 100%;
+  background-repeat: no-repeat;
+  background-size: 100% 6px;
+}
+
+.scroll-shadows-horizontal-covers(@shadow-cover-bg-color: rgba(255,255,255,1), @shadow-cover-bg-color-transparent: rgba(255,255,255,0)) {
+  background-attachment: local;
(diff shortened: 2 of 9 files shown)


serenamarie125:
@rhamilto agree with your suggestions as to when clip versus scroll!
But I'd like to suggest some visual suggestions after speaking with both Colleen & Kyle.  When clipping, use an ellipses rather than fading out.  Fading makes it feel like the content scrolls.

When there's a scrollable area, can we use the fade mechanism rather than the drop shadow at the bottom, we feel that's a better visual indication?  If not, at short, wide desktop resolution let's move the drop shadow down towards the horizontal separator, so it's not on top of input fields - currently it could be perceived as a visual bug imo.

spadgett:
@serenamarie125 We can only use the ellipsis if we're OK making the URLs non-clickable and having it break mid-line after a hard-coded number of characters. See my comment https://github.com/openshift/origin-web-catalog/pull/209#issuecomment-299899676

rhamilto:
> @rhamilto agree with your suggestions as to when clip versus scroll!
But I'd like to suggest some visual suggestions after speaking with both Colleen & Kyle. When clipping, use an ellipses rather than fading out. Fading makes it feel like the content scrolls.

I've removed the fade out until we can determine the best approach for ellipsis-ing.  [Screenshots above](https://github.com/openshift/origin-web-catalog/pull/214#issue-226649136)

> When there's a scrollable area, can we use the fade mechanism rather than the drop shadow at the bottom, we feel that's a better visual indication? If not, at short, wide desktop resolution let's move the drop shadow down towards the horizontal separator, so it's not on top of input fields - currently it could be perceived as a visual bug imo.

So that we're intra-app consistent, I've stuck with the shadows, but moved them to the edges.   [Screenshots above](https://github.com/openshift/origin-web-catalog/pull/214#issue-226649136)

rhamilto:
Am thinking we should go ahead and merge and do the ellipsis-ing in a follow on...

rhamilto:
Opened https://github.com/openshift/origin-web-catalog/issues/220 to address the truncation indicator.

spadgett:
@jwforres @rhamilto When we pull this change into the console, we'll need to update `process-template-dialog`. It's not currently using `.wizard-pf-contents`, which causes layout problems. We should check the bind dialogs, too.

serenamarie125:
sounds good to me @spadgett

spadgett:
Thanks @serenamarie125

PR #216, opened 2017-05-09
Extend background image below service items

### LINKED ISSUE #198: Look of the OOTB empty SaaS offering section in IE11
Are we happy with the way the empty SaaS offering section looks?  If we are shipping both origin / OCP with nothing in that section (which is the current plan) then this is what everyone will see when they install openshift OOTB.

no-saas-services

cshinn:
Adds margin around services section and extends background image to bottom of screen.

Closes #198 

screen shot 2017-05-09 at 10 38 54 am


cshinn:
One question on this is whether we also want that margin above and below the services panel. I prefer it without, but I don't have a particularly strong opinion and am willing to update accordingly.

PR #219, opened 2017-05-11
Remember selected-project when going between steps of Ordering Panel

dtaylor113:
Hi,
This fixes https://bugzilla.redhat.com/show_bug.cgi?id=1449268
Also tested that the 'create-project' form fields are persisted between wizard steps as well.

### DIFF
--- a/src/components/select-project/select-project.controller.ts
+++ b/src/components/select-project/select-project.controller.ts
@@ -73,7 +73,9 @@ export class SelectProjectController implements angular.IController {
       this.ctrl.existingProjectNames = _.map(this.ctrl.projects, 'metadata.name');
 
       // get most recently created
-      this.ctrl.selectedProject = this.$filter('mostRecent')(this.ctrl.projects);
+      if (!this.ctrl.selectedProject) {
+        this.ctrl.selectedProject = this.$filter('mostRecent')(this.ctrl.projects);
+      }
 
       if (this.ctrl.canCreate) {
         this.ctrl.projects.unshift(createProject);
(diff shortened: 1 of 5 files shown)


src/services/catalog.service.ts
```diff
@@ -245,7 +245,7 @@ export class CatalogService {
     deferred.resolve([results, errorMsg]);
   }
 
-  private formatArray(arr){
+  private formatArray(arr: any) {
```
spadgett: If we are going to add a type, should it be `string[]` or at least `any[]`?

src/components/select-project/select-project.controller.ts
```diff
@@ -73,7 +73,9 @@ export class SelectProjectController implements angular.IController {
       this.ctrl.existingProjectNames = _.map(this.ctrl.projects, 'metadata.name');
 
       // get most recently created
-      this.ctrl.selectedProject = this.$filter('mostRecent')(this.ctrl.projects);
+      if (!this.ctrl.selectedProject) {
```
spadgett: Just confirming that this selects the correct project since the outside object reference won't match the one that comes back in the DataService.list call from inside this component.

src/services/catalog.service.ts
dtaylor113: Yes, fixed to 'string[]'

src/components/select-project/select-project.controller.ts
dtaylor113: `this.ctrl.selectedProject` is bindable on the `<select-project>` component.  Initially it is undefined.  I've tested this in my local environment and the selected project is initially set to the most recently created project, and once another (or the same) selected-project is set, it remains 'selected' between wizard steps.

PR #222, opened 2017-05-11
View All Link for Project Summary List

cdcabrera:
View All for project summary listing, always enabled


dtaylor113:
Screenshot would be really helpful :-)

src/components/projects-summary/projects-summary.html
```diff
@@ -30,7 +30,7 @@ <h4 class="catalog-modal-title">
     </a>
   </div>
   <div ng-if="$ctrl.projects && $ctrl.projects.length" class="catalog-project-summary-list">
-    <div ng-if="$ctrl.totalProjects > $ctrl.maxDisplayProjects" class="projects-count">
+    <div class="projects-count">
       <strong>{{$ctrl.maxDisplayProjects}}</strong>
       of
       <strong>{{$ctrl.totalProjects}}</strong>
```
spadgett: @beanh66 Should we be showing "3 of 3 Projects" in addition to the view all link?

beanh66: @spadgett @cdcabrera Yes the project count should be displayed as well. As long as there is at least 1 project, the project count and view all link should be shown.

spadgett: Thanks, should it say "3 of 3" or just "3"?

beanh66: "3 of 3"  so the user is clear that they may not see anything additional in the "View All" 👍

PR #223, opened 2017-05-11
Fixing bug where wizard steps indicator was too narrow in IE11

rhamilto:
Before:

screen shot 2017-05-11 at 4 40 55 pm

After:

screen shot 2017-05-11 at 4 28 31 pm

### DIFF
--- a/dist/less/overlay-panel.less
+++ b/dist/less/overlay-panel.less
@@ -80,6 +80,7 @@ body.overlay-open {
   }
 
   .wizard-pf-steps-indicator {
+    display: flex; // PatternFly doesn't set this for IE
     height: @wizard-pf-steps-indicator-height;
     margin: @wizard-pf-steps-indicator-margin-top-bottom 0;
     padding: 0;
--- a/dist/origin-web-catalogs.css
+++ b/dist/origin-web-catalogs.css
@@ -402,6 +402,7 @@ body.overlay-open .landing-side-bar {
   background-color: rgba(234, 234, 234, 0.5);
 }
 .catalogs-overlay-panel .wizard-pf-steps-indicator {
+  display: flex;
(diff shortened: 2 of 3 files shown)


serenamarie125:
@jeff-phillips-18 did you create a PF issue?

rhamilto:
> LGTM, we should file an issue to get this fixed in patternfly.

This was an [intentional implementation decision in PatternFly](https://github.com/patternfly/patternfly/blame/master/src/less/wizard.less#L115-L117).  Paging @priley86

priley86:
@rhamilto yes, I still prefer autoprefixer and think it saves us a ton more effort. 

in this case, I have opted to ease the stylint rules to ignore `"consecutive-duplicates-with-different-values"`. This allows us to still add `display: flex` after `display: inline-block` for ie11 (which does not support `@support` blocks like other browsers), and let autoprefixer take care of the prefixing for ie10/11. will this fix work for you?
https://github.com/patternfly/patternfly/pull/645

Out of curiosity, what minimum IE support does your app have these days? Microsoft has ended support for all IE versions except IE11.  It'd be nice to rid ourselves of these IE10 prefixes...and of course I can't wait for the day until we can just say "flex" and use BS4...
https://www.microsoft.com/en-us/windowsforbusiness/end-of-ie-support

rhamilto:
> in this case, I have opted to ease the stylint rules to ignore "consecutive-duplicates-with-different-values". This allows us to still add display: flex after display: inline-block for ie11 (which does not support @support blocks like other browsers), and let autoprefixer take care of the prefixing for ie10/11. will this fix work for you?
patternfly/patternfly#645

It will.  Thanks, @priley86.

> Out of curiosity, what minimum IE support does your app have these days? 

We only support 11.

PR #226, opened 2017-05-12
Truncates service names to 3 lines

### LINKED ISSUE #189: We should truncate long service title names
screen shot 2017-04-28 at 12 31 02 pm

screen shot 2017-04-28 at 12 33 55 pm

cc @rhamilto @sg00dwin @serenamarie125 @jeff-phillips-18

rhamilto:
Fixes #189 

Lots of reorganizing of less makes this PR looks more significant than it really is.

screen shot 2017-05-12 at 2 09 36 pm
screen shot 2017-05-12 at 2 09 48 pm


spadgett:
Are we using `word-break-all` because of flex? Would rather not break words in the middle if not necessary :/

rhamilto:
> Are we using word-break-all because of flex? Would rather not break words in the middle if not necessary :/

Yep.

spadgett:
If we remove `word-break-all` will it just truncate long unbroken words since we have `overflow: hidden`? I think we should prioritize the normal case and make sure words don't break in the middle if possible. The service class titles come from the broker, so they probably be reasonable.

rhamilto:
> If we remove word-break-all will it just truncate long unbroken words since we have overflow: hidden?

It will not.  Long, unbroken words will spill out of the dropdown or the panel column because of the flex-ed layouts.  As you noted, the flex-ed layouts are what is requiring the word-break-all (for Firefox and IE).  

So I think the options are:

1.  Keep what we've got.
2.  Remove the flex-ed layouts and see if we can get the desired results some other way.
3.  Don't apply a fix at all and let long, unbroken strings overflow.  
4.  Other?

Paging @sg00dwin

rhamilto:
> 4. Other?

Thanks to @sg00dwin, I've implemented the unknown-to-me fourth option, which is to put word-break() on the flexed child with a flex shorthand declaration of 1 1 0% (e.g., flex: 1 1 0%), which for some reason makes .word-break work on flexed children.  I'm going to [update .word-break-all()](https://github.com/openshift/origin-web-common/blob/master/src/styles/_mixins.less#L23-L29) accordingly. 

Please take another look, @spadgett

spadgett:
@rhamilto I'm curious if `min-width: 0` works in place of `flex: 1 1 0%`

rhamilto:
> @rhamilto I'm curious if min-width: 0 works in place of flex: 1 1 0%

It does not.  :(  IE11 requires the flex shorthand.

src/styles/order-service.less
```diff
@@ -1,6 +1,74 @@
...
+      font-weight: 600;
+      line-height: 1.4;
+      max-height: 4.2em; // limit to 3 lines
+      overflow: hidden;
+      @media (min-width: @order-service-title-sm-min) {
+        font-size: 22px;
+      }
+    }
+    .service-title-area {
+      flex: 1 1 0%;
```
spadgett: Probably worth a comment saying that this is needed for `word-break`

rhamilto:
Added comments about flex shorthand need for .word-break

spadgett:
@rhamilto thanks, OK to merge if travis passes

PR #228, opened 2017-05-16
Upgrade Patternfly versions to 3.24 and origin-common to 0.0.22

jeff-phillips-18:


package.json
```diff
@@ -30,16 +30,16 @@
     "angular-animate": "1.5.11",
     "angular-drag-and-drop-lists": "2.0.0",
     "angular-moment": "1.0.0",
-    "angular-patternfly": "3.20.0",
+    "angular-patternfly": "3.24.0",
     "angular-sanitize": "1.5.11",
     "angular-ui-bootstrap": "0.14.x",
     "angular-ui-router": "1.0.0-beta.1",
     "bootstrap": "^3.3.6",
     "font-awesome": "^4.6.1",
```
spadgett: Should we update to match what the resolution in bower.json?

jeff-phillips-18: Yes. Updated.

PR #230, opened 2017-05-16
Line-clamping service names and description, moving Learn more link

### LINKED ISSUE #220: Indicate truncation of long service names with a visual indicator
We either are or will be limiting service names to 3 lines of text [1]; however, we're providing no visual indicator that we're doing so.  We need to determine a cross-browser compatible solution and implement it.

[1] 
* Search results dropdown
* Browse
* Sidebar
* Panel overlay
   Note:  make sure $ctrl.serviceClass.resource.osbMetadata.documentationUrl is always visible! (move to top near name?)

rhamilto:
partially fixes #220

This maintains the existing behavior where long names are truncated to x number of lines, but it adds ellipses at the end of the truncated text for [browsers that support line-clamp](http://caniuse.com/#feat=css-line-clamp).
screen shot 2017-05-16 at 4 29 58 pm
screen shot 2017-05-16 at 4 36 27 pm
screen shot 2017-05-16 at 4 30 46 pm

Additionally, this moves the optional "Learn more" link for services up under tags so that the link doesn't get forced off the page when the description is truncated.
screen shot 2017-05-16 at 4 32 52 pm

Known issues:

* ellipses sometimes get clipped in Chrome
screen shot 2017-05-16 at 4 35 17 pm


src/styles/mixins.less
```diff
@@ -1,3 +1,13 @@
+.line-clamp(@lines: 1, @line-height: @line-height-base) {
```
spadgett: We should eventually move this to common.

rhamilto: I'll go ahead and do that now...

PR #231, opened 2017-05-16
Support service instance parameters

spadgett:
https://trello.com/c/qvua05Kh

Initial support for service instance parameters using angular-schema-form.

openshift catalog components 2017-05-16 18-20-42

@jwforres FYI

### DIFF
--- a/src/components/order-service/order-service.controller.ts
+++ b/src/components/order-service/order-service.controller.ts
@@ -48,6 +48,8 @@ export class OrderServiceController implements angular.IController {
     this.ctrl.longDescription = this.ctrl.serviceClass.longDescription;
     this.ctrl.plans = _.get(this, 'ctrl.serviceClass.resource.plans', []);
     this.ctrl.applications = [];
+    this.ctrl.parameterData = {};
+    this.ctrl.forms = {};
 
     // Preselect the first plan. If there's only one plan, skip the wizard step.
     this.ctrl.selectedPlan = _.first(this.ctrl.plans);
@@ -161,6 +163,8 @@ export class OrderServiceController implements angular.IController {
 
   public selectPlan(plan: any) {
     this.ctrl.selectedPlan = plan;
+    // Clear any previous parameter data since each plan has its own parameter schema.
+    this.ctrl.parameterData = {};
   }
 
   public provisionService = () => {
@@ -332,6 +336,18 @@ export class OrderServiceController implements angular.IController {
   private makeServiceInstance() {
     let serviceClassName = _.get(this, 'ctrl.serviceClass.resource.metadata.name');
 
+    // Omit parameters values that are the empty string. These are always
+    // optional parameters since you can't submit the form when it's missing
+    // required values. The template broker will not generate values for
+    // "generated" parameters that are there, even if the value is empty
+    // string, which breaks many templates.
+    //
+    // Check specifically for the empty string rather than truthiness so that
+    // we don't omit other values like `false` for boolean parameters.
+    let parameters: any = _.omit(this.ctrl.parameterData, (parameterValue) => {
+      return parameterValue === '';
+    });
+
     let serviceInstance = {
       kind: 'Instance',
       apiVersion: 'servicecatalog.k8s.io/v1alpha1',
@@ -341,7 +357,8 @@ export class OrderServiceController implements angular.IController {
        },
        spec: {
          serviceClassName: serviceClassName,
-         planName: this.ctrl.selectedPlan.name
+         planName: this.ctrl.selectedPlan.name,
+         parameters: parameters
        }
     };
 
--- a/src/decorators/bootstrap/array.html
+++ b/src/decorators/bootstrap/array.html
@@ -1,0 +1,32 @@
+<div  class="schema-form-array {{form.htmlClass}}"
+      sf-field-model="sf-new-array"
+      sf-new-array>
+  <label class="control-label" ng-show="showTitle()">{{ form.title }}</label>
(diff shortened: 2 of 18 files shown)


spadgett:
@jeff-phillips-18 It looks like all of the tests failed in travis and I'm not sure why. Any idea?

spadgett:
@jeff-phillips-18 Nevermind, I have the tests passing.

src/components/order-service/order-service.controller.ts
```diff
@@ -176,7 +179,8 @@ export class OrderServiceController implements angular.IController {
        },
        spec: {
          serviceClassName: serviceClassName,
-         planName: this.ctrl.selectedPlan.name
+         planName: this.ctrl.selectedPlan.name,
+         parameters: this.ctrl.parameterData
```
jwforres: depending on when this gets in might need to update this to the new parameter model format before sending

src/decorators/bootstrap/array.html
```diff
@@ -0,0 +1,32 @@
+<div  class="schema-form-array {{form.htmlClass}}"
```
jwforres: as far as naming these files, can it be src/decorators/schema-form-bootstrap/array.html ?
decorators/bootstrap is a little too generic to be clear what its for

jwforres: also how much did you have to change these files? it would be nice if we could keep track of how much we were having to change the OOTB decorators we copied these from.  Also should probably have a comment saying which version of the decorators we based it off of in case we need / want to update them.

spadgett:
@jwforres Here are the diffs:

array.html

```diff
13,14c13,14
13,14c13,14
<               style="position: relative; z-index: 20;"
<               type="button" class="close pull-right">
---
>               style="position: absolute; z-index: 20; right: 0; top: 12px; font-size: 20px;"
>               type="button" class="close">
29d28
<       <i class="glyphicon glyphicon-plus"></i>
```

checkbox.html

```diff
1c1,2
< <div class="checkbox schema-form-checkbox {{form.htmlClass}}"
---
> <div class="form-group">
>   <div class="checkbox schema-form-checkbox {{form.htmlClass}}"
3c4
<   <label class="{{form.labelHtmlClass}}">
---
>     <label class="{{form.checkboxLabelHtmlClass}}">
13c14,15
<   <div class="help-block" sf-message="form.description"></div>
---
>     <div class="help-block {{form.checkboxHelpHtmlClass}}" sf-message="form.description"></div>
>   </div>
---
>   <div class="help-block" sf-message="form.description"></div>
```

checkboxes.html

```diff
9a10
>   <div class="{{form.fieldWrapperHtmlClass}}">
20c21
<
---
>     </div>
```

default.html

```diff
3c3
<   <label class="control-label {{form.labelHtmlClass}}" ng-class="{'sr-only': !showTitle()}" for="{{form.key.slice(-1)[0]}}">{{form.title}}</label>
---
>   <label class="control-label {{form.labelHtmlClass}}" ng-class="{required: form.required, 'sr-only': !showTitle()}" for="{{form.key.slice(-1)[0]}}">{{form.title}}</label>
4a5
>   <div class="{{form.fieldWrapperHtmlClass}}">
14a16
>            ng-required="form.required"
51a54
>   </div>
```

select.html (to change to ui-select)

```diff
3c3
<   <label class="control-label {{form.labelHtmlClass}}" ng-show="showTitle()">
---
>    <label class="control-label {{form.labelHtmlClass}}" ng-class="{ required: form.required }" ng-show="showTitle()">
6c6,7
<   <select sf-field-model
---
>   <div class="{{form.fieldWrapperHtmlClass}}">
>     <ui-select sf-field-model
7a9
>             ng-required="form.required"
9,13c11,18
<           class="form-control {{form.fieldHtmlClass}}"
<           schema-validate="form"
<           ng-options="item.value as item.name group by item.group for item in form.titleMap"
<           name="{{form.key.slice(-1)[0]}}">
<   </select>
---
>             schema-validate="form">
>       <ui-select-match>
>         {{$select.selected.name}}
>       </ui-select-match>
>       <ui-select-choices repeat="item.value as item in form.titleMap | filter : $select.search">
>         <span ng-bind-html="item.name | highlightKeywords : $select.search"></span>
>       </ui-select-choices>
>     </ui-select>
14a20
>   </div>
```

src/decorators/bootstrap/array.html
spadgett: > as far as naming these files, can it be src/decorators/schema-form-bootstrap/array.html

I was trying to match the template path to make the files and originals easier to find (see index.ts), but I'm OK changing it too

```js
$templateCache.put('decorators/bootstrap/array.html', require('./decorators/bootstrap/array.html'));
```

spadgett:
I think eventually we'll need to write our own decorator for secrets and `valueFrom` (rather than just modifying the views from the bootstrap decorator), so this might be a point in time thing.

spadgett:
Rebased and disabled the horizontal form styles for now.

jwforres:
@spadgett looks like travis failed on dist, otherwise this is ready to merge once master re-opens

spadgett:
Updated the client to not send empty parameter values. See discussion in https://github.com/openshift/origin/issues/14445

@jim-minter FYI

spadgett:
Rebased. @jwforres PTAL, the changes to omit empty parameters are here:

https://github.com/openshift/origin-web-catalog/pull/231/files#diff-7a34cbbd85f9a19385ee34c6b893075fR339

jwforres:
LGTM, ready to merge it?

Issue #177, opened 2017-04-26
Overlay panel parameters


spadgett:
@serenamarie125 @jwforres @rhamilto @ncameronbritt 

I'm looking for input on service / template parameters.  This is a pretty typical template, and many templates have over 40 parameters. I'm concerned the two-panel dialog doesn't scale well for more complicated forms. Any ideas on how to improve the experience?

Help below each field:

![template-fields-inline-help](https://cloud.githubusercontent.com/assets/1167259/25443758/cef3dd12-2a76-11e7-9c81-71048e347276.gif)

Help tooltips:

![template-fields-info-icons](https://cloud.githubusercontent.com/assets/1167259/25444139/1e32b0fa-2a78-11e7-9c9b-bf83588ad367.gif)

dtaylor113:
Can they be separated into Required .vs Advanced Settings?  The latter we can hide in a dropdown

beanh66:
@spadgett Was there any thought on having the left content static? It seems like a waste of space to have nothing displayed as the user scrolls.

spadgett:
@beanh66 We can do that, although we'd need a separate scrollbar for the description if it's too long

beanh66:
@spadgett Could we have a max number of characters we allow? I'm worried double scroll may look odd.

ncameronbritt:
Being able to chunk up the form into proper wizard steps would be good too. I think that's been mentioned as a future goal. 

For now, is there a way to do it that where, if the user is going to accept the defaults ("simple" version), we only show the fields they would need to fill out? And if a user needs more advanced control, they could get to a full-page version of this form? Would the user who did a simple version then be able to go back and edit parameters?

spadgett:
Here's how it looks with this template if we only show required parameters. @jwforres had concerns hiding required parameters with default values.

![screencast 2017-04-26 15-03-34](https://cloud.githubusercontent.com/assets/1167259/25451873/a2255b24-2a91-11e7-8caf-eb230899de82.gif)

spadgett:
> Would the user who did a simple version then be able to go back and edit parameters?

No, and that's the concern. Some of the parameters might be difficult to change later.

serenamarie125:
The original intent was that the left side would not scroll, unfortunately I wasn't straight forward in the documentation about that, sorry!  

My opinion is to use Field Level Help for help text.  In the walkthroughs that I have had with customers, there is a clear preference for cleaner UIs with the field level help.  Can we make this change now?

[...]

PR #233, opened 2017-05-17
Add binding capabilities to the order flow. Use Angular PF wizard.

### LINKED ISSUE #232: Overlay close button too close to wizard step text on mobile
openshift catalog components 2017-05-17 07-25-12

@jeff-phillips-18 I suspect when we add a fourth step for bind, they won't fit anymore on mobile.

jeff-phillips-18:
Bumps origin-web-common version to v0.0.24

Fixes #232

### DIFF
--- a/dist/less/order-service.less
+++ b/dist/less/order-service.less
@@ -1,5 +1,7 @@
 // Breakpoint for showing a smaller icon and service title.
 @order-service-title-sm-min: 450px;
+@order-service-page-height: 410px;
+@order-service-wizard-min-external-height: 300px;
 
 .order-service-details {
   flex: 0 1 auto;
@@ -77,7 +79,7 @@
     margin-bottom: -(@grid-gutter-width / 2);
     margin-top: -(@grid-gutter-width / 2);
     overflow-y: auto;
-    padding-right: @grid-gutter-width;
+    padding-right: 0;
     .scroll-shadows-horizontal(65%, 0.25);
     width: 50%;
     _::-webkit-full-page-media, _:future, :root & { // only target Safari
@@ -138,6 +140,7 @@
     margin-right: -10px;
   }
   .footer-panel {
+    padding-top: 10px;
     text-align: center;
 
     a {
@@ -147,6 +150,9 @@
         text-decoration: none;
       }
     }
+  }
+  .sub-title {
+    margin: -5px 0 10px 25px;
   }
   .success-check {
     color: @color-pf-green-400;
@@ -172,17 +178,29 @@
       }
     }
   }
+  .review-message {
+    padding-left: 10px;
+  }
+  .review-status {
+    display: flex;
+    font-size: 16px;
+
+    .pficon {
+      margin-top: 3px;
+    }
+    .spinner.spinner-sm {
+      height: 16px;
+      margin: 3px 0 0;
+      width: 16px;
+    }
+  }
 }
 
-.wizard-pf-contents {
+.order-service-wizard-step.wizard-pf-main {
   @media (min-width: @screen-sm-min) {
-    display: flex;
-    min-height: 100%;
-    padding-bottom: (@grid-gutter-width / 2);
-    padding-top: (@grid-gutter-width / 2);
-    .scroll-shadows-horizontal-covers();
-    _::-webkit-full-page-media, _:future, :root & { // only target Safari
-      background: none;
-    }
+    height: @order-service-page-height + @grid-gutter-width;
+    max-height: calc(100vh ~"-" @order-service-wizard-min-external-height);
+    min-height: 200px;
+    overflow-y: auto;
   }
 }
--- a/dist/less/overlay-panel.less
+++ b/dist/less/overlay-panel.less
@@ -26,17 +26,18 @@ body.overlay-open {
   box-shadow: 0 3px 9px rgba(0,0,0,.5);
   margin: 0 auto;
   max-width: 1200px;
-  padding-bottom: @catalogs-overlay-panel-bottom-padding;
   position: relative;
 
   &.catalogs-overlay-panel-single-column {
     max-width: 800px;
   }
 
-  .wizard-pf-footer {
-    background: @color-pf-white;
-    padding-bottom: 15px;
-    padding-right: @grid-gutter-width;
+  .wizard-pf-contents {
+    @media (min-width: @screen-sm-min) {
+      display: flex; // only required for IE
+      flex: 1 1 auto;
+      width: 100%;
+    }
   }
 
   .wizard-pf-main {
@@ -53,6 +54,20 @@ body.overlay-open {
     }
   }
 
+  .wizard-pf-main-form-contents {
+    @media (min-width: @screen-sm-min) {
+      min-height: 100%;
+      padding: (@grid-gutter-width / 2) @grid-gutter-width (@grid-gutter-width / 2) (@grid-gutter-width / 2);
+      .form-horizontal .control-label {
+        padding-right: 0;
+      }
+      .scroll-shadows-horizontal-covers();
+      _::-webkit-full-page-media, _:future, :root & { // only target Safari
+        background: none;
+      }
+    }
+  }
+
   .wizard-pf-main-inner-shadow-covers {
     padding: (@grid-gutter-width / 2);
 
@@ -64,8 +79,6 @@ body.overlay-open {
     @media(min-width: @screen-sm-min) {
       display: flex;
       flex: 1 1 auto;
-      max-height: calc(100vh ~"-" (@navbar-pf-height + (@wizard-pf-steps-indicator-margin-top-bottom * 2) + @wizard-pf-steps-indicator-height + @wizard-pf-footer-height + @grid-gutter-width));
-      min-height: 300px;
       overflow: hidden;
       padding-right: 0;
     }
@@ -89,6 +102,9 @@ body.overlay-open {
     height: @wizard-pf-steps-indicator-height;
     margin: @wizard-pf-steps-indicator-margin-top-bottom 0;
     padding: 0;
+    @media(max-width: @screen-sm-min) {
+      margin-top: calc(@wizard-pf-steps-indicator-margin-top-bottom + 10px);
+    }
   }
 }
 
--- a/src/components/create-from-builder/create-from-builder.controller.ts
+++ b/src/components/create-from-builder/create-from-builder.controller.ts
@@ -3,7 +3,17 @@ import * as _ from 'lodash';
 import * as URI from 'urijs';
 
 export class CreateFromBuilderController implements angular.IController {
-  static $inject = ['$scope', '$filter', '$location', '$q', 'BuilderAppService', 'DataService', 'Logger'];
+  static $inject = [
+    '$scope',
+    '$filter',
+    '$location',
+    '$q',
+    'BuilderAppService',
+    'ProjectsService',
+    'DataService',
+    'BindingService',
+    'Logger',
+    'Constants'];
 
   public ctrl: any = this;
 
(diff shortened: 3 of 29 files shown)


dist/less/overlay-panel.less
```diff
@@ -89,6 +103,9 @@ body.overlay-open {
     height: @wizard-pf-steps-indicator-height;
     margin: @wizard-pf-steps-indicator-margin-top-bottom 0;
     padding: 0;
+    @media(max-width: @screen-sm-min) {
+      margin-top: calc(@wizard-pf-steps-indicator-margin-top-bottom + 10px);
```
rhamilto: Rather than adding more margin to accommodate the close button (which is why I assume you're doing this), how about we move the close closer to the edges of the modal, thus saving valuable vertical screen real estate?

```
.catalogs-overlay-panel-close {
  @media(max-width: @screen-sm-min) {
    right: 15px;
    top: 10px;
  }
}
```

looks like it works.

screen shot 2017-05-17 at 9 58 44 am

dist/less/overlay-panel.less
```diff
@@ -132,3 +149,7 @@ body.overlay-open {
     animation: modalSlideUp 0.2s ease-out forwards;
   }
 }
+
+.wizard-overlay .catalogs-overlay-panel {
+  padding-bottom: 0;
```
rhamilto: I don't think you need to override here.  I think you just need to remove padding-bottom on .catalog-overlay-panel?

dist/less/order-service.less
```diff
@@ -170,17 +175,24 @@
...
   @media (min-width: @screen-sm-min) {
-    display: flex;
-    min-height: 100%;
-    padding-bottom: (@grid-gutter-width / 2);
-    padding-top: (@grid-gutter-width / 2);
-    .scroll-shadows-horizontal-covers();
-    _::-webkit-full-page-media, _:future, :root & { // only target Safari
-      background: none;
-    }
+    height: calc(@order-service-page-height + @grid-gutter-width);
```
spadgett: I know we've had some debate on this and I understand why height/max-height are added, but I agree with @rhamilto. As a user, only being able to use a small portion of my screen to fill out what could be a complicated form is frustrating. We don't control what parameters come back from the brokers, and the form will get longer when we add parameters in #231. I feel like the minor inconvenience of the dialog resizing and the buttons moving is the lesser evil, particularly when our wizard only has a few steps.

spadgett: We can also add a reasonable min-height at non-mobile screen sizes so that it's effectively a fixed height unless there's a lot of content causing it to grow.

jeff-phillips-18: @serenamarie125 @beanh66 @cshinn Thoughts?

serenamarie125: From an experience perspective, having the wizard change sizes will be unexpected as well as change the location of the buttons.  Based on my understanding of users and use cases, and applying the 80/20 rule, I would suggest that the behavior of the wizard be that the page size is consistent, and we use scrolling as needed.

Whether that means we need to increase the minimum height to make sure that our "typical" number of parameters is shown without a scroll is a different question.  

@jeff-phillips-18 @rhamilto @spadgett

spadgett: > Based on my understanding of users and use cases, and applying the 80/20 rule, I would suggest that the behavior of the wizard be that the page size is consistent, and we use scrolling as needed.

Most of our templates have dozens of parameters. If we're applying the 80/20 rule, the common case is to have a lot, at least for the template broker. It's hard to say what other brokers will do.

> Whether that means we need to increase the minimum height to make sure that our "typical" number of parameters is shown without a scroll is a different question.

Even our full screen template page has a scrollbar. I don't think there is a minimum height that avoids it.

As someone who uses this dialog every day: having it resize doesn't bother me, but not being able to use the screen space makes it hard to use (see #177). I can only speak for myself, though.

serenamarie125: @spadgett dozens of parameters are within the advanced options, or are regular options?  I thought that you and Jessica asked us to get customer feedback on how many of these options are actually specified upfront, implying that we think we are offering too many parameters.  Did i misunderstand that?

From a usability perspective, i still disagree and think that the wizard should remain a consistent size.  @cshinn @beanh66 thoughts?

beanh66: @serenamarie125 Personally I would rather deal with scrolling than deal with the modal size changing and buttons moving around. One main advantage of a wizard is the consistent layout. I agree we do want to show as much as possible, so changing the minimum height makes sense, but as @spadgett mentioned, users often have to scroll regardless.

jwforres:
On Wed, May 17, 2017 at 2:24 PM, Serena Doyle <notifications@github.com>
wrote:

> *@serenamarie125* commented on this pull request.
> ------------------------------
>
> In dist/less/order-service.less
> <https://github.com/openshift/origin-web-catalog/pull/233#discussion_r117076082>
> :
>
> >    @media (min-width: @screen-sm-min) {
> -    display: flex;
> -    min-height: 100%;
> -    padding-bottom: (@grid-gutter-width / 2);
> -    padding-top: (@grid-gutter-width / 2);
> -    .scroll-shadows-horizontal-covers();
> -    _::-webkit-full-page-media, _:future, :root & { // only target Safari
> -      background: none;
> -    }
> +    height: calc(@order-service-page-height + @grid-gutter-width);
>
> @spadgett <https://github.com/spadgett> dozens of parameters are within
> the advanced options, or are regular options? I thought that you and
> Jessica asked us to get customer feedback on how many of these options are
> actually specified upfront, implying that we think we are offering too many
> parameters. Did i misunderstand that?
>

We were asking about creating a new app from a source builder image, that's
a different question because we have total control over what the form
contains.

Templates are like any other provisioned services in that we have zero
control over what people install into their cluster. Yes most of our OOTB
origin templates can be created with no additional input from the user.
But I can't say if the same is true for all of the JBoss and middleware
templates that ship with OCP, would need to check.  And customers can do
anything.  The parameter schema is defined on the service, we auto-generate
the UI form off of that.

dist/less/order-service.less
spadgett: > dozens of parameters are within the advanced options, or are regular options? I thought that you and Jessica asked us to get customer feedback on how many of these options are actually specified upfront, implying that we think we are offering too many parameters. Did i misunderstand that?

There is no advanced options page for service instance parameters (yet anyway). For context, there are three different forms we show in the overlay from the landing page catalog currently:

1. Builder image forms. We have control over what's shown here. This is the from from #27 that we wanted customer feedback on.
2. Template forms. We **don't** control what the parameters are.
3. Service instance forms (provisioning services from the catalog). We **don't** control what the parameters are.

For (1) and (2), we've reused existing pages in the web console to add the advanced options link. We have not done that for (3), and we don't have an existing page to reuse.

(2) will go away after tech preview when we enable the template broker, so the templates will then show up like other service instances (3) and we'll have to revisit how we deal with parameters.

In short, we've (sort of) solved it for 2 of the 3 cases, but I'm not sure we have a good solution for service instance parameters. If we need to create a standalone provision service page for advanced options, we should open a Trello card because that's another chunk of work that isn't being tracked.

src/filters/Applications.ts
```diff
@@ -0,0 +1,13 @@
+export function applicationHasDeploymentFilter ($filter: any) {
```
jwforres: lets not recreate these filters in catalog with slightly different names, I don't want this in multiple repos, its going to be a pain to track down if it ever changes. Move the filters up to common if needed.

dist/less/order-service.less
jeff-phillips-18: I'm really torn on this. Options:

1. Make the wizard resize per page

  This jumps the wizard height around even when selecting options on a single page. This is a bit jarring to me and the next/back buttons move up and down while trying to navigate quickly.

2. Keep a consistent height as it is now

This means more scrolling on pages with lots of parameters. Worse when there are a moderate amount and I have to scroll for one or two parameters when I wouldn't otherwise have to scroll at all. When there are a slew of them, I'm scrolling either way.

3. Keep a consistent height, but use as much window vertical space as possible

Less scrolling for sure, can probably fit an extra 8-10 parameters before needing to scroll. When simply selecting an existing project, the page is largely empty and lots of mouse movement to get down to the next button.

dist/less/overlay-panel.less
jeff-phillips-18: But if there are more steps then eventually it runs over. Will happen in non-mobile as well if there are lots of steps so I was leaning more towards always having the extra margin so that it would never overlap.

dist/less/overlay-panel.less
jeff-phillips-18: I've assumed that would still be wanted by other uses (non-wizard). I'm OK with this if we feel it is OK to change the generic panel and accommodate the change in other uses if necessary.

src/components/create-from-builder/create-from-builder.controller.ts
```diff
@@ -233,4 +224,192 @@ export class CreateFromBuilderController implements angular.IController {
     let name = this.ctrl.imageStream.resource.metadata.name + ":" + this.ctrl.istag.name;
     return this.DataService.get("imagestreamtags", name, { namespace: 'openshift' });
   }
+
+  private preselectService() {
```
jeff-phillips-18: Agreed, however the two were different enough that I did not create a binding service. I can add an issue to create a service in common to handle this.

dist/less/overlay-panel.less
rhamilto: IMO, the steps area should be allowed to expand if the steps begin to stack.  No amount of extra margin is going to accommodate and indeterminate number of steps.

src/components/order-service/order-service.controller.ts
```diff
@@ -28,80 +40,114 @@ export class OrderServiceController implements angular.IController {
...
+    this.ctrl.nextTitle = 'Create';
+    this.ctrl.steps[3].allowed = this.ctrl.steps[2].valid;
 
-  public previousStep() {
-    let step = this.ctrl.steps[this.currentStepIndex - 1];
-    this.gotoStep(step);
-  }
+    this.validityWatcher = this.$scope.$watch("$ctrl.forms.bindForm.$valid", (isValid: any, lastValue: any) => {
+      this.ctrl.steps[2].valid = isValid;
+      this.ctrl.steps[3].allowed = this.ctrl.steps[2].valid;
```
jeff-phillips-18: Agreed. The indexing got used more and more as I continued to develop and I didn't look back over it. Makes good sense, I will update.

src/components/create-from-builder/create-from-builder.controller.ts
```diff
@@ -233,4 +224,192 @@ export class CreateFromBuilderController implements angular.IController {
...
+          return  rightCreated.localeCompare(leftCreated);
+        } else {
+          return leftReady ? -1 : 1;
+        }
+      });
+      this.ctrl.serviceInstances = instances;
+    }
+  };
+
+  private updateSteps() {
```
jeff-phillips-18: We set the title when we enter the page, updateSteps is slightly misnamed as it only occurs when the project selection changes. I'll try to simplify.

src/components/order-service/order-service.controller.ts
```diff
@@ -199,6 +364,9 @@ export class OrderServiceController implements angular.IController {
 
       this.ctrl.orderComplete = readyCondition && readyCondition.status === 'True';
       this.ctrl.error = _.find(conditions, {type: "ProvisionFailed"});
+      if (this.ctrl.shouldBindToApp !== 'none' && this.ctrl.orderComplete && !this.ctrl.error) {
```
jeff-phillips-18: OK, great. I didn't know the binding could happen in parallel.

jwforres:
Yep the service catalog backend will wait until the provision completes
before starting the bind operation, the binding object will just stay in a
"not ready" state until then.

jeff-phillips-18:
Marked as WIP since this will require an update to origin-common to include https://github.com/openshift/origin-web-common/pull/73

jeff-phillips-18:
two remaining issues:

wizard height
Firefox stupid long name messes up two column layout.

spadgett:
I won't block the PR on wizard height.

> Firefox stupid long name messes up two column layout.

@rhamilto I know you had this working. Do you know what change might have caused the problem?

jeff-phillips-18:
I figured it out. I had to set wizard-pf-contents width=100%

jeff-phillips-18:
I think adding another 100px to the wizard height is a reasonable compromise for now. Once we see it with lots of parameters it can be easily adjusted if desired.

jeff-phillips-18:
I believe all comments are addressed now. Please take another peek.

Just waiting on the updated version of origin-common now.

jeff-phillips-18:
Removed WIP, bumped origin-web-common to v0.0.24

dist/less/order-service.less
rhamilto: If the decision is to go with a fixed height, there is no need to additionally specify a max-height and a min-height.

jeff-phillips-18: The min-height prevents the wizard from getting too small if the user resizes the window to be very short. The max-height is the standard height. So, its not *really fixed height.

rhamilto: No need to use calc on height since this is within less?

jeff-phillips-18: removed the calc. thanks.

jeff-phillips-18:
@jwforres @rhamilto Any further issues/concerns?

src/components/order-service/order-service.controller.ts
```diff
@@ -164,6 +249,75 @@ export class OrderServiceController implements angular.IController {
     }
   }
 
+  private updateBindability() {
+    this.bindStep.hidden = _.size(this.ctrl.applications) < 1;
```
jwforres: Should probably fix this in a follow up PR but there is an existing issue https://github.com/openshift/origin-web-console/issues/1470 around binding even if you have no applications, the bind dialog is incorrect today (it actually shows up always in the overview but gives you the option to pick apps when there are none).  We should always let you create the generic secret binding, even if you don't have applications.

src/components/order-service/order-service.controller.ts
```diff
@@ -183,22 +337,21 @@ export class OrderServiceController implements angular.IController {
...
 
   private watchResults = (resource: any, data: any, context: any) => {
     this.watches.push(this.DataService.watchObject(resource, data.metadata.name, context, (instanceData: any, action: any) => {
       var conditions: any = _.get(instanceData, 'status.conditions');
       var readyCondition: any = _.find(conditions, {type: "Ready"});
 
+      this.ctrl.serviceInstanceName = _.get(instanceData, 'metadata.name');
       this.ctrl.orderComplete = readyCondition && readyCondition.status === 'True';
       this.ctrl.error = _.find(conditions, {type: "ProvisionFailed"});
+      if (this.ctrl.shouldBindToApp !== 'none' && !this.ctrl.bindInProgress) {
```
jwforres: I don't think it makes sense to call this from here, its not something you will do more than once, I would just move this to before your call to watchResults.  As long as the creation of the service instance object was successful, which you know at the point where you call watchResults, then you can create the binding.

jeff-phillips-18: before the first callback to watch results, we do not know the name of the service instance which is needed to call bindService. I added the check for inProgress to prevent any additional calls to bindService.

src/components/order-service/order-service.controller.ts
jeff-phillips-18: OK, I can fix this as part of checking if a service is bindable.

src/components/order-service/order-service.controller.ts
jwforres: You should have the data you need. The success handler to DataService.create gives you the result of the creation (the data arg), which is the updated serviceInstance object that contains the name

jwforres: sorry might be the resource arg not the data arg, one of those is the response object

jeff-phillips-18: super cool, updated. Thanks!

jwforres: I'm not seeing the change, did you push it?

PR #235, opened 2017-05-17
Optimize Catalog Item Retrieval

### LINKED ISSUE #215: Optimize Categorization of Items
Currently, each catalog item receives a map of categories and sub-categories it matches, which is then used for filtering when categories and sub-categories are clicked upon in the Catalog Service View.

Need to investigate whether it would be cleaner to simply add items to a global category/sub-category object.  Similar to console's 'countByCategory' and/or 'buildersByCategory'

dtaylor113:
Previously, each catalog item kept track of it's own associated categories & sub-categories.  This PR removes categories/sub-categories from catalog items and instead  categorizes each item into an items array under each sub-category in the main Catalog object (similar to console's `buildersByCategory` object).   Categorization also dynamically adds 'All' and 'Other' main and sub-categories when needed.

This allows quicker processing as we only have to cycle through the category items once to categorize them whereas previously each catalog item ran through each category/sub-category, and then a second pass to remove those categories/sub-categories which weren't associated with any items.

This also allows for cleaner filtering code in the `<service-view>` component.

All unit tests passed include service-view.spec which tests many of the categorization rules.

Fixes #215

### DIFF
--- a/src/components/services-view/services-view.controller.ts
+++ b/src/components/services-view/services-view.controller.ts
@@ -3,19 +3,21 @@ import * as _ from 'lodash';
 import * as $ from 'jquery';
 
 export class ServicesViewController implements angular.IController {
-  static $inject = ['Constants', 'Catalog', '$filter', '$scope', '$timeout'];
+  static $inject = ['Constants', 'Catalog', 'Logger', '$filter', '$scope', '$timeout'];
 
   public ctrl: any = this;
   private constants: any;
   private catalog: any;
+  private logger: any;
   private $filter: any;
   private $scope: any;
   private $timeout: any;
   private debounceResize: any;
 
-  constructor(constants: any, catalog: any, $filter: any, $scope: any, $timeout: any) {
+  constructor(constants: any, catalog: any, logger: any, $filter: any, $scope: any, $timeout: any) {
     this.constants = constants;
     this.catalog = catalog;
+    this.logger = logger;
     this.$filter = $filter;
     this.$scope = $scope;
     this.$timeout = $timeout;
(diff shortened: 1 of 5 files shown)


src/services/catalog.service.ts
```diff
@@ -113,85 +112,101 @@ export class CatalogService {
...
 
-  public hasCategory(item: any, category: string) {
-    let found: boolean = false;
+  private getAllOrOtherMainCategory(category: string) {
 
-    _.each(item.catsBySubCats, (cats: any) => {
-      found = _.includes(cats, category);
-      return !found;  // to break out of _.each when found
-    });
+    let categoryObj: any = _.find(this.categories, {id: category});
```
jwforres: this find is going to be run for every single item in the catalog since every item goes into the top level all, not terrible at current scales but we need to be efficient about handling thousands of items potentially. It would be better to create private variables that point to the top level "all" and "other" categories.  Also why not just always add all and other? You already have a concept of hiding/showing categories depending on whether hasItems is set.

src/components/services-view/services-view.controller.ts
```diff
@@ -81,8 +80,11 @@ export class ServicesViewController implements angular.IController {
         subCats = subCats.concat(categoryObj.subCategories);
       }
     });
-    if (subCats.length > 1) {
-      subCats.unshift({id: 'all', label:  'All'});
+    subCats = _.filter(subCats, {hasItems: true});
+    // if 'all' and one other sub-cat, remove the 'all' and just
+    // show the single subCat
+    if (subCats[0].id === 'all' && subCats.length === 2) {
```
dtaylor113: It should always be 'all' but doesn't hurt to double-check :-)

src/services/catalog.service.ts
dtaylor113: Always creating top level 'all' and 'other' main categories; and using private variables that point to them.
- Thanks

dtaylor113:
Hi @jwforres, I believe I have addressed your last review issues. -thanks!

PR #236, opened 2017-05-17
Show one row of SasS offerings, add Show More/Less

jeff-phillips-18:
image

image


dist/less/saas-list.less
```diff
@@ -26,10 +32,12 @@
       @media (min-width: @screen-sm-min) {
         border-bottom: none;
         flex: 0 1 auto;
```
rhamilto: I believe if you set flex-grow to 1 (flex: 1 1 auto), you don't need the min-widths (here and line 40).

src/components/saas-list/saas-list.controller.ts
```diff
@@ -2,15 +2,52 @@ import * as angular from 'angular';
...
+    var windowWidth: number = this.$window.innerWidth;
+    var offeringCount: number = _.size(this.ctrl.saasOfferings);
+
+    this.ctrl.itemsOverflow = (windowWidth >= this.$window.patternfly.pfBreakpoints.tablet) &&
+      ((offeringCount > 4) || (windowWidth < this.$window.patternfly.pfBreakpoints.desktop && offeringCount > 2));
+  }
+
+  private onWindowResize = () => {
+    this.updateListExpandVisibility();
+    this.$scope.$digest();
```
spadgett: Suggest `$scope.$evalAsync` rather than calling `$digest` directly.

src/components/saas-list/saas-list.controller.ts
```diff
@@ -2,15 +2,52 @@ import * as angular from 'angular';
...
+  constructor($scope: any, $window: any, $element: any) {
+    this.$scope = $scope;
+    this.$window = $window;
+    this.$element = $element;
+    this.ctrl.sassListExpanded = false;
+    this.ctrl.itemsOverflow = false;
+  }
+
+  public $postLink() {
+    angular.element(this.$window).bind('resize', this.onWindowResize);
```
spadgett: Can we debounce the listener?

dist/less/saas-list.less
jeff-phillips-18: If i take out the min-width, the item shrink and fit so it doesn't limit the number per row correctly.

rhamilto: Then set flex shrink to 0 (flex: 1 1 auto).
screen shot 2017-05-17 at 4 53 42 pm

Point is, you shouldn't need to set both width and min-width with flex (assuming you set flex-grow and flex-shrink properly)

jeff-phillips-18: Gotcha.

src/components/saas-list/saas-list.controller.ts
jeff-phillips-18: Done. Thanks.

src/components/saas-list/saas-list.controller.ts
jeff-phillips-18: Done. Thanks.

jeff-phillips-18:
Fixed issue where a horizontal scrollbar appears on the landing page.

jeff-phillips-18:
@spadgett any further thoughts/concerns?

src/components/saas-list/saas-list.controller.ts
```diff
@@ -2,15 +2,58 @@ import * as angular from 'angular';
...
+    var windowWidth: number = this.$window.innerWidth;
+    var offeringCount: number = _.size(this.ctrl.saasOfferings);
+
+    this.ctrl.itemsOverflow = (windowWidth >= this.$window.patternfly.pfBreakpoints.tablet) &&
+      ((offeringCount > 4) || (windowWidth < this.$window.patternfly.pfBreakpoints.desktop && offeringCount > 2));
+  }
+
+  private onWindowResize = () => {
+    this.updateListExpandVisibility();
+    this.$scope.$evalAsync();
```
spadgett: Should call `updateListExpandVisibility` inside the `$evalAsync` callback.

```js
this.$scope.$evalAsync(() => {
  this.updateListExpandVisibility();
});
```

jeff-phillips-18: ah yes...

PR #247, opened 2017-05-24
Hide the back button in the builder dialog

spadgett:


jeff-phillips-18:
There can be a back button when binding. Why are we removing it?

spadgett:
> There can be a back button when binding.

Sorry I didn't realize you had added that to the builder dialog.

PR #248, opened 2017-05-24
Update showcase app, remove obsolete navigation pages

jeff-phillips-18:
This gives a more console-like experience when demo’ing the showcase app.


jeff-phillips-18:
@jwforres Any objections to this?  Only effects the the app not the dist. Would help with designers demoing in dev mode.

PR #252, opened 2017-05-25
[WIP] Catalog Item Filtering, View All in Search

dtaylor113:
Need to rebase to #261 after it is merged.

image

image

[Additional Screenshots](https://docs.google.com/a/redhat.com/document/d/1_OICX2TRhgL8iNalpK-nWi87YQmCJTulAWTE23qTuTE/edit?usp=sharing)

- Clicking 'View All' from Search again, clears any existing keyword filters
- Additional keyword filters are AND'ed together
- Switching to new Category/Sub-Category clears any applied keyword filters
- Filtering input disabled if (originally) only one catalog item
- When no applied keyword filters, filter results label says: 'n Items'.  Tried to hide when no filters, but causes flickering -need to fix in angular-patternfly.


serenamarie125:
@dtaylor113 

1. Please change "Active filters:" to "Active Filters:" using a capital F.
2. When active filters wrap to a second row, I would suggest having 10 px margin above and 20 px below @cshinn thoughts here ?

dtaylor113:
> Please change "Active filters:" to "Active Filters:" using a capital F.

Will do

> When active filters wrap to a second row, I would suggest having 10 px margin above and 20 px below @cshinn thoughts here ?

It gets a little tricky because we are dynamically calculating the height of the 'expansion card' which holds all the items.  Putting margins for sm-screens but not others causes some issues.

dtaylor113:
> 1. Please change "Active filters:" to "Active Filters:" using a capital F.

Fixed: 
image

dtaylor113:
> When active filters wrap to a second row, I would suggest having 10 px margin above and 20 px below @cshinn thoughts here ?

@serenamarie125 , @cshinn Added 10px top margin and 20px bottom margin/padding:

image

src/components/services-view/services-view.controller.ts
```diff
@@ -93,14 +128,55 @@ export class ServicesViewController implements angular.IController {
...
+    this.$timeout(() => {
+      if (this.ctrl.filterConfig.appliedFilters.length > 0) {
+        this.ctrl.filterConfig.resultsCount = this.ctrl.filteredItems.length;
+        $('.toolbar-pf-results h5').text(this.ctrl.filterConfig.resultsCount + ' of ' + this.ctrl.totalCount + ' items');
+        $('.toolbar-pf-results p').first().text('Active Filters:');
+      } else {
+        $('.toolbar-pf-results h5').text(this.ctrl.totalCount + ' items');
+
+        if (this.ctrl.totalCount <= 1) {
+          $('.filter-pf.filter-fields input').attr('disabled', '');
```
jeff-phillips-18: This should be available via Angular Patternfly as well if this is a desired behavior.

src/styles/services-view.less
```diff
@@ -116,11 +126,23 @@
...
       position: absolute;
       right: 0;
+
+      .services-items {
+        left: 0;
+        position: absolute;
+        right: 0;
+        margin-top: 0px;
+        padding-top: 24px;
+
```
jeff-phillips-18: nit: alphabetization

src/styles/services-view.less
```diff
@@ -234,6 +256,33 @@
...
+      }
+    }
+
+    .toolbar-pf-results {
+      border-top: none;
+      position: absolute;
+      top: 46px;
+      left: 24px;
+      margin-top: 10px;
+      z-index: 1;
```
jeff-phillips-18: nit: alphabetization

src/styles/services-view.less
```diff
@@ -234,6 +256,33 @@
...
+
+      @media (min-width: @screen-md-min) {
+        margin-top: -26px;
+        top: 37px;
+        left: 312px;
+      }
```
jeff-phillips-18: nit: alphabetization

src/components/services-view/services-view.controller.ts
```diff
@@ -93,14 +128,55 @@ export class ServicesViewController implements angular.IController {
...
   private updateActiveCardStyles() {
     this.$timeout(this.resizeExpansion, 50);
   }
+
+  private updateFilterControls() {
+    this.$timeout(() => {
+      if (this.ctrl.filterConfig.appliedFilters.length > 0) {
+        this.ctrl.filterConfig.resultsCount = this.ctrl.filteredItems.length;
+        $('.toolbar-pf-results h5').text(this.ctrl.filterConfig.resultsCount + ' of ' + this.ctrl.totalCount + ' items');
+        $('.toolbar-pf-results p').first().text('Active Filters:');
```
dtaylor113: Hi, filed the following issue in angular-patternfly repo.:  [pfFilter Enhancements](https://github.com/patternfly/angular-patternfly/issues/472).

src/components/services-view/services-view.controller.ts
dtaylor113: Yes, it is included in (pfFilter Enhancements)[https://github.com/patternfly/angular-patternfly/issues/472)

src/styles/services-view.less
dtaylor113: Fixed

src/styles/services-view.less
dtaylor113: Fixed

src/styles/services-view.less
dtaylor113: Fixed

dtaylor113:
Yay!  Thanks @jeff-phillips-18 for finding and fixing Travis!
@spadgett , @jwforres This PR is once again ready to be reviewed -thanks

cshinn:
@dtaylor113 looks and works great!

spadgett:
@rhamilto @sg00dwin FYI, might conflict with the work you're doing on the services view

spadgett:
I find it odd that there are two search inputs so close when I first open the page, and they both work differently. @openshift/team-ux-review Is this the intented behavior?

openshift catalog components 2017-06-01 16-19-16

beanh66:
@spadgett Yes, this is our intended behavior. With Search, a user can choose to view all results, which will clear current filters and automatically create a single keyword filter on the "All" tab regardless of where they were. The Filter in context is specific to the category or subcategory the user is viewing. When already on the "All" tab, it does function similar to search, but once a user is inside a category, the Filter is different than search. Users can create additional keyword filters from within a particular category, which cannot be done from search. This distinction will also become more clear once additional filter mechanisms are added.

spadgett:
@beanh66 My worry is that the differences aren't obvious looking at the UI since they're both just input boxes. I think even saying "Filter in Category" in the placeholder would help make the difference more clear.

I also suggest we add an `Add` button with the filter input instead of just relying on the enter key. It gives an additional cue that this input is different and is better for accessibility.

beanh66:
@spadgett We discussed this in our daily meeting, but just recording here too so we have it documented. The consensus was to change the label in the second filter box to read "Filter by keyword in Category" and for now we are going to do without the "Add" button, and will plan on testing this for possible future changes.

dtaylor113:
Hi @rhamilto & @sg00dwin, 
I've rebased this PR and I've gotten all functionality to work.   The filter controls display correctly under 'All' and 'Other' main categories.  However, I'm having trouble setting up the less for when the filtering is under a sub-category.  -I was hoping @rhamilto or @sg00dwin might have some time to take a look?   
I'll continue to work on it, but the Less is super complex to me, especially all the mobile modes.  -Thanks

rhamilto:
@dtaylor113, if you still need help, I'm free.  @spadgett is going to pick up the pod preset stuff.

src/components/services-view/services-view.controller.ts
spadgett: Are you able to use CSS `text-transform: capitalize` instead?

dtaylor113:
Thanks @rhamilto !   Let me push my lastest Less, it's another approach where I play with the filter control's flex bias to take up entire first row in the wrap; instead of trying to put in a container <div>.

dtaylor113:
Hi, pushed some more Less adjustments, looking pretty good at mobile screen sizes:

https://docs.google.com/a/redhat.com/document/d/1_OICX2TRhgL8iNalpK-nWi87YQmCJTulAWTE23qTuTE/edit?usp=sharing

I agree with @rhamilto that any additional adjustments, if needed, can be done via bug fixes.

src/components/services-view/services-view.controller.ts
dtaylor113: Hi @spadgett , fixed to use ```text-transform: capitalize```

src/components/services-view/services-view.controller.ts
```diff
@@ -3,35 +3,61 @@ import * as _ from 'lodash';
...
+    this.ctrl.filterConfig = {};
   }
 
   public $onInit() {
     this.debounceResize = _.debounce(this.resizeExpansion, 50, { maxWait: 250 });
     angular.element(window).bind('resize', this.debounceResize);
     $(window).on('resize.services', this.debounceResize);
+
+    this.$rootScope.$on('filter-catalog-items', (event: any, searchCriteria: any) => {
+      let keyWordFilter: any = angular.copy(this.ctrl.filterConfig.fields[0]);
```
jwforres: i would prefer to have named references vs array indexing for things like this, create a named reference to the field, and then when you setup ctrl.filterConfig fields just include the field via the ref there, and then use the ref here as well

dtaylor113: Hi, fixed.

src/components/services-view/services-view.controller.ts
```diff
@@ -3,35 +3,60 @@ import * as _ from 'lodash';
...
   private htmlService: any;
   private $filter: any;
+  private $rootScope: any;
   private $scope: any;
   private $timeout: any;
   private debounceResize: any;
+  private keyWordFilterFld: any = {
+    id: 'keyword',
+    title:  'Keyword',
+    placeholder: 'Filter by keyword in Category',
```
spadgett: Casing looks odd to me, is this what you meant @beanh66 ?

The placeholder in the search is `Search Catalog` so should we just always use headline case?

src/components/services-view/services-view.controller.ts
```diff
@@ -3,35 +3,60 @@ import * as _ from 'lodash';
...
     this.ctrl.mobileView = 'categories';
+    this.ctrl.filterConfig = {};
   }
 
   public $onInit() {
     this.debounceResize = _.debounce(this.resizeExpansion, 50, { maxWait: 250 });
     angular.element(window).bind('resize', this.debounceResize);
     $(window).on('resize.services', this.debounceResize);
+
+    this.$rootScope.$on('filter-catalog-items', (event: any, searchCriteria: any) => {
```
spadgett: This listener needs to be removed on destroy. Otherwise we leak memory and it will keep getting called even after we leave the page.

src/components/services-view/services-view.controller.ts
```diff
@@ -3,35 +3,60 @@ import * as _ from 'lodash';
...
   private catalog: any;
+  private keywordService: any;
   private logger: any;
   private htmlService: any;
   private $filter: any;
+  private $rootScope: any;
   private $scope: any;
   private $timeout: any;
   private debounceResize: any;
+  private keyWordFilterFld: any = {
```
spadgett: Inconsistent casing with `keywordService` above? Can we call it `keywordFilterField`?

src/components/services-view/services-view.controller.ts
```diff
@@ -131,4 +183,20 @@ export class ServicesViewController implements angular.IController {
...
+        $('.toolbar-pf-results h5').text(this.ctrl.filterConfig.resultsCount + ' of ' + this.ctrl.totalCount + ' items');
+      } else {
+        $('.toolbar-pf-results h5').text(this.ctrl.totalCount + ' items');
+        if (this.ctrl.totalCount <= 1) {
+          $('.filter-pf.filter-fields input').attr('disabled', '');
+        } else {
+          $('.filter-pf.filter-fields input').removeAttr("disabled");
+        }
+      }
+    }, 50);
```
spadgett: Does this work with `0` instead of `50`? I think as long as you use `$timeout` it will run after the next digest loop, so the elements should be there.

What does it look like before the text replacements?

src/components/services-view/services-view.html
```diff
@@ -31,24 +31,26 @@
...
-              <span class="{{item.iconClass}}"></span>
-            </div>
-            <div ng-if="item.imageUrl" class="services-item-icon">
-              <img ng-src="{{item.imageUrl}}" alt="">
-            </div>
-            <div class="services-item-name" title="{{item.name}}">
-              {{item.name}}
-            </div>
-          </a>
+            <div pf-filter config="$ctrl.filterConfig"></div>
```
spadgett: Identation

src/styles/services-view.less
```diff
@@ -227,6 +241,42 @@ services-view,
...
+      .toolbar-pf-results {
+        border-top: none;
+        left: -5px;
+        margin-top: 20px;
+        position: absolute;
+        top: 106px;
+        z-index: 1;
+
+        p {
+          text-transform: capitalize;
```
spadgett: Better, thanks :)

src/components/catalog-search/catalog-search.controller.ts
```diff
@@ -60,6 +68,12 @@ export class CatalogSearchController implements angular.IController {
   private filterForKeywords(searchText: string) {
     let keywords = this.KeywordService.generateKeywords(searchText);
     let items = this.KeywordService.filterForKeywords(this.ctrl.catalogItems, ['name', 'tags'], keywords);
-    return _.take(items, 5);
+    let totalNumItems: number = _.size(items);
+    if (totalNumItems > this.maxResultsToShow) {
+      items.splice(this.maxResultsToShow, 0, {id: 'viewAll', name: searchText, totalNumResults: totalNumItems});
```
spadgett: Instead of using `splice` can you use `_.take(items, this.maxResultsToShow)` and then just append to the array? I think it is easier to follow that way.

spadgett:
Nothing major in my review except that we're leaking the rootScope listener

src/components/services-view/services-view.controller.ts

src/components/services-view/services-view.controller.ts
dtaylor113: Fixed, removed in $destroy()

src/components/services-view/services-view.controller.ts
dtaylor113: Yes, 0 did work.
Without these modifications the filter results label would always say "n Results".

src/components/services-view/services-view.html
dtaylor113: Fixed

PR #255, opened 2017-05-26
Check services/plans for bindability before allowing binding 

### LINKED ISSUE #217: Service classes have a boolean for whether they are bindable
We assume all service classes (and their instances) are bindable today.  But the spec allows for services to declare themselves as non-bindable.  The bindable boolean is returned in the service class object.  We need to handle this both from the ordering flow in the main catalog and in the console overview.

cc @ncameronbritt @spadgett @jeff-phillips-18

### LINKED ISSUE #251: Provision/bind wizard steps are inconsistent between verbs/nouns
Not a major issue, but its the kind of minor inconsistency that bothers me.  Some of the steps are nouns and some are verbs.
wizard-step-names

### LINKED ISSUE #257: Provision service bind step should not be hidden even when no apps
The "Bind" step in the provision dialog service should always be displayed if the service class and plan are bindable. It looks like we're hiding it when there are no applications, however.

https://github.com/openshift/origin-web-catalog/blob/master/src/components/order-service/order-service.controller.ts#L260

cc @jwforres

jeff-phillips-18:
Fixes #251 
Fixes #257
Fixes #217 

Also fixes displaying errors when creating a project which already exists
Removes filters that were added to common.

### DIFF
--- a/src/components/order-service/order-service.controller.ts
+++ b/src/components/order-service/order-service.controller.ts
@@ -60,7 +60,7 @@ export class OrderServiceController implements angular.IController {
 
     this.planStep = {
       id: 'plans',
-      label: 'Plans',
+      label: 'Plan',
       view: 'order-service/order-service-plans.html',
       hidden: this.ctrl.plans.length < 2,
       allowed: true,
@@ -77,7 +77,7 @@ export class OrderServiceController implements angular.IController {
       onShow: this.showConfig
     };
     this.bindStep = {
-      label: 'Bind',
+      label: 'Binding',
       id: 'bind',
       view: 'order-service/order-service-bind.html',
       hidden: false,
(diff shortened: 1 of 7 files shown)


jeff-phillips-18:
Prerequisites are complete, removed WIP

src/components/order-service/order-service.controller.ts
```diff
@@ -257,15 +257,25 @@ export class OrderServiceController implements angular.IController {
   }
 
   private updateBindability() {
-    this.bindStep.hidden = _.size(this.ctrl.applications) < 1;
-    this.reviewStep.allowed = this.bindStep.hidden;
-
-    if (this.bindStep.hidden) {
-      this.ctrl.nextTitle = "Create";
+    var planBindable = _.get(this.ctrl.selectedPlan, "bindable");
```
jwforres: Can you just add a comment here explaining that plan bindable is optional so the explicit `===` checks are required.  Don't want someone looking at this later to try to optimize the checks and then break the behavior :)

jeff-phillips-18: Included a comment. Thanks.

jeff-phillips-18:
Added WIP as this now depends on https://github.com/openshift/origin-web-common/pull/90

jeff-phillips-18:
Removed WIP status, ready for final reviews.

PR #265, opened 2017-05-31
Workaround for running unit tests on MacOS

jeff-phillips-18:
Uses Safari as a test browser when available.
Known issue: runs tests 3 times in Safari for some reason.


dtaylor113:
Wonder if the postDetection function is being called multiple times, perhaps for each available browser?

jeff-phillips-18:
@dtaylor113 No, if I don't use detect browsers and just specify Safari as the browser, I get the same.

PR #269, opened 2017-06-02
Handle project name being changed by project template when ordering

jwforres:
Partial fix for https://github.com/openshift/origin-web-console/issues/1588

### DIFF
--- a/src/components/order-service/order-service.controller.ts
+++ b/src/components/order-service/order-service.controller.ts
@@ -185,7 +185,8 @@ export class OrderServiceController implements angular.IController {
       this.DataService
         .create('projectrequests', null, projReqObj, this.$scope)
         .then( (data: any) => {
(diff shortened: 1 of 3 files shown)


src/components/order-service/order-service.controller.ts
```diff
@@ -185,7 +185,8 @@ export class OrderServiceController implements angular.IController {
       this.DataService
         .create('projectrequests', null, projReqObj, this.$scope)
         .then( (data: any) => {
-          this.ctrl.projectDisplayName = newProjDisplayName || newProjName;
+          this.ctrl.selectedProject = data;
+          this.ctrl.projectDisplayName = this.$filter('displayName')(this.ctrl.selectedProject) || this.ctrl.selectedProject.metadata.name;
```
spadgett: I think the `displayName` filter falls back to `metadata.name`, so the `||` shouldn't be necessary.

jwforres:
@spadgett fixed

Issue #259, opened 2017-05-26
Default language subcategories should include .NET and Golang


spadgett:
We have a .NET category in the 3.5 catalog (tag `dotnet`) as well as a Golang category.

https://github.com/openshift/origin-web-console/blob/master/app/scripts/constants.js#L357-L361
https://github.com/openshift/origin-web-console/blob/master/app/scripts/constants.js#L383-L387

cc @jwforres

PR #270, opened 2017-06-02
Add golang and .Net to the ootb language subcategories

jwforres:
Fixes https://github.com/openshift/origin-web-catalog/issues/259

### DIFF
--- a/dist/origin-web-catalogs.js
+++ b/dist/origin-web-catalogs.js
@@ -194,6 +194,11 @@ webpackJsonp([ 0, 1 ], [ function(e, t) {
                 label: "JavaScript",
                 icon: "font-icon icon-js"
             }, {
+                id: "dotnet",
+                label: ".NET",
+                tags: [ "dotnet" ],
+                icon: "font-icon icon-dotnet"
+            }, {
                 id: "perl",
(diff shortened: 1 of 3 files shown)


dist/origin-web-catalogs.js
```diff
@@ -213,6 +218,11 @@ webpackJsonp([ 0, 1 ], [ function(e, t) {
                 label: "Python",
                 tags: [ "python" ],
                 icon: "font-icon icon-python"
+            }, {
+                id: "golang",
+                label: "Go",
+                tags: [ "Golang", "golang", "go" ],
```
jwforres: included Golang with capital G for backwards compatibility with the the old categories

spadgett: I think the old catalog tag checking was case insensitive (but I could be wrong).

spadgett: https://github.com/openshift/origin-web-console/blob/master/app/scripts/services/catalog.js#L28

jwforres: ok i'll double check what it was doing

jwforres:
@spadgett updated to make tag matching case insensitive

PR #280, opened 2017-06-13
Scroll service categories to top on selection

### LINKED ISSUE #267: Tour callout offscreen when no SaaS offerings
screen shot 2017-06-02 at 2 57 23 pm

jeff-phillips-18:
This PR scrolls the ‘Browse Catalog’ section to the top of the landing page when a category is selected. This makes it so the page doesn't 'jump' while browsing through the available categories.

Fixes #267 

Jumping issue:

Before:

scroll-issue

After"

scroll-issue-fixed

### DIFF
--- a/src/styles/services-view.less
+++ b/src/styles/services-view.less
@@ -170,6 +170,18 @@ services-view,
   }
 
   @media(min-width: @screen-xs-min) {
+    .services-no-sub-categories {
+      background: fade(@color-pf-blue-700, 90%);
+      display: flex;
+      flex: 1 1 auto;
+      padding: 15px (@grid-gutter-width / 2) 12px;
+      .services-items {
+        left: 0;
+        margin: 0 0 8px;
+        right: 0;
+      }
+    }
+
     .services-sub-categories {
       align-content: flex-start;
       align-items: flex-start;
--- a/dist/less/services-view.less
+++ b/dist/less/services-view.less
@@ -170,6 +170,18 @@ services-view,
   }
 
   @media(min-width: @screen-xs-min) {
+    .services-no-sub-categories {
+      background: fade(@color-pf-blue-700, 90%);
+      display: flex;
+      flex: 1 1 auto;
+      padding: 15px (@grid-gutter-width / 2) 12px;
+      .services-items {
+        left: 0;
+        margin: 0 0 8px;
+        right: 0;
+      }
+    }
+
     .services-sub-categories {
       align-content: flex-start;
       align-items: flex-start;
--- a/dist/origin-web-catalogs.css
+++ b/dist/origin-web-catalogs.css
@@ -1185,6 +1185,17 @@ services-view,
   text-transform: capitalize;
 }
 @media (min-width: 480px) {
+  .services-view .services-no-sub-categories {
+    background: rgba(0, 34, 53, 0.9);
+    display: flex;
+    flex: 1 1 auto;
+    padding: 15px 20px 12px;
+  }
+  .services-view .services-no-sub-categories .services-items {
+    left: 0;
+    margin: 0 0 8px;
+    right: 0;
+  }
   .services-view .services-sub-categories {
     align-content: flex-start;
     align-items: flex-start;
--- a/dist/origin-web-catalogs.js
+++ b/dist/origin-web-catalogs.js
@@ -356,23 +356,22 @@ webpackJsonp([ 0, 1 ], [ function(e, t) {
             }, {
                 title: "Browse Catalog",
                 content: "If you don’t know exactly what you are looking for, you can browse all available catalog items under the first tab in the catalog.",
-                target: ".services-view-container",
-                targetScrollElement: ".landing",
-                placement: "top",
+                target: ".services-view-container h1",
+                placement: "bottom",
                 xOffset: 20,
+                delay: 300,
                 preShow: c
             }, {
                 title: "Browse by Category",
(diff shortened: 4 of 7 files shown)


jeff-phillips-18:
Also, after further UX feedback, added the margins around the All and Other services areas to be more in line with the sub-categories areas.

rhamilto:
> Also, after further UX feedback, added the margins around the All and Other services areas to be more in line with the sub-categories areas (not shown in the fixed image above).

screen shot 2017-06-13 at 9 18 34 am

src/styles/services-view.less
```diff
@@ -134,6 +134,17 @@ services-view,
       }
     }
 
+    .services-no-sub-categories {
```
rhamilto: Nit:  this rule should come before .services-sub-categories per alpha rule

jeff-phillips-18: Fixed.

jeff-phillips-18:
Fixed unit tests.

beanh66:
Thanks  @jeff-phillips-18! Much improved interaction :)

cshinn:
@jeff-phillips-18 I think it works great, but I wonder if the services area should have side and bottom padding on mobile. Seems a bit narrow.

jwforres:
its a visual nit but it bothers me that the margin above and below items in the white section is not the same

whitespace

spadgett:
I might have found a bug: when I click on a tab in the second roll, it scrolls the items offscreen for me:

screen shot 2017-06-13 at 9 24 56 pm

spadgett:
I'm having trouble consistently reproducing though. It might have been because I resized the window?

serenamarie125:
This stuff looks great!  @jeff-phillips-18 thanks for the improvements, @rhamilto definitely like the additional margins!

Autoscrolling doesn't bother me, it seems useful imo.

jwforres:
Hmm... yeah I'm seeing some weird jumping behavior during browser resize sometimes.

jwforres:
Ok got a screen capture of what I'm seeing, watch the bottom of the SaaS section jump around as you resize.  I suspect it has to do with how resize is being debounced.

jeff-phillips-18:
@jwforres re: margins  Now that the filter stuff is in as well can you re-look and see if it still bothers you.

jeff-phillips-18:
@spadgett Is this that issue w/ scrolling just due to not showing scrollbars until you hover? I notice there are no scrollbars for the project summary panel either.

jeff-phillips-18:
@jwforres I can cut down the debounce timer maxWait, it helps a bit but there will still be some amount of jumping. Any other thoughts?

jwforres:
Still looks bad to me.  I dont think the problem is actually new on the subcategories, it just didnt exist on the All/Other pages before because the white background went to the bottom.  The spacing between the logo and the text also seems quite large to me:

whitespace2

jeff-phillips-18:
Rebased. Set the maxWait for the resize debounce to 50ms.

jeff-phillips-18:
@jwforres it is a bit better with the filter layout refactoring:

image

rhamilto:
I agree with @jwforres regarding the margin around and lack of full height for "All" and "Other".  The lack of connection to the tabs creates a disassociation between the tabs and the content (that is, they no longer feel associated).  And the white not extending all the way to the bottom of the page actually makes the items appear less important because my eye is drawn by the negative space surrounding them.

jeff-phillips-18:
Is this something new to this PR? The only thing changed in this PR is to add the left/right margins. The top/bottom distances ought not have changed. Is there a difference in this disassociation here than there is for the sub-category tiles? It looks the same to me (maybe worse for the sub-categories since there is even more 'negative space').

spadgett:
@jeff-phillips-18 Didn't the white extend to the bottom of the page before? I know it's an intentional change, but I think it's what causes the feeling on unbalanced margins.

For what it's worth, I preferred the All and Other categories as they were before, but I see there's a pretty strong split on that.

jeff-phillips-18:
@spadgett yes, that too. I can go either way. I was just trying to make it consistent when clicking through the categories.

jeff-phillips-18:
@jwforres @rhamilto @spadgett Is this more favorable?

image

jwforres:
I definitely prefer that over the shortened white block. One nit, I'd expect the bottom gutter to be the same as the sides

spadgett:
I prefer that, but I still like the current style better :) I think @rhamilto hit on it when he said the tabs don't feel connected to the content.

spadgett:
I think the autoscroll is an improvement, though

jeff-phillips-18:
I've pushed this change, with the bottom margin fix:

image

jwforres:
i dunno how to make the resize less jumpy, so if we think the improvement we get from the auto-scroll justifies the resize behavior then I'm ok with this.  I would lean towards yes and sounds like @spadgett does?

spadgett:
I wouldn't be against removing debounce if it's causing problems.

If we keep it, it feels more like a `_.throttle` than `_.debounce` when max wait is the same as wait.

https://lodash.com/docs/3.10.1#throttle

We could try `_.throttle` with an even smaller value or remove it altogether.

jeff-phillips-18:
Found a way to not update on resize (using 100vh - offset)! No more jumping!

@jwforres Give it another go and see what you think

jwforres:
soooo much better 👍

PR #286, opened 2017-06-13
Addition of project deletion warning icon and tooltip

sg00dwin:
Brings consistency with how it functions in the web console.
Required inverting the tooltip background/text color, since it would overlay a dark background. 

screen shot 2017-06-13 at 2 47 21 pm
console project list view.


sg00dwin:
@rhamilto 
cc @jwforres @spadgett

jeff-phillips-18:
The Future Work  designs for showing status show the status indicator before the project name:

http://openshift.github.io/openshift-origin-design/web-console/1-homepage/project-panel#projects---list-view

Not sure if we want to head in that direction now, @serenamarie125 @beanh66 ?

jwforres:
FWIW we went with consistency with the project list for now.  @sg00dwin and @rhamilto are looking at better ways to indicate a project has been deleted (strikethrough, grayed out, etc) and we could look at reordering the location of the warning icon as part of that

jwforres:
the changes LGTM, but will wait for @serenamarie125 and @beanh66 to chime in

beanh66:
@jwforres @jeff-phillips-18  This makes sense for 3.6 because we do want to remain consistent with the Homepage. In the next release, we would like to revisit changing this to the left side in both the projects list and homepage though so users can more easily scan for alerts when they exist.

Issue #290, opened 2017-06-13
Display issue when no filter matches at mobile


spadgett:
![openshift web console 2017-06-13 16-31-34](https://user-images.githubusercontent.com/1167259/27103402-7dc17a8e-5056-11e7-8986-810265e97431.png)

PR #293, opened 2017-06-14
Fixes Issues #289

### LINKED ISSUE #289: "1 Items" displayed below filter in services view
Should be `1 Item` (singular).

openshift catalog components 2017-06-13 16-17-58

dtaylor113:
## Adjusted filter results label to be 'item' or 'items', fixes #289
image

image


src/components/services-view/services-view.controller.ts
```diff
@@ -192,7 +192,7 @@ export class ServicesViewController implements angular.IController {
         this.ctrl.filterConfig.resultsCount = this.ctrl.filteredItems.length;
         $('.toolbar-pf-results h5').text(this.ctrl.filterConfig.resultsCount + ' of ' + this.ctrl.totalCount + ' items');
       } else {
-        $('.toolbar-pf-results h5').text(this.ctrl.totalCount + ' items');
+        $('.toolbar-pf-results h5').text(this.ctrl.totalCount + ' item' + (this.ctrl.totalCount > 1 ? 's' : ''));
```
spadgett: Do we ever need to worry about "0 items", which should be plural? I guess the subcategory would be hidden, but in case that ever changes, suggest instead `this.ctrl.totalCount !== 1`

spadgett: Nit: This will be easier to translate later on if we change it to

```js
(this.ctrl.totalCount === 1) ? 'item' : 'items'
```

dtaylor113: Hi @spadgett, for 0 results it will read "0 of 15 items".
Also, sorry, snuck into this PR a fix for another small issue.

rhamilto:
It appears @dtaylor113 and I both have fixes for #290.  I strongly recommend we go with [my CSS fixes](https://github.com/openshift/origin-web-catalog/pull/295) as there was unnecessarily complex layout being done and duplication of rules.

dtaylor113:
> It appears @dtaylor113 and I both have fixes for #290. I strongly recommend we go with my CSS fixes as there was unnecessarily complex layout being done and duplication of rules.

Ok @rhamilto, I'll remove #290 from this PR

spadgett:
Thanks @dtaylor113

spadgett:
Looks like this needs a rebase

dtaylor113:
> Looks like this needs a rebase

@spadgett, all set

PR #301, opened 2017-06-15
Bug fix where filter was too big at 320px - 334px viewports

rhamilto:
Before:
screen shot 2017-06-15 at 9 49 42 am

After:
screen shot 2017-06-15 at 9 49 22 am

### DIFF
--- a/dist/less/services-view.less
+++ b/dist/less/services-view.less
@@ -132,7 +132,9 @@ services-view,
 
       .form-group {
         margin-bottom: 0;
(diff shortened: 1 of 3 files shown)


jeff-phillips-18:
Is 320px the lowest resolution we need to support? I could see this still being an issue if the resolution got smaller.

rhamilto:
> Is 320px the lowest resolution we need to support?

320px is the lowest the console supports, so I assume the same is true for the catalog.

PR #303, opened 2017-06-15
Fixes 'Next never enables if provision service and no projects exist'

### LINKED ISSUE #294: Next never enables if provision service and no projects exist
If you are provisioning a service the Create Project option is automatically selected, if you then fill out the rest of the form as needed the Next button never gets enabled.

Note this does not happen with Create Project if any projects already exist.

dtaylor113:
Fixes #294 

Bug only affected provisioning a service class.  Provisioning builder images uses slightly different logic and isn't effected by this bug.

### DIFF
--- a/app/mockServices/mockData/services.ts
+++ b/app/mockServices/mockData/services.ts
@@ -15,6 +15,7 @@ export const servicesData = {
     plans: [
       {
         name: 'rh-ded-topic', osbGuid: '1', displayName: 'Red Hat - Dedicated - Topic', description: '$.65 / 1 Million messages', bullets: ['One', 'Two', 'Three'],
+        bindable: true,
         alphaInstanceCreateParameterSchema: {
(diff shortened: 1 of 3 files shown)


app/mockServices/mockData/services.ts
```diff
@@ -15,6 +15,7 @@ export const servicesData = {
     plans: [
       {
         name: 'rh-ded-topic', osbGuid: '1', displayName: 'Red Hat - Dedicated - Topic', description: '$.65 / 1 Million messages', bullets: ['One', 'Two', 'Three'],
+        bindable: true,
```
spadgett: @dtaylor113 `bindable` is required on the service class. The value on the plan is only needed to override the service class value.

We should add it to the all the mock data service classes. (Does not need to be this PR.)

spadgett: I opened #304 to add bindable to the mock data.

PR #307, opened 2017-06-16
Minor Spacing & Sizing Tweaks

### LINKED ISSUE #299: Minor Spacing & Sizing Tweaks
@jeff-phillips-18 @serenamarie125  Hi! Here are two things that shouldn't take very long, but will provide a lot of "bang for your buck" to make the screen look more elegant:

- The names of the catalog items (ex: “Node.js”) should be closer to their respective icons… moved up 10px closer.

- The four large icons in that top portion should be smaller… about 30% smaller. Making them smaller would create a bit more "breathing room" (aka whitespace) if the same panel height stays the same. They seem a little "in your face" at the current size.

This example of the spacing/sizing looks rough because I was editing a flat screenshot (hence the weird blue cutout boxes in the background that wouldn't really be there) Let me know if you have any questions... thanks!

changes

jeff-phillips-18:
Fixes #299 

Adds a more accurate calculation for size of services view.

Resulting console:

image

@jennyhaines @serenamarie125 @beanh66 @cshinn

### DIFF
--- a/src/styles/variables.less
+++ b/src/styles/variables.less
@@ -1,4 +1,3 @@
-@card-icon-lg:                                        108px;
 @card-icon-sm:                                        72px;
 @card-icon-xs:                                        24px;
 @landing-bg:                                          url(@landing-image) no-repeat fixed @landing-main-bg;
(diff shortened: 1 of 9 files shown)


rhamilto:
"Icons" can also be images.  Did you test the the image sizing is still good?

src/styles/variables.less
```diff
@@ -1,4 +1,4 @@
-@card-icon-lg:                                        108px;
+@card-icon-lg:                                        72px;
```
rhamilto: Since the value of @card-icon-lg is now the same as @card-icon-sm, we don't both in addition to the overrides that make use of @card-icon-lg.

jeff-phillips-18: Yeah, I left it in case we ever wanted to change it to be different but I can remove it.

jeff-phillips-18:
@rhamilto Tested with images and it works well.

jennyhaines:
Looks great. Thanks for the sizing/spacing tweaks! Ready to Merge

src/styles/variables.less
jeff-phillips-18: Removed @card-icon-lg variable and references to it.

PR #313, opened 2017-06-20
Utilizing openshift-logos-icon dependency instead of local copy

rhamilto:
And removing obsolete openshift-icon


jwforres:
going to go ahead and merge since this doesn't affect what gets shipped in dist

PR #317, opened 2017-06-20
Browse animation of catalog items and nav

sg00dwin:
@spadgett


spadgett:
@sg00dwin I have to say this is so much better. Thanks for working on this. It's a lot more polished.

I noticed some strangeness when filtering items using this input:

openshift web console 2017-06-20 20-14-44

It already triggered an animation my with previous PR (my bug), but it seems a little worse with these changes.

sg00dwin:
@spadgett updated pr addresses filter transition issues

src/styles/animations.less
```diff
@@ -34,3 +34,12 @@
     transform: translateY(-50px);
   }
 }
+
+@keyframes catalogItemFade {
```
rhamilto: Nit:  alpha (this rule should come first in the file)

src/styles/services-view.less
```diff
@@ -223,17 +230,16 @@ services-view,
...
         }
       }
 
-      &.active.items-shown {
-        .services-sub-category-tab:after,
-        .services-items {
+      &.active {
+        .services-sub-category-tab:after {
           opacity: 1;
+          animation: catalogItemFade 200ms ease-out;
```
rhamilto: Nit:  alpha

src/styles/services-view.less
```diff
@@ -340,13 +360,59 @@ services-view,
...
+      }
+
+      &.ng-leave {
+        transition: 200ms ease-in opacity;
+        opacity: 1;
+        &.ng-leave-active {
+          opacity: 0;
+        }
+      }
+      &.ng-enter {
```
rhamilto: Nit:  alpha

rhamilto:
I'm seeing an unnecessary scrollbar now.  That is, the height of the items are is much larger than it needs to be.

screen shot 2017-06-22 at 10 57 31 am

Related, the subcategories background doesn't extend all the way to the bottom.
screen shot 2017-06-22 at 10 59 51 am

sg00dwin:
tested across Chrome, FF, Safari, IE
@rhamilto  Found a minor glitch on Safari & IE that I'll be pushing a fix for shortly

sg00dwin:
@rhamilto pushed changes to address comments. Except for the excessive height issue, which appears to be in master.

src/styles/services-view.less
```diff
@@ -223,16 +230,15 @@ services-view,
...
         .services-sub-category-tab-image img {
           opacity: 1;
         }
       }
 
-      &.active.items-shown {
-        .services-sub-category-tab:after,
-        .services-items {
+      &.active {
+        .services-sub-category-tab:after {
```
rhamilto: These should be part of the rule at line 220, no?  Line 231 sets the opacity to 0, but this rule sets it back to 1.

src/styles/services-view.less
```diff
@@ -340,15 +359,57 @@ services-view,
...
+        bottom: 0;
+        left: 0;
+        position: absolute;
+        right: 0;
+        top: 80px;
+      }
+    }
+
+    .services-no-sub-categories.ng-leave.ng-leave-active {
+          // fade out inner items
```
rhamilto: Nit:  indentation.  I recommend setting Atom to use spaces instead of tabs.

sg00dwin:
@rhamilto pr updated

PR #323, opened 2017-06-21
By default, do not auto-launch landing page tour on first access

### LINKED ISSUE #321: Disabled the tour auto-start by default
@jeff-phillips-18 Per the latest test results from Summit, the auto-start for the guided tour should be disabled by default, but this functionality should remain available.

jeff-phillips-18:
Fixes #321


spadgett:
LGTM, we should wait for the stream to reopen for non-blockers. Thanks @jeff-phillips-18

PR #332, opened 2017-06-27
Fix to show error in ordering panel when selecting a project user cannot add to

### LINKED ISSUE #318: Project select in ordering panel shouldn't show projects where I am only a viewer
The select project dropdown when ordering anything should not show projects that you can't add to. 

We normally use AuthorizationService.canIAddToProject to check a particular project, but this would require a separate rules check against the server for every project that comes back in your project list.

An alternative would be to show the projects in the dropdown but do a rules check on it after its been selected and disable the Next button if you don't have the correct rights.  Not optimal user experience but possibly the best we can do from a performance perspective right now.

This was never an issue before because we were in the context of a project when adding and had already done the rules check.

dtaylor113:
image

* Switching between projects, including 'Create Project', clears/sets error appropriately
* Next button enabled/disabled appropriately
* Tested ordering a Service and a Builder Image
* Local bower linked console to use these changes, no issues seen
* Let me know if you want a perhaps less severe error message.

Fixes #318


jwforres:
Should be more clear that is an issue of permissions in the error? Otherwise I may not understand why I can't add to that project.  Maybe "You are not authorized to add to this project." ?

beanh66:
Also wondering if the field should be highlighted in red. In PatternFly the [errors and validation](http://www.patternfly.org/pattern-library/forms-and-controls/errors-and-validation/) for forms has the field highlighted. Thoughts @openshift/team-ux-review?

ncameronbritt:
Should you even be able to select a project which you don't have permission to add to?

dtaylor113:
> Should you even be able to select a project which you don't have permission to add to?

Hi @ncameronbritt , pls see the original comments for #318, this was a compromise for performance issues.
Thanks,
- Dave

ncameronbritt:
Thanks @dtaylor113

jwforres:
@ncameronbritt yeah i wish we had a way to hide them but the permission check is very expensive even for just a single project

dtaylor113:
> Should be more clear that is an issue of permissions in the error? Otherwise I may not understand why I can't add to that project. Maybe "You are not authorized to add to this project." ?

@jwforres , updated the error message, pls see updated screenshot -thanks

dtaylor113:
> Also wondering if the field should be highlighted in red. In PatternFly the errors and validation for forms has the field highlighted. Thoughts @openshift/team-ux-review?

If we want to do this then I'll file an Issue because, to be consistent, we'd want to update all the ordering panel form fields to have this behavior.

jwforres:
thanks @dtaylor113 and I agree on opening a separate issue to address red highlighting on errors, I thought that was working in other form inputs that we have, its possible this is specific to everywhere that we have ui-select inputs

dtaylor113:
Hi @jwforres, you are right, I see it working on the input field, but not the ui-select:

image

I filed Issue #334 to track this

PR #335, opened 2017-06-28
Bump version to 0.0.27, use common 0.0.39

### LINKED ISSUE #279: Need to address terminology inconsistencies in the binding step for ordering
Binding from the overview in the console probably has same issues.

jeff-phillips-18:
Updates parameters to common bind forms as necessary.

Fixes #279


src/components/order-service/order-service.controller.ts
```diff
@@ -164,6 +164,12 @@ export class OrderServiceController implements angular.IController {
     this.ctrl.nextTitle = 'Create';
     this.reviewStep.allowed = this.bindStep.valid;
 
+    if (this.isNewProject()) {
+      this.ctrl.projectDisplayName = this.ctrl.selectedProject.metadata.annotations['new-display-name'] || this.ctrl.selectedProject.metadata.name;
+    } else {
+      this.ctrl.projectDisplayName = this.$filter('displayName')(this.ctrl.selectedProject);
+    }
+
```
dtaylor113: So this PR seems to be a little more than just a version bump.  What are these other changes? -thanks

src/components/order-service/order-service-bind.html
```diff
@@ -3,6 +3,7 @@
                    applications="$ctrl.applications"
                    form-name="$ctrl.forms.bindForm"
                    allow-no-binding="true"
+                   project-name="$ctrl.projectDisplayName"
```
jwforres: we probably should have named the variable project-display-name so that someone doesn't try to use it later as the unique name for something, can you open an issue to fix that later?

jeff-phillips-18:
@jwforres opened issue https://github.com/openshift/origin-web-common/issues/122

PR #336, opened 2017-06-28
WIP - Update catalog browsing snap mechanism

### LINKED ISSUE #325: Unnecessary scrollbar when saas offerings are present
When the saas offerings are present, the height of the main content area is set to a height larger than needed to display the relevant items.

In this screenshot, note there is plenty of room for all the items to display, yet a scrollbar is present that allows you to scroll to nothing.  If the saas offerings are removed, the bug goes away.

screen shot 2017-06-23 at 9 03 12 am

It's very noticeable in Windows (and Linux) where the scrollbars are visible:
screen shot 2017-06-23 at 9 11 45 am

jeff-phillips-18:
Changed from adding additional height to the services view section to
hiding/showing the SAAS offerings.

Fixes #325

Still working on animating the snappage

@openshift/team-ux-review

### DIFF
--- a/app/pages/home/home-page.html
+++ b/app/pages/home/home-page.html
@@ -1,4 +1,4 @@
-<landing-page base-project-url="{{$ctrl.baseProjectUrl}}">
+<landing-page base-project-url="{{$ctrl.baseProjectUrl}}" show-header="$ctrl.saasOfferings.length" collapsed-header-text="Show SAAS Offerings">
   <landingsearch>
     <catalog-search catalog-items="$ctrl.catalogItems" base-project-url="{{$ctrl.baseProjectUrl}}"></catalog-search>
   </landingsearch>
--- a/dist/less/landing-page.less
+++ b/dist/less/landing-page.less
@@ -95,6 +95,26 @@ landingbody,
   min-height: 100%;
 }
 
+.landing-header-expand-container {
+  margin: 5px 0;
+  text-align: center;
+}
+
+.landing-header-expander {
+  color: @color-pf-white;
+  font-size: @font-size-base;
+  &:hover, &:focus {
+    color: @color-pf-white;
+    text-decoration: none;
+  }
+  &:after {
+    content: @fa-var-angle-down;
+    font-family: @icon-font-name-fa;
+    font-size: (@font-size-base + 2);
+    padding: 0 5px;
+  }
+}
+
 .landing-search-form {
   margin: 0 auto;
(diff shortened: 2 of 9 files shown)


jeff-phillips-18:
snap-services

app/pages/home/home-page.html
```diff
@@ -1,4 +1,4 @@
-<landing-page base-project-url="{{$ctrl.baseProjectUrl}}">
+<landing-page base-project-url="{{$ctrl.baseProjectUrl}}" show-header="$ctrl.saasOfferings.length" collapsed-header-text="Show SAAS Offerings">
```
jwforres: I would not refer to these as SaaS Offerings in any UI text

jwforres:
So selecting any categories hides the top section completely and you have to explicitly choose to expand it? Is that really the behavior we want?

I'm not really comfortable making an experience change like this so late in the release.

beanh66:
@jeff-phillips-18 I think the animation now helps give the user context that we are scrolling them down to the relevant info. IMO allowing users to scroll back up would do the trick, not positive we need the link/button for this. @rhamilto, @serenamarie125, @ncameronbritt thoughts?

app/pages/home/home-page.html
rhamilto: > I would not refer to these as SaaS Offerings in any UI text

+1

Additionally, I find the "Show..." link has visual affordances of a dropdown and/or a toggle for the catalog below (i.e., it feels like a control for the catalog, not for the now collapsed offerings area) because of the hard lines above and below the link.

screen shot 2017-06-28 at 3 21 39 pm

When the offerings are present, I that's not the case.

screen shot 2017-06-28 at 3 22 26 pm

jeff-phillips-18:
This is just an attempt at resolving #325 
Personally, I prefer the extra scroll space to the hide/show mechanism. I just put this out there to see what the UX preferences are.

ncameronbritt:
I agree with @beanh66. The animation should show the user what has happened. IMHO those "back to top" buttons are most useful when the page is REALLY long.

jwforres:
Just making sure its clear that there is a major behavioral difference
here.  You will not be able to just scroll back up to see the SaaS
section.  You MUST click that link to see that section again.

ncameronbritt:
Sorry, I think @beanh66  and I were both suggesting that the page would scroll up, and you could scroll back down--no link. That, I believe, was the behavior before.

PR #338, opened 2017-06-28
Show search 'View All' regardless of results count

jeff-phillips-18:
also removed the auto-select of the first item and autocomplete for the input. First item shows selected on a down arrow key.

Fixes https://github.com/openshift/origin-web-console/issues/1758

@openshift/team-ux-review 

search-typeahead


jwforres:
im guess its saying "view all 1 results" when there is only 1 since no special logic was added to handle 1 result

jeff-phillips-18:
@jwforres correct. Not sure what the appropriate wording would be there. "View all 1 result" doesn't make much sense either.

jwforres:
Yeah it sounds weird either way :)

ncameronbritt:
Hopefully that one result is the thing they're looking for and the won't care what it says :-)

jeff-phillips-18:
For one result:

image


For no results:

image

jwforres:
works for me  @openshift/team-ux-review do yall have any other feedback before we merge it?

ncameronbritt:
Works for me.

PR #340, opened 2017-06-29
Partially rollback the authorization checks for project selector

jwforres:


jwforres:
See my comments with the change for more details but basically the changes necessary to make this work correctly are significant enough at this point that I don't think its worth the risk just for tech preview.  You'll still get an error from the API as soon as you actually try to create the service instance or new app.

It wasn't obvious an error existed when using mocked data / controllers. I really feel like we should stop using mock data as the default when developing, it is sufficiently easy at this point to get the catalog running with oc cluster up.

dtaylor113:
> It wasn't obvious an error existed when using mocked data / controllers. I really feel like we should stop using mock data as the default when developing, it is sufficiently easy at this point to get the catalog running with oc cluster up.

This ones on me, I did run a quick test in console but obviously it wasn't enough.   My apologies, going forward I'll be sure to fully test code in the console.
Thanks,
Dave

PR #356, opened 2017-07-11
Update to angular-patternfly v4.1.1 and patternfly v3.25.1.  Bumped version

dtaylor113:
WIP - Do Not Merge!

Dependency | Version & Changes
-- | -- |
patternfly | 3.25.1
angular-patternfly | 4.1.1
jQuery | ~2.1.4 => 3.2.1
lodash | ~3.10.1 => 4.17.4 <br>_The following changes were made:_ <br>_.first to _.head <br> _.indexBy to _.keyBy <br>_.trunc to _.truncate <br>_.sortByAll into _.sortBy <br> _.assign into _.assignWith <br> _.omit into _.omitBy

**Note:** angular-bootstrap resolving to lower version (0.14.3) to match console.  When console updates to angular-bootstrap v2.2.x (to match angular-patternfly v4.1.1), then catalog will update as well.

Affected areas are the pfWizards used in the Ordering Panel and the pfFilter used in Service (Browse Catalog) view.


bower.json
```diff
@@ -34,12 +34,12 @@
     "angular-resource": "1.5.11",
     "angular-sanitize": "1.5.11",
     "angular-route": "1.5.11",
-    "angular-bootstrap": "0.14.3",
-    "angular-patternfly": "3.24.0",
+    "angular-bootstrap": "2.2.x",
```
spadgett: We can't bump this here until we also bump it in origin-web-console

dtaylor113: Why not?  In console we resolve to the lower version.

spadgett: You're right: we can. But I'd prefer to keep them at the same level. If we (unintentionally) use a new angular-bootstrap API here, it will work, but break when we pull it into console. Those kinds of errors are hard to track down.

I don't think the changes are too hard to just bump it everywhere. See my comments in https://github.com/openshift/origin-web-console/issues/1861

dtaylor113: Fixed the angular-bootstrap issue in console, think we're good to go.  Thanks for the tip!

bower.json
```diff
@@ -84,8 +84,11 @@
   },
   "resolutions": {
     "moment": "~2.18.1",
-    "patternfly": "~3.24.0",
-    "jquery": "2.1.4",
-    "font-awesome": "~4.7.0"
+    "patternfly": "~3.25.1",
+    "jquery": "3.2.1",
+    "font-awesome": "~4.7.0",
+    "lodash": "4.x",
```
spadgett: A little worried that we need this resolution. Is one of our dependencies relying on lodash 3?

dtaylor113: I think this was due to pulling in the current origin-web-common.   Once the common PR is merged and released I'll try again without the resolution.

spadgett: OK, we need to test everything together in the console before we merge any of the PRs. I can assist with this.

package.json
```diff
@@ -91,7 +91,7 @@
     "less": "^2.3.1",
     "less-loader": "^2.2.2",
     "live-server": "^0.9.2",
-    "lodash": "3.10.1",
+    "lodash": "4.17.4",
```
spadgett: Here I think it can be `~4.17.4`. We'll look in origin-web-console

PR #360, opened 2017-07-13
Avoid error when no exposed ports in builder image

spadgett:
Fixes https://github.com/openshift/origin-web-console/issues/1842

cc @johnkim76


spadgett:
@jwforres This is the same behavior as ApplicationGenerator:

https://github.com/openshift/origin-web-console/blob/master/app/scripts/services/applicationGenerator.js#L398-L400

jwforres:
can merge once travis passes

PR #362, opened 2017-07-13
Avoid runtime error when service class displayName set to null

spadgett:
Fix a problem where a runtime error occurs when the `displayName` for a service class is explicitly set to `null` (rather than undefined).

@dymurray


spadgett:
@jwforres Hold, there is another problem

spadgett:
@jwforres Provision dialog has runtime errors with this service class. I'll look at this more tomorrow.

PR #364, opened 2017-07-14
Bug 1471040 - Don't block builder flow if can't get instances

spadgett:
- Don't block the user from creating from builder if there's an error getting the instances
- Fix an error where the watch isn't closed when switching projects, which means the instances listed in the dialog might be for the wrong project

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1471040


jwforres:
approve for 3.6 once travis passes

PR #365, opened 2017-07-14
Adding Deploy Image and Import YAML / JSON options

rhamilto:
https://trello.com/c/Dk9IxmCH

FOLLOW ONs:

- [ ] change the presentation of the links to a dropdown at smaller resolutions once @jeff-phillips-18 addition of add from another project lands


rhamilto:
@openshift/team-ux-review, we may want to rethink the placement of the Deploy Image and Import YAML / JSON links.

With the default set of categories, we're running out of space to accommodate the new links on a single line at various resolutions between 480px and ~1000px.  Given the categories are customizable, it's even less predictable if the links will fit given there could be more categories.

Here are a few screenshots to illustrate what is happening:

localhost-9001-home 1
localhost-9001-home
localhost-9001-home 2
localhost-9001-home ipad
localhost-9001-home iphone 6

One option would be to put the new links alongside "Browse Categories", which may better communicate that these options are siblings of browsing rather than a child.  Other ideas?

ncameronbritt:
What about including those links at the same level as "Browse Catalog"? Those three are lumped together in the same menu inside of a project.

serenamarie125:
we met on this on Friday. @beanh66 should be adding mocks with suggested designs on Monday

beanh66:
@rhamilto We met Friday as @serenamarie125 mentioned and came up with the following:
-On the two smallest size screens we can use the mobile approach where the two options are placed below the catalog. 
-On all other size screens we will put the links on the same line as "Browse Catalog" as @ncameronbritt suggested. 
-The links should be white but we will maintain the same blue hover color.

artboard 1
artboard 2
artboard 3
artboard 4
artboard 5

rhamilto:
Having the mobile layouts differ that much poses some technical challenges that probably aren't worth the overhead given our limited mobile usage.  To keep things simpler, I propose we look at maintaining the link position alongside "Browse Catalog", but maybe truncate a label as needed for the smallest mobile layouts (320px - ~400px)?  For example, if "YAML / JSON" is hidden from "Import YAML / JSON", everything still fits at the smallest viewport width (320px), and there is plenty of room for full labels at 400px and up.  Thoughts?

320px
localhost-9001-home iphone 5

400px
localhost-9001-home 1

beanh66:
@rhamilto Understood on the overhead and agreed we don't want to spend too much extra time on the mobile view. What you have should work pending a few suggestions from @jennyhaines in order to distinguish these options from the "Browse Catalog" title. 
-Change the link color to: #7bd2f2 at all sizes and remain at the 12px font size for mobile

jennyhaines:
@rhamilto - don't kill me, but I prefer the previous version, of the links in white, and at 11px font. I know technically from an accessibility standpoint, the smallest size that text should ever be on mobile is 12px, but 11px looks so much nicer, and this is just one instance.

src/components/services-view/services-view.html
```diff
@@ -3,7 +3,18 @@
     <div class="spinner spinner-xl"></div>
   </div>
   <div ng-if="$ctrl.loaded" class="services-view-container mobile-{{$ctrl.mobileView}}-view">
-    <h1>Browse Catalog</h1>
+    <div class="add-methods">
+      <h1>Browse Catalog</h1>
+      <!-- TODO:  make it so add-other links only appear if cb exists -->
+      <ul class="add-other">
+        <li>
```
jeff-phillips-18: adding ng-if="$ctrl. onDeployImageSelected" to this li should resolve the TODO

src/components/services-view/services-view.html
```diff
@@ -3,7 +3,18 @@
...
+          <a href="" ng-click="$ctrl.deployImageClicked()">Deploy Image</a>
+        </li>
+        <li>
```
jeff-phillips-18: and  ng-if="$ctrl. onFromFileSelected" to this li

src/components/services-view/services-view.html
```diff
@@ -3,7 +3,18 @@
     <div class="spinner spinner-xl"></div>
   </div>
   <div ng-if="$ctrl.loaded" class="services-view-container mobile-{{$ctrl.mobileView}}-view">
-    <h1>Browse Catalog</h1>
+    <div class="add-methods">
+      <h1>Browse Catalog</h1>
+      <!-- TODO:  make it so add-other links only appear if cb exists -->
```
jeff-phillips-18: You can remove this  TODO now.

rhamilto:
Thanks to @jeff-phillips-18, the links now conditionally appear.

PR #370, opened 2017-07-24
Support secret parameters for service instances

spadgett:
Use a secret for parameters when provisioning a service.

https://trello.com/c/IlDIJAz2

### DIFF
--- a/src/components/order-service/order-service.controller.ts
+++ b/src/components/order-service/order-service.controller.ts
@@ -3,7 +3,7 @@ import * as _ from 'lodash';
 
 export class OrderServiceController implements angular.IController {
 
-  static $inject = ['$scope', '$filter', 'AuthService', 'ProjectsService', 'DataService', 'BindingService', 'Logger', 'Constants'];
+  static $inject = ['$scope', '$filter', 'AuthService', 'ProjectsService', 'DataService', 'BindingService', 'Logger', 'Constants', 'DNS1123_SUBDOMAIN_VALIDATION'];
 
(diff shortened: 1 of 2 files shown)


spadgett:
@jeff-phillips-18 PTAL, this should work now with the latest catalog image

src/components/order-service/order-service.controller.ts
```diff
@@ -218,18 +220,29 @@ export class OrderServiceController implements angular.IController {
...
-      this.ctrl.serviceInstance = data;
+      this.watchResults(resource, serviceInstance, context);
+      this.ctrl.serviceInstance = serviceInstance;
+
+      // Create the parameters secret if necessary.
+      if (secretName) {
+        let secret = this.makeParametersSecret(secretName, parameters, serviceInstance);
+        this.DataService.create('secrets', null, secret, context).then(_.noop, (e: any) => {
+          this.ctrl.error = _.get(e, 'data');
+        });
```
jeff-phillips-18: Is there an href to the secret that would get could get and pass to the bind results component now?

spadgett: There are two secrets now: one for the parameters and one for the bind result. I think the bind result is the interesting one. I'm not sure it's worth linking to the parameters secret.

Ideally we'd just have the instance/bind parameters on the provisioned service detail page.

jeff-phillips-18: OK, gotcha.

spadgett:
Thanks @jeff-phillips-18

PR #375, opened 2017-08-07
In Project Summary sidebar, order by most recently viewed projects

dtaylor113:
DO NOT MERGE until origin-web-common's version is bumped!

This is the second PR towards resolving issue [console:#1895](https://github.com/openshift/origin-web-console/issues/1895)

ProjectSummary updated to use RecentlyViewedProjects.orderByMostRecentlyViewed(..).  CreateFromBuilder and OrderService wizards updated to use ProjectServices.create(...).

### DIFF
--- a/app/mockServices/mockProjects.service.ts
+++ b/app/mockServices/mockProjects.service.ts
@@ -49,4 +49,32 @@ export class ProjectsService implements IProjectsService {
     return deferred.promise;
   }
 
+  public create(name: string, displayName: string, desc: string) {
+    let deferred = this.$q.defer();
+    let context: any = {};
+
+    let project: any = {
+      apiVersion: "v1",
+      kind: "Project",
+      metadata: {
+        name: name,
+        annotations: {}
+      }
+    };
+
+    if (displayName) {
+      project.metadata.annotations["openshift.io/display-name"] = displayName;
+    }
+
+    if (desc) {
+      project.metadata.annotations["openshift.io/description"] = desc;
+    }
+
+    setTimeout(() => {
+      deferred.resolve([project, context]);
+    }, 300);
+
+    return deferred.promise;
+  }
+
 }
--- a/src/components/order-service/order-service.controller.ts
+++ b/src/components/order-service/order-service.controller.ts
@@ -202,20 +202,11 @@ export class OrderServiceController implements angular.IController {
       let newProjName = this.ctrl.selectedProject.metadata.name;
       let newProjDisplayName = this.ctrl.selectedProject.metadata.annotations['new-display-name'];
       let newProjDesc = this.$filter('description')(this.ctrl.selectedProject);
-      let projReqObj: any = {
-        apiVersion: "v1",
-        kind: "ProjectRequest",
-        metadata: {
-          name: newProjName
-        },
-        displayName: newProjDisplayName,
-        description: newProjDesc
-      };
-      this.DataService
-        .create('projectrequests', null, projReqObj, this.$scope)
-        .then( (data: any) => {
-          this.ctrl.selectedProject = data;
-          this.ctrl.projectDisplayName = this.$filter('displayName')(this.ctrl.selectedProject);
+      this.ProjectsService
+        .create(newProjName, newProjDisplayName, newProjDesc)
+        .then( (project: any) => {
+          this.ctrl.selectedProject = project;
+          this.ctrl.projectDisplayName = this.$filter('displayName')(project);
           this.createService();
         }, (result: any) => {
           this.ctrl.error = result.data;
--- a/src/components/projects-summary/projects-summary.controller.ts
+++ b/src/components/projects-summary/projects-summary.controller.ts
@@ -4,16 +4,14 @@ import * as _ from 'lodash';
 export class ProjectsSummaryController implements angular.IController {
   // alphabetically please
   static $inject = [
-    '$element',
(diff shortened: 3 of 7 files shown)


bower.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "origin-web-catalog",
-  "version": "0.0.34",
+  "version": "0.0.35",
```
jeff-phillips-18: We  shouldn't bump the version as part of a separate PR. Bumping the version should be its own PR to avoid conflicts and version churn.

dtaylor113: Ok, going forward I will do separate PRs for version bumps.

jeff-phillips-18: I would still suggest removing it from the PR. If/when it is merged, a separate PR to bump the version can be added and resolved quickly.

Issue being, if we want to bump w/o waiting on the PR or want to hold the version bump for additional PRs, then we can if we do not bump within this PR.

app/mockServices/mockProjects.service.ts
```diff
@@ -49,4 +49,16 @@ export class ProjectsService implements IProjectsService {
     return deferred.promise;
   }
 
+  public create(name: string, displayName: string, desc: string) {
+    let deferred = this.$q.defer();
+    var project: any = {};
```
spadgett: Let's return something like

```js
let project: any = {
  apiVersion: "v1",
  kind: "Project",
  metadata: {
    name: name,
    annotations: {}
  }
}

if (displayName) {
  project.metadata.annotations["openshift.io/display-name"] = displayName;
}

if (desc) {
  project.metadata.annotations["openshift.io/description"] = desc;
}
```

src/components/order-service/order-service.controller.ts
```diff
@@ -202,19 +202,10 @@ export class OrderServiceController implements angular.IController {
...
-      };
-      this.DataService
-        .create('projectrequests', null, projReqObj, this.$scope)
-        .then( (data: any) => {
-          this.ctrl.selectedProject = data;
+      this.ProjectsService
+        .create(newProjName, newProjDisplayName, newProjDesc)
+        .then( (project: any) => {
+          this.ctrl.selectedProject = project;
           this.ctrl.projectDisplayName = this.$filter('displayName')(this.ctrl.selectedProject);
```
spadgett: Not new, but let's pass `project` here instead of `ctrl.selectedProject`. The project request template could potentially change the display name, so better to use what we get back.

src/components/projects-summary/projects-summary.controller.ts
```diff
@@ -31,7 +32,8 @@ export class ProjectsSummaryController implements angular.IController {
   private AuthService: any;
   private DataService: any;
   private Constants: any;
-  private RecentlyViewed: any;
+  private RecentlyViewedItems: any;
+  private RecentlyViewedProjects: any;
```
spadgett: Nit: I'd rather make it match the service name exactly, so `RecentlyViewedProjectsService` here and below.

src/components/projects-summary/projects-summary.controller.ts
```diff
@@ -136,6 +140,7 @@ export class ProjectsSummaryController implements angular.IController {
     var projects: any = _.toArray(projectData.by('metadata.creationTimestamp'));
     var orderByDate: any = this.$filter('orderObjectsByDate');
     this.ctrl.projects = orderByDate(projects, true);
+    this.ctrl.projects = this.RecentlyViewedProjects.orderByMostRecentlyViewed(this.ctrl.projects);
```
spadgett: Per my comment in the common PR, would just call `orderByMostRecentlyViewed` here and do the fallback ordering by creation date inside of `orderByMostRecentlyViewed`.

app/mockServices/mockProjects.service.ts
dtaylor113: Fixed

src/components/order-service/order-service.controller.ts
dtaylor113: Fixed

src/components/projects-summary/projects-summary.controller.ts
dtaylor113: Fixed

src/components/projects-summary/projects-summary.controller.ts
dtaylor113: Fixed

dtaylor113:
Removed the version bump from this PR

src/components/projects-summary/projects-summary.controller.ts
```diff
@@ -14,6 +14,7 @@ export class ProjectsSummaryController implements angular.IController {
     'DataService',
     'Logger',
     'ProjectsService',
+    'RecentlyViewedProjectsService',
     'RecentlyViewedServiceItems'
   ];
```
jeff-phillips-18: $filter is no longer used.

jeff-phillips-18: $element and Catalog are not used (not this PR).

dtaylor113: Removed those unused injections.  Fixed

jeff-phillips-18:
Bah! Did not mean to hit merge.

spadgett:
Probably easiest to just bump common in a follow on now that this is merged

PR #385, opened 2017-08-17
Changing display of add other links to dropdown at sm and md resolutions

rhamilto:
so that a third link can be accommodated.

Note:  this screenshot shows a third link, but that's just for mock purposes.  @jeff-phillips-18 is to add the third link.
screen shot 2017-08-17 at 9 08 38 am


cshinn:
Looks great @rhamilto! "Custom Add" should be in headline case to conform to PF guidelines though

rhamilto:
> Looks great @rhamilto! "Custom Add" should be in headline case to conform to PF guidelines though

Thanks, @cshinn.  I've fixed the casing.

src/components/services-view/services-view.html
```diff
@@ -5,14 +5,34 @@
...
       <h1>Browse Catalog</h1>
-      <ul class="add-other" ng-if="$ctrl.onDeployImageSelected || $ctrl.onFromFileSelected">
-        <li ng-if="$ctrl.onDeployImageSelected">
-          <a href="" ng-click="$ctrl.onDeployImageSelected()">Deploy Image</a>
-        </li>
-        <li ng-if="$ctrl.onFromFileSelected">
-          <a href="" ng-click="$ctrl.onFromFileSelected()">Import <span class="import-file-types">YAML / JSON</span></a>
-        </li>
-      </ul>
+      <div ng-if="$ctrl.onDeployImageSelected || $ctrl.onFromFileSelected">
```
jeff-phillips-18: I know its an odd case but if there is only one item should this be a dropdown?

rhamilto: That would be the ideal UX, but I'm not sure how common it'll be.  Any idea, @spadgett?

spadgett: I think in practice you will either have all 3 (running in the console) or none. @jeff-phillips-18 OK to leave as is?

jeff-phillips-18: Yeah for sure

PR #387, opened 2017-08-18
Don't select a default project if there is more than one project

dtaylor113:
The following changes have been made to the ```<select-project>``` component:

- Accepts a list of projects, so another projects retrieval is not needed (performant)
- Callback method when project selected.
- Does not select a project by default, unless there is only one project
- required asterisk, error msg if user clicks on dropdown, but doesn't select a project 
- Sorts projects by displayName

**No Project Selected**
image

**User clicks on then off dropdown, and doesn't select a project:**
image

**Project Selected**
image

Fixes:  https://github.com/openshift/origin-web-console/issues/1833

### DIFF
--- a/src/components/select-project/select-project.controller.ts
+++ b/src/components/select-project/select-project.controller.ts
@@ -49,10 +49,16 @@ export class SelectProjectController implements angular.IController {
     if (onChangesObj.nameTaken && !onChangesObj.nameTaken.isFirstChange()) {
       this.ctrl.forms.createProjectForm.name.$setValidity('nameTaken', !this.ctrl.nameTaken);
     }
+    if (onChangesObj.availableProjects && !onChangesObj.availableProjects.isFirstChange()) {
+      this.filterProjects(this.ctrl.availableProjects);
+    }
   }
 
   public onSelectProjectChange () {
     this.canIAddToProject();
+    if (angular.isFunction(this.ctrl.onProjectSelected)) {
+      this.ctrl.onProjectSelected(this.ctrl.selectedProject);
+    }
   }
 
   public onNewProjectNameChange() {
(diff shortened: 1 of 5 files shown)


spadgett:
We should have `class="required"` on the label so it gets the asterisk.

We shouldn't show the red error text when the page first loads. Only if the user focuses the project select, but doesn't select anything.

dtaylor113:
@spadgett added required and error text when no project selected after losing focus.

serenamarie125:
@spadgett curious on "Sorts projects by displayName" .  Should we enhance this to sort primary by last accessed by the user & secondary by displayName?

@dtaylor113 also curious if this is ui-select if we are including the search capability here?

dtaylor113:
> @dtaylor113 also curious if this is ui-select if we are including the search capability here?

Yup :-)
image

spadgett:
> @spadgett curious on "Sorts projects by displayName" . Should we enhance this to sort primary by last accessed by the user & secondary by displayName?

Yeah, I think so. It's also possible to group items in ui-select, so we could have "Recent" and "All" groups if we want.

spadgett:
> removed unused name-taken attribute

@dtaylor113 `name-taken` is being used 

https://github.com/openshift/origin-web-console/search?utf8=%E2%9C%93&q=name-taken&type=

dtaylor113:
> @dtaylor113 name-taken is being used

I see that the HTML is using ```projectNameTaken```, but I don't see that variable being used/set in any of the controllers.   I think it needs to be removed from the HTML.

spadgett:
To be clear, we should still use `osc-unique` so we can warn users earlier when we know a project name already exists. But we also have to handle the case where we only find out later when we try to create the project.

dtaylor113:
> We need to add name-taken back and make sure it's used properly everywhere. It's a bug if it was removed from any of the controllers.

Hi @spadgett, ok added back ```name-taken``` and filed this issue: https://github.com/openshift/origin-web-console/issues/1955

src/components/select-project/selectProject.html
```diff
@@ -1,10 +1,13 @@
 <ng-form name="$ctrl.forms.selectProjectForm">
-  <div class="form-group" ng-class="{'has-error' : $ctrl.forms.selectProjectForm.selectProject.$error.cannotAddToProject }">
-    <label class="control-label" for="project">Add to Project</label>
+  <div class="form-group" ng-class="{'has-error' : $ctrl.forms.selectProjectForm.selectProject.$error.cannotAddToProject ||
+                                                   ($ctrl.forms.selectProjectForm.selectProject.$touched &&
+                                                    $ctrl.forms.selectProjectForm.selectProject.$invalid)}">
+    <label class="control-label required" for="project">Add to Project</label>
```
dtaylor113: Ok, removed ```for="project"```

src/components/select-project/select-project.controller.ts
```diff
@@ -45,19 +43,11 @@ export class SelectProjectController implements angular.IController {
     });
   }
 
-  public $onChanges(onChangesObj: angular.IOnChangesObject) {
```
dtaylor113: Fixed

spadgett:
Thanks @dtaylor113 LGTM

PR #393, opened 2017-08-22
Update failure checks during service provisioning

jeff-phillips-18:
First part of fix for https://github.com/openshift/origin/issues/15898 to show failures in the results page of the wizard.

### DIFF
--- a/src/components/order-service/order-service.controller.ts
+++ b/src/components/order-service/order-service.controller.ts
@@ -418,7 +418,7 @@ export class OrderServiceController implements angular.IController {
       var readyCondition: any = _.find(conditions, {type: "Ready"});
 
       this.ctrl.orderComplete = readyCondition && readyCondition.status === 'True';
-      this.ctrl.error = _.find(conditions, {type: "ProvisionFailed"});
+      this.ctrl.error = _.find(conditions, {type: "Failed"});
     }));
   }
 }
--- a/dist/origin-web-catalogs.js
+++ b/dist/origin-web-catalogs.js
@@ -1171,7 +1171,7 @@ webpackJsonp([ 0, 1 ], [ function(e, t) {
                         type: "Ready"
                     });
                     l.ctrl.orderComplete = r && "True" === r.status, l.ctrl.error = i.find(n, {
-                        type: "ProvisionFailed"
+                        type: "Failed"
                     });
                 }));
             }, this.$scope = e, this.$filter = t, this.AuthService = n, this.ProjectsService = r, 


dtaylor113:
I don't know why travis is suddenly failing on these typescript issues.   I thought if a variable didn't have a ': <whatever>'  it would default to ": any".

src/components/order-service/order-service.controller.ts
```diff
@@ -418,7 +418,11 @@ export class OrderServiceController implements angular.IController {
       var readyCondition: any = _.find(conditions, {type: "Ready"});
 
       this.ctrl.orderComplete = readyCondition && readyCondition.status === 'True';
-      this.ctrl.error = _.find(conditions, {type: "ProvisionFailed"});
+      this.ctrl.error = _.find(conditions, function(condition: any) {
+        return condition.reason === "ProvisionFailed" ||
+               condition.reason === "ProvisionCallFailed" ||
+               condition.reason === 'ErrorCallingProvision';
+      });
```
spadgett: We should be checking specifically for the `Failed` condition. Something like

```js
_.some(conditions, { type: 'Failed', status: 'True' });
```

jeff-phillips-18: When I get the failure I am seeing:
```
type: "Ready"
status: "False"
reason: "ProvisionCallFailed"
```

spadgett: There should be another condition type `Failed` now. You might need to pull the latest origin-service-catalog image.

cc @pmorie to keep me honest

jeff-phillips-18: @pmorie Provisioning a service is failing with:

> "Provision call failed. Error provisioning Instance "whee/django-psql-persistent-4l3nw" of ServiceClass "django-psql-persistent" at Broker "template-broker": Service instance with same id but different attributes exists"

But the condition is still:
type: 'Ready'
status: 'False'
reason: 'ProvisionCallFailed'

Should the type got to 'Failed'?

jeff-phillips-18: I am getting the type going to 'Failed' now that I have updated to the latest origin-service-catalog.

jeff-phillips-18: @spadgett Fixed.

src/components/order-service/order-service.controller.ts
```diff
@@ -418,7 +418,7 @@ export class OrderServiceController implements angular.IController {
       var readyCondition: any = _.find(conditions, {type: "Ready"});
 
       this.ctrl.orderComplete = readyCondition && readyCondition.status === 'True';
-      this.ctrl.error = _.find(conditions, {type: "ProvisionFailed"});
+      this.ctrl.error = _.find(conditions, {type: "Failed"});
```
spadgett: Ugh missed this before I merged, but I think it needs to be

```
this.ctrl.error = _.find(conditions, {type: "Failed", status: "True"});
```

PR #394, opened 2017-08-22
Add Select from Project option to the services view.

jeff-phillips-18:


src/services/catalog.service.ts
```diff
@@ -67,6 +67,41 @@ export class CatalogService {
...
+        errorMsg.push('builder images');
+      }).finally(() => {
+        this.returnCatalogItems(deferred, catalogItems, ++numPromisesExecuted, totalNumPromises, errorMsg);
+      });
+    }
+
+    if (includeTemplates) {
+      totalNumPromises++;
+      this.dataService.list("templates", {namespace: projectName}, null, {partialObjectMetadataList: partialObjectMetadataList}).then((resources: any) => {
+        catalogItems.templates = _.values(resources._data);
```
spadgett: Hate to access `_.data` directly since it's meant to be private. Any reason not to use `resources.by('metadata.name')`?

jeff-phillips-18: Fixed

PR #399, opened 2017-08-23
Read and use form definition for parameters if supplied by broker.

cfchase:
Allows the OpenShift UI to enforce order, grouping, and display type, if that information is supplied by the broker.  Brokers are responsible for adding the form definition in the plan metadata as `schemas.service_instance.create.form_definition`  (A similar path for service_binding.create can also be supplied this way, but out of scope fort his PR).   

Unfortunately, regarding grouping, fieldset and section html is generated properly, but has no visual difference.  This would take additional work to style appropriately. (picture below)

Example broker code:
https://github.com/openshift/ansible-service-broker/pull/386 

Example json gist:
https://gist.github.com/cfchase/8614236dcb8e22cc0aa8a2b4f1ee52d0

Screenshots
Full Wizard:
wizard-full

Example display types of fields:
wizard-fields-1
wizard-fields-2
wizard-fields-3

Fieldset not styled:
wizard-fieldset-no-style

### DIFF
--- a/src/components/catalog-parameters/catalog-parameters.controller.ts
+++ b/src/components/catalog-parameters/catalog-parameters.controller.ts
@@ -1,11 +1,22 @@
+import * as _ from 'lodash';
+
 export class CatalogParametersController implements angular.IController {
+
+  private static ALLOWED_FORM_INPUT_TYPES = {
+    'fieldset': true,
+    'text': true,
+    'textarea': true,
+    'password': true,
+    'checkbox': true,
+    'select': true
+  };
 
   public ctrl: any = this;
 
   public $onInit() {
     // https://github.com/json-schema-form/angular-schema-form/blob/development/docs/index.md
-    // Show all fields in the schema.
-    this.ctrl.parameterForm = ['*'];
+    // If no form definition is supplied, show all fields in the schema using '*'.
+    this.ctrl.parameterForm = this.cloneParameterForm(this.ctrl.parameterFormDefinition) || ['*'];
     this.ctrl.parameterFormDefaults = {
       formDefaults: {
         // Add Bootstrap horizontal form styles.
@@ -28,4 +39,43 @@ export class CatalogParametersController implements angular.IController {
       }
     };
   }
+
+  // clones a form definition with only the accepted keys (key, type, items)
+  // and type values (fieldset, text, textarea, password, checkbox, select)
+  private cloneParameterForm(original: any) {
+
+    if (_.isString(original)) {
+      return original;
+    }
+
+    if (_.isArray(original)) {
+      return _.map(original, _.bind(this.cloneParameterForm, this));
+    }
+
+    if (_.isObject(original)) {
+      let newObject: any = {};
+
+      if (original.key) {
+        newObject.key = original.key;
+      }
+
+      if (CatalogParametersController.ALLOWED_FORM_INPUT_TYPES[original.type]) {
+        newObject.type = original.type;
+      }
+
+      if (newObject.type === 'fieldset' && _.isArray(original.items)) {
+        if (original.title) {
+          newObject.title = original.title;
+        }
+        newObject.items = this.cloneParameterForm(original.items);
+      }
+
+      //if empty/invalid, pass back null, not a truthy object
+      if (!newObject.key && !newObject.type) {
+        return null;
+      }
+
+      return newObject;
+    }
+  };
 }
--- a/src/components/order-service/order-service.controller.ts
+++ b/src/components/order-service/order-service.controller.ts
@@ -323,6 +323,7 @@ export class OrderServiceController implements angular.IController {
       this.sendRequesterUsername = false;
     }
     this.ctrl.parameterSchema = schema;
+    this.ctrl.parameterFormDefinition = _.get(this, 'ctrl.selectedPlan.externalMetadata.schemas.service_instance.create.openshift_form_definition');
   }
 
   private onProjectUpdate = () => {
--- a/src/styles/order-service.less
+++ b/src/styles/order-service.less
@@ -203,4 +203,16 @@
     min-height: 200px;
     overflow-y: auto;
   }
+
+  .schema-form-fieldset {
+    margin-bottom: 40px;
+    margin-top: 40px;
+
+    legend {
+      border-bottom: none;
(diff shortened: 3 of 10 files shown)


cfchase:
[(8) Parameter Ordering and Grouping [8]](https://trello.com/c/Orp4rGpI/441-8-parameter-ordering-and-grouping-8)

src/components/order-service/order-service.controller.ts
```diff
@@ -312,6 +313,34 @@ export class OrderServiceController implements angular.IController {
     this.ctrl.parameterSchema = schema;
   }
 
+  private updateParameterForm() {
+    let formDefinition: any = _.get(this, 'ctrl.selectedPlan.externalMetadata.schemas.service_instance.create.form_definition');
+    this.removeUnsupportedParameterFormItems(formDefinition);
+    this.ctrl.parameterForm = formDefinition || ['*'];
+  }
+
+  private removeUnsupportedParameterFormItems(formItems: any) {
```
spadgett: It would be better to move this in the catalog-parameters component so that we will be able to reuse it for bind parameters.

src/components/order-service/order-service.controller.ts
```diff
@@ -312,6 +313,34 @@ export class OrderServiceController implements angular.IController {
...
+      'key': true,
+      'type': true,
+      'items': true
+    };
+
+    _.forEach(formItems, (item: any) => {
+      if (_.isObject(item)) {
+        _.forOwn(item, (value, key) => {
+          if (!allowedKeys[key]) {
+            delete item[key];
```
spadgett: We should avoid mutating the original API object since it's cached by our DataService and possibly used outside of this controller. We should only modify a copy.

src/components/order-service/order-service.controller.ts
```diff
@@ -312,6 +313,34 @@ export class OrderServiceController implements angular.IController {
...
+  }
+
+  private removeUnsupportedParameterFormItems(formItems: any) {
+    const allowedKeys = {
+      'key': true,
+      'type': true,
+      'items': true
+    };
+
+    _.forEach(formItems, (item: any) => {
```
spadgett: Nit: we've been trying to consistently use `_.each` so we don't have a mix

src/components/order-service/order-service.controller.ts
```diff
@@ -312,6 +313,34 @@ export class OrderServiceController implements angular.IController {
...
+  private updateParameterForm() {
+    let formDefinition: any = _.get(this, 'ctrl.selectedPlan.externalMetadata.schemas.service_instance.create.form_definition');
+    this.removeUnsupportedParameterFormItems(formDefinition);
+    this.ctrl.parameterForm = formDefinition || ['*'];
+  }
+
+  private removeUnsupportedParameterFormItems(formItems: any) {
+    const allowedKeys = {
+      'key': true,
+      'type': true,
```
spadgett: I'd like to lock down the possible values for type as well. I think this would be a good set:

- fieldset
- text
- textarea
- password
- checkbox
- select

https://github.com/json-schema-form/angular-schema-form/blob/development/docs/index.md#form-types

We might need to move away from angular-schema-form someday, and this list seems manageable if we ever need to implement it ourselves.

src/components/order-service/order-service.controller.ts
```diff
@@ -312,6 +313,34 @@ export class OrderServiceController implements angular.IController {
     this.ctrl.parameterSchema = schema;
   }
 
+  private updateParameterForm() {
+    let formDefinition: any = _.get(this, 'ctrl.selectedPlan.externalMetadata.schemas.service_instance.create.form_definition');
```
spadgett: Let's add a prefix to avoid ever colliding with a future field added to OSB API. I'd suggest `openshift_form_definition`. That also makes it clear to anyone looking at the broker that the field is specific to the OpenShift catalog.

@pmorie Any objections?

cfchase:
* changed field name to openshift_form_definition
* added comment back in with slight change
* changed deletion if bad keys to copying of object only with good keys and type values
* use _.each instead of _.forEach

@spadgett let me know if this addresses your comments or if there's more to do.

src/components/catalog-parameters/catalog-parameters.controller.ts
```diff
@@ -28,4 +39,39 @@ export class CatalogParametersController implements angular.IController {
...
     };
   }
+
+  // clones a form definition with only the accepted keys (key, type, items)
+  // and type values (fieldset, text, textarea, password, checkbox, select)
+  private cloneParameterForm(original: any) {
+    let newValue: any;
+
+    if (_.isString(original)) {
+      newValue = original;
```
spadgett: I think early returns would make this easier to follow:

```js
if (_.isString(original)) {
  return original;
}

// ...
```

src/components/catalog-parameters/catalog-parameters.controller.ts
```diff
@@ -28,4 +39,39 @@ export class CatalogParametersController implements angular.IController {
...
+    } else if (_.isArray(original)) {
+      newValue = [];
+      _.each(original, (originalItem) => {
+        let newItem = this.cloneParameterForm(originalItem);
+        if (newItem) {
+          newValue.push(newItem);
+        }
+      });
```
spadgett: `_.map` should help here

```js
if (_.isArray(original)) {
  return _.map(original, _.bind(this.cloneParameterForm, this));
}
```

cfchase:
@spadgett let me know how this styling looks to you or if it needs less spacing, etc.

* Cleaned up cloneParameterForm
* Added fieldset titles and styling
* Added mock data service for APB.

Screenshots of grouping

wizard-full-groups

wizard-groups-1
wizard-groups-2
wizard-groups-3

src/styles/order-service.less
```diff
@@ -203,4 +203,16 @@
...
+
+  .schema-form-fieldset {
+    margin-top: 40px;
+    margin-bottom: 40px;
+
+    legend {
+      padding-top: 20px;
+      margin-bottom: 10px;
+      border-top: solid 1px rgba(0, 0, 0, 0.15);
+      border-bottom: none;
```
spadgett: nit: we try to keep the CSS styles alphabetized

PR #408, opened 2017-08-30
Update angular-patternfly dependency to "at least" syntax

benjaminapetersen:
- use `>` for angular-patternfly 
- add DS_Store to `.gitignore`

We are using `~` right now, which allows for a range in the patch.  I think we should use `>`. 

Perhaps `>=` would be better:

`"angular-patternfly": ">=4.3.0"`

 as we should be able to expect future minor changes to work as well. We can add a limit if it makes sense:

`"angular-patternfly": ">=4.3.0 <5.0.0"`

This should clarify that `openshift-web-console` is the source of truth & simplify bumping the dependencies.  It does assume `semver` is followed closely & breaking changes will not be introduced without a major version bump.

@spadgett @jeff-phillips-18 ?

Relevant for [this PR](https://github.com/openshift/origin-web-console/pull/1988)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,4 +12,4 @@ dist/*.svg
 dist/*.gif
 dist/*.jpg
 dist/*.png
-
+.DS_Store
--- a/bower.json
+++ b/bower.json
@@ -35,7 +35,7 @@
     "angular-sanitize": "1.5.11",
     "angular-route": "1.5.11",
     "angular-bootstrap": "0.14.3",
-    "angular-patternfly": "~4.7.1",
+    "angular-patternfly": ">=4.7.1 <5.0.0",
     "angular-schema-form": "^0.8.13",
     "patternfly": "~3.26.1",
     "jquery": "~3.2.1",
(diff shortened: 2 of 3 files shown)


spadgett:
We should at least down down to version 4.x. We don't have any idea if this will work with new major versions.

benjaminapetersen:
Just changed it to: `"angular-patternfly": ">=4.7.1 <5.0.0",`

spadgett:
Need to update bower.json as well

benjaminapetersen:
Curious why its in both `package.json` and `bower.json`?  Updating now.

spadgett:
So that downstream projects can use either npm or bower

benjaminapetersen:
I wonder if we should update some of the other dependencies like this as well.  `origin-web-common` is probably a good candidate.   I tend to trust `jQuery` and `lodash` not to break unexpectedly (per above, just passing the real responsibility to specify exact versions to `origin-web-console`).

spadgett:
origin-web-common isn't a problem until we cut a >=0.1.0 release. I think for now it's OK and we can reevaluate later.

origin-web-common changes often need to happen in step with catalog changes, so it might be better to lock versions honestly.

benjaminapetersen:
Ok works for me.

This one is needed to resolve the issue bumping `angular-patternfly` for fixing `track by` in the `web-console` notification drawer PR.  https://github.com/openshift/origin-web-console/pull/1988

PR #409, opened 2017-08-30
Support bind parameters

spadgett:
https://trello.com/c/Sws4p5jU

Requires https://github.com/openshift/origin-web-common/pull/163
See also https://github.com/openshift/origin-web-console/pull/1987

@jeff-phillips-18 FYI

### DIFF
--- a/src/components/order-service/order-service.controller.ts
+++ b/src/components/order-service/order-service.controller.ts
@@ -20,8 +20,10 @@ export class OrderServiceController implements angular.IController {
   private planStep: any;
   private configStep: any;
   private bindStep: any;
+  private bindParametersStep: any;
   private reviewStep: any;
   private selectedProjectWatch: any;
+  private bindTypeWatch: any;
   private deploymentConfigs: any;
   private deployments: any;
   private replicationControllers: any;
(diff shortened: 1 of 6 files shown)


src/components/order-service/order-service.controller.ts
```diff
@@ -313,6 +351,14 @@ export class OrderServiceController implements angular.IController {
     }
   }
 
+  private updateBindParametersStepVisibility() {
+    // Show the bind parameters step if the bind step if not hidden and the plan has a bind parameter schema.
+    this.bindParametersStep.hidden = this.bindStep.hidden ||
+                                     this.ctrl.bindType === 'none' ||
+                                     !_.has(this.ctrl, 'bindParameterSchema.properties');
+    this.bindParametersStep.allowed = !this.bindParametersStep.hidden && this.bindStep.valid;
```
jeff-phillips-18: nit: don't really need to check hidden here, if it is hidden then allowed will never be used.

src/components/order-service/order-service.controller.ts
```diff
@@ -174,7 +196,17 @@ export class OrderServiceController implements angular.IController {
 
     this.validityWatcher = this.$scope.$watch("$ctrl.forms.bindForm.$valid", (isValid: any, lastValue: any) => {
       this.bindStep.valid = isValid;
-      this.reviewStep.allowed = this.bindStep.valid;
+      this.bindParametersStep.allowed = !this.bindParametersStep.hidden && isValid;
```
jeff-phillips-18: nit: don't really need to check hidden here, if it is hidden then allowed will never be used.

spadgett:
Needs a common bump :(

spadgett:
Sorry I shouldn't have removed WIP without the common release

PR #382, opened 2017-08-16
Catalog Filter Panel

dtaylor113:
https://trello.com/c/UWgrj6bL

Hi, here is what I have so far.  Please see list of open issues at bottom.

**Catalog Filter Panel:**

image

**Vendor Warning** (needs wordsmithing, warning icon replaced with info icon :-)

image

## Open Issues

1. Is the Vendor alert icon and tooltip ok?

2. If none of the catalog items have vendor information, then we just show the Keyword input filter in the panel:
image
Is it worth having a filter dropdown with just one input field?  In this scenario, it seems like the input field should be visible on the same lines as the "15 Items".

3. I notice in the mock data we have for service-classes 'metadata' and 'externalMetadata', currently I'm putting vendor info. in 'metadata.providerDisplayName' just want to verify this is correct.

Thanks,
- Dave

PR #410, opened 2017-08-31
Look for providerDisplayName in externalMetadata

spadgett:
We're looking for `providerDisplayName` in the wrong place, which is why the faceted filter for Vendor is not showing up.

### DIFF
--- a/src/services/catalog.service.ts
+++ b/src/services/catalog.service.ts
@@ -363,7 +363,7 @@ export class ServiceItem implements IServiceItem {
   }
 
   private getVendor(): string {
-    return _.get(this.resource, 'metadata.providerDisplayName') as string || '';
+    return _.get(this.resource, 'externalMetadata.providerDisplayName') as string || '';
   }
 }
 
@@ -439,7 +439,7 @@ export class ImageItem implements IServiceItem {
   }
 
   private getVendor(): string {
-    return _.get(this.resource, 'metadata.providerDisplayName') as string || '';
+    return '';
   }
 
   private getDescription() {
@@ -503,6 +503,6 @@ export class TemplateItem implements IServiceItem {
   }
 
   private getVendor(): string {
-    return _.get(this.resource, 'metadata.providerDisplayName') as string || '';
+    return _.get(this.resource, ['metadata', 'annotations', 'template.openshift.io/provider-display-name']) as string || '';
   }
 }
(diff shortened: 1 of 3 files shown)


src/services/catalog.service.ts
```diff
@@ -439,7 +439,7 @@ export class ImageItem implements IServiceItem {
   }
 
   private getVendor(): string {
-    return _.get(this.resource, 'metadata.providerDisplayName') as string || '';
+    return '';
```
jeff-phillips-18: Should we just remove the vendor field from the ImageItem class?

spadgett: It's in the `IServiceItem` interface as a public field, although I could remove the method and just initialize it to `''`

jeff-phillips-18: I see. It's OK as is then. Maybe someday it has one.

spadgett:
> Funny, this was a question/open issue I raised in the Catalog Filter PR #382 :-) LGTM

My fault, I missed that. Sorry :(

src/services/catalog.service.ts
spadgett: @bparees Any opinion on adding the `template.openshift.io/provider-display-name` annotation to builder images? Without it, all of the builders will disappear in the UI when you filter for Provider: Red Hat

bparees: you mean to the imagestreamtags?

bparees: i'd prefer we come up with an annotation that's not tied to templates, for setting on things that aren't templates.

bparees: you're also going to have to get all the xpaas imagestreamtags updated w/ whatever annotation.

spadgett: @bparees The image stream shows up as one item in the catalog, so it'd probably need to be on the image stream itself rather than the image stream tags if we want the filter to work.

Agree about the annotation name.

spadgett: > you're also going to have to get all the xpaas imagestreamtags updated w/ whatever annotation.

@bparees Yeah, @sspeiche and I were just talking about the xpaas image streams. I think we could get the annotation added.

spadgett:
We can do a follow on if we add an image stream annotation since we're not sure what it would be.

src/services/catalog.service.ts
bparees: I'm open to adding it.

PR #414, opened 2017-09-01
Do not allow binding to applications if pod_presets is false.

jeff-phillips-18:
Also, do not show binding in progress until the service instance has been created.

Depends on origin-web-common release with https://github.com/openshift/origin-web-common/pull/167


jeff-phillips-18:
Bumped version of origin-web-common. Removed WIP status.

@spadgett @dtaylor113

PR #416, opened 2017-09-06
Update Middleware categories

spadgett:
openshift web console 2017-09-06 08-55-50

### DIFF
--- a/dist/origin-web-catalogs.js
+++ b/dist/origin-web-catalogs.js
@@ -293,25 +293,50 @@ webpackJsonp([ 0, 1 ], [ function(e, t) {
             id: "middleware",
             label: "Middleware",
             subCategories: [ {
-                id: "jboss",
-                label: "JBoss",
-                tags: [ "jboss" ],
-                icon: "font-icon icon-openjdk"
-            }, {
-                id: "fuse",
-                label: "Fuse",
-                tags: [ "fuse" ],
-                icon: "font-icon icon-openjdk"
-            }, {
                 id: "amq",
                 label: "A-MQ",
                 tags: [ "amq" ],
-                icon: "font-icon icon-openjdk"
+                icon: "font-icon icon-jboss"
             }, {
-                id: "bpm",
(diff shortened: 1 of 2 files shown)


dist/origin-web-catalogs.js
```diff
@@ -293,25 +293,50 @@ webpackJsonp([ 0, 1 ], [ function(e, t) {
...
+                icon: "font-icon icon-jboss"
+            }, {
+                id: "processserver",
+                label: "BPM Suite",
+                tags: [ "processserver" ],
+                icon: "font-icon icon-jboss"
+            }, {
+                id: "decisionserver",
+                label: "BRMS",
+                tags: [ "processserver" ],
```
knrc: decisionserver

spadgett:
@rhamilto Need to decide if we want a max-width for long subcategory names

spadgett:
@sspeiche @serenamarie125

dist/origin-web-catalogs.js
```diff
@@ -293,25 +293,50 @@ webpackJsonp([ 0, 1 ], [ function(e, t) {
...
+                tags: [ "decisionserver" ],
+                icon: "font-icon icon-jboss"
+            }, {
+                id: "datagrid",
+                label: "Data Grid",
+                tags: [ "datagrid" ],
+                icon: "font-icon icon-jboss"
+            }, {
+                id: "datavirt",
+                label: "Data Virtualization",
```
spadgett: @knrc Is there a shorter label that we could use here? "Data Virt" maybe? This makes the subcategory tab larger than others.

knrc:
On 6 September 2017 at 06:57, Sam Padgett <notifications@github.com> wrote:

> @knrc <https://github.com/knrc> Is there a shorter label that we could
> use here? "Data Virt" maybe? This makes the subcategory tab larger than
> others.
>
I would have no objections to this, make the change and if someone objects
they can raise it as an issue :)

   Kev

PR #424, opened 2017-09-08
Use new ApplicationsService.getApplications utility

jeff-phillips-18:
Requires https://github.com/openshift/origin-web-common/pull/178


jeff-phillips-18:
Removed WIP status

PR #434, opened 2017-09-14
Publisher Synonyms

### LINKED ISSUE #429: Let admins provide synonyms for the publisher filter
I could see values like "Red Hat", "Red Hat, Inc", and "Red Hat, Inc." (with a period) all showing up in the filter panel. Since the OSB API metadata value is `providerDisplayName` rather than a unique identifier, brokers could refer to the same provider in slightly different ways.

We might add a way in constants.js to fix this. For instance, a map of synonyms like

```js
OPENSHIFT_CONSTANTS.PUBLISHER_SYNONYMS = {
  'Red Hat': 'Red Hat, Inc.'
  'Red Hat, Inc': 'Red Hat, Inc.'
};
```

then all of these values would be treated as "Red Hat, Inc." in the filter panel.

cc @serenamarie125 @ncameronbritt @jeff-phillips-18

dtaylor113:
Closes #429 

image

### DIFF
--- a/app/mockServices/mockConstants.service.ts
+++ b/app/mockServices/mockConstants.service.ts
@@ -4,6 +4,14 @@ export class ConstantsService {
   public constants: any;
   constructor () {
     let win: any = window;
+
+    let publisherSynonyms = {
+      'Vendor B': 'Vendor B, Inc.',
+      'Vendor B, Inc': 'Vendor B, Inc.',
+      'Vendor B, Co.': 'Vendor B, Inc.'
+    };
+    _.set(win, 'OPENSHIFT_CONSTANTS.PUBLISHER_SYNONYMS', publisherSynonyms);
+
     return _.clone(win.OPENSHIFT_CONSTANTS || {});
   }
 }
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -161,3 +161,11 @@ var guidedTours = {
 };
 
 _.set(window, 'OPENSHIFT_CONSTANTS.GUIDED_TOURS', guidedTours);
+
+// Example publisher synonyms to normilize vendor names shown in publisher filter in Browse Catalog
+_.set(window, 'OPENSHIFT_CONSTANTS.PUBLISHER_SYNONYMS',
+  {
+    // 'Red Hat': 'Red Hat, Inc.'
+    // 'Red Hat, Inc': 'Red Hat, Inc.'
+  }
+);
--- a/src/services/catalog.service.ts
+++ b/src/services/catalog.service.ts
@@ -157,6 +157,10 @@ export class CatalogService {
     return new TemplateItem(resource, this);
   }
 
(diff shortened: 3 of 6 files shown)


src/services/catalog.service.ts
```diff
@@ -157,6 +157,10 @@ export class CatalogService {
     return new TemplateItem(resource, this);
   }
 
+  public getPublisherSynonym(rawVendor: string): string {
+    return (this.constants.PUBLISHER_SYNONYMS ? (_.get(this.constants.PUBLISHER_SYNONYMS, rawVendor) as string) || rawVendor : rawVendor);
```
spadgett: You have to pass `rawVendor` to `_.get` in an array, otherwise special characters like `.` in `rawVendor` cause problems.

```ts
return (this.constants.PUBLISHER_SYNONYMS ? (_.get(this.constants.PUBLISHER_SYNONYMS, [rawVendor]) as string) || rawVendor : rawVendor);
```

But this is more concise and does the same thing

```ts
return _.get(this.constants, ['PUBLISHER_SYNONYMS', rawVendor]) as string || rawVendor;
```

src/constants.ts
```diff
@@ -161,3 +161,12 @@ var guidedTours = {
 };
 
 _.set(window, 'OPENSHIFT_CONSTANTS.GUIDED_TOURS', guidedTours);
+
+var publisherSynonyms = {
+  'Vendor B': 'Vendor B, Inc.',
+  'Vendor B, Inc': 'Vendor B, Inc.',
+  'Vendor B, Co.': 'Vendor B, Inc.'
+};
```
spadgett: Let's not add test data to constants.ts because these will get set as real values in the actual catalog.

dtaylor113: Moved to mockServices

src/services/catalog.service.ts
dtaylor113: Thanks, fixed

app/mockServices/mockConstants.service.ts
```diff
@@ -4,6 +4,14 @@ export class ConstantsService {
   public constants: any;
   constructor () {
     let win: any = window;
+
+    let publisherSynonyms = {
+      'Vendor B': 'Vendor B, Inc.',
+      'Vendor B, Inc': 'Vendor B, Inc.',
+      'Vendor B, Co.': 'Vendor B, Inc.'
+    };
+    _.set(win, 'OPENSHIFT_CONSTANTS.PUBLISHER_SYNONYMS', publisherSynonyms);
```
spadgett: Do you mind adding an empty value to constants.ts with a comment explaining what it is. We might even add the Red Hat values from #429 

It's nice to keep the constants files well commented since it's open source so that people can see all the options together.

dtaylor113: Done

spadgett:
Closing and reopening to kick Travis tests

PR #435, opened 2017-09-14
Hitting Enter key on Catalog Search triggers 'View All Results'

dtaylor113:
Fixes [console#1949](https://github.com/openshift/origin-web-console/issues/1949)


src/components/catalog-search/catalog-search.controller.ts
```diff
@@ -42,6 +42,12 @@ export class CatalogSearchController implements angular.IController {
     }
   }
 
+  public onKeyPress = (keyEvent: any) => {
+    if (keyEvent.which === 13) {
+      this.$rootScope.$emit('filter-catalog-items', {searchText: this.ctrl.searchText});
```
spadgett: What happens if there is no search text?

dtaylor113: Nothing :-)  No filter is applied in the service-view/browse catalog

spadgett: If I had a previous search, is it cleared?

I think I'd rather check 

```js
if (keyEvent.which === 13 && this.ctrl.searchText)
```

here to avoid event firing the event. Then we're not relying on any specific service view search behavior.

dtaylor113: > If I had a previous search, is it cleared?

No, it remains
I'll add the other condition, thanks

dtaylor113: ok, updated

spadgett: great, thanks @dtaylor113 !

PR #436, opened 2017-09-15
Update catalog-parameters form definition when changed.

cfchase:
When changing plans, the form definition wasn't being updated.  As a result, only the first plan's form definition was shown.  This change updates the displayed form definition whenever the input form definition is changed (such as by changing plans)


cfchase:
[(8) Parameter Ordering and Grouping](https://trello.com/c/Orp4rGpI/441-8-parameter-ordering-and-grouping)

PR #413, opened 2017-09-01
Changing overlay layout to a single column

rhamilto:
localhost-9000-dev-console-
localhost-9000-dev-console- iphone 6

PR #438, opened 2017-09-18
Fixing catalog wizard

rhamilto:
Supports https://github.com/openshift/origin-web-console/issues/2087

Changes to PatternFly's wizard resulted in no content appearing in our catalog overlay modals.  This PR resolves that issue.  Given https://github.com/openshift/origin-web-catalog/pull/413 is to change the layout of the wizard to a single column at all times, I intend to further refine the display and possibly eliminate the use of 100vh because of [issues with mobile browsers' interpretation of vh units](https://adactio.com/journal/11690) as part of #413.


bower.json
```diff
@@ -37,7 +37,7 @@
     "angular-bootstrap": "0.14.3",
     "angular-patternfly": ">=4.7.1 <5.0.0",
     "angular-schema-form": "^0.8.13",
-    "patternfly": "~3.26.1",
+    "patternfly": ">=3.27.2",
```
spadgett: Maybe

```json
"patternfly": ">=3.27.2 <4.0.0"
```

bower.json
```diff
@@ -47,7 +47,6 @@
     "ace-builds": "1.2.2",
     "angular-extension-registry": "1.2.6",
     "angular-inview": "1.5.7",
-    "angular-key-value-editor": "3.0.0",
```
spadgett: Good catch

PR #440, opened 2017-09-18
Workaround for webpack-dev-server issue with Safari

rhamilto:
https://github.com/webpack/webpack-dev-server/issues/1090 includes details of the webpack-dev-server issue.  

**Important: in order to make use of this fix, add:**

```
window.DEV_TOOL_SETTING = 'inline-source-map';
```

to your `app/config.local.js` file when you need to test in Safari.  Note this will significantly slow down build and rebuild, so use it only when testing in Safari.

Thanks, @jeff-phillips-18, for the assist on this one.

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -5,6 +5,7 @@
 // To change configuration for local development, copy this file to
 // assets/app/config.local.js and edit the copy.
 window.DEV_SERVER_PORT = 9001;
+window.DEV_TOOL_SETTING = 'cheap-eval-source-map';
 
 window.OPENSHIFT_CONFIG = {
   apis: {
--- a/config/webpack.dev.js
+++ b/config/webpack.dev.js
@@ -14,9 +14,10 @@ try {
 
 var useHTTPS = !window.MOCK_ORIGIN_SERVICES;
 var port = window.DEV_SERVER_PORT || 9001;
+var devtoolMap = window.DEV_TOOL_SETTING || 'cheap-eval-source-map';
 
 module.exports = webpackMerge(commonConfig, {
-  devtool: 'cheap-eval-source-map',
+  devtool: devtoolMap,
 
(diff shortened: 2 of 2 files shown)


jeff-phillips-18:
Ugh... this makes development painful...

rhamilto:
> Ugh... this makes development painful...

Maybe we opt for no use of devtool until the bug is resolved?

spadgett:
@jeff-phillips-18 Is it possible to set this as a command-line parameter on `npm run start`? Then we can leave this as the default, but change it as needed for development.

jeff-phillips-18:
I will play around with that a bit to see if we can do something there. I think this is OK for now to get past the current issue. If I find a way around it, I can open another PR.

jeff-phillips-18:
Actually, can we use a setting in the config.js to determine this? Then it can be overridden by config.local.js?

spadgett:
That sounds OK to me

jeff-phillips-18:
@rhamilto Did you want to add the changes to use config.js to allow overriding this? I sent you a PR for your branch so you could see what I mean.

rhamilto:
> @rhamilto Did you want to add the changes to use config.js to allow overriding this? I sent you a PR for your branch so you could see what I mean.

Yeah.  Makes sense.

I dislike that Github doesn't send you a notification when someone opens a PR to your fork.

Jay-PC:
After working through this and a related issue, and not wanting to rely on Safari 11 upgrades, I found a different work-around... in the project folder, i forced the local CLI up to the latest version...
 
`npm install --save-dev @angular/cli@latest`

my app ran fine in Safari 10.x after that. Granted, some of my npm dependencies will now probably need to be updated, but fortunately all of them appeared to work even with some warnings upon ng serve.

PR #441, opened 2017-09-18
Prevent navigating to review step by clicking number

spadgett:
* Prevent users from accidentally creating something by clicking one of the numbered steps.
* Fix a bug in the builder dialog where the button label incorrectly said "Next >" instead of "Create" when pod presets are disabled

See https://github.com/openshift/origin-web-console/issues/1829


spadgett:
cc @ncameronbritt

PR #444, opened 2017-09-19
Remove special handling of requester-username parameter

spadgett:
Requires https://github.com/openshift/origin/pull/16434

The template broker no longer needs the special `template.openshift.io/requester-username` parameter now that service-catalog 0.0.20 sends the originating identity header. Remove special handling for `requester-username`.


spadgett:
@jeff-phillips-18 PTAL, this change can probably go in soon

spadgett:
rebased, origin changes have merged

PR #455, opened 2017-09-29
Generate Generic and GitHub webhooks for builds

spadgett:
cc @ncameronbritt @rhamilto


spadgett:
@jeff-phillips-18 PTAL

spadgett:
Thanks @jeff-phillips-18

PR #458, opened 2017-10-02
Handle large project lists in select-project component

spadgett:
* Make the user start typing a name and debounce searches in the select project dropdown when there are at least 500 projects.
* Add a `MOCK_LARGE_PROJECT_LIST` option for testing large project lists.

Small project lists have no delay and work the same as before:

screencast 2017-10-02 09-16-42

Lists with 5000 projects are still pretty snappy, but the user must type to filter:

screencast 2017-10-02 09-18-12


spadgett:
@jeff-phillips-18 PTAL. There's probably more that we can do here, but this takes the component from "completely unusable" to "works pretty well" when there are a lot of projects.

spadgett:
See https://github.com/openshift/origin-web-console/issues/1923

PR #459, opened 2017-10-02
In ordering panel, if user selects project which they have 'view only' access, show error

### LINKED ISSUE #318: Project select in ordering panel shouldn't show projects where I am only a viewer

dtaylor113:
Issue was that the 'canIAddToProject' rule was expected to be loaded via the project 'rules' via a 'get'.  Fixed so that project rules are loaded if they are not in the AuthorizationService project rules cache.

Screenshots of launching the Ordering Panel without ever going to a View Project screen:

image

image

Fixes #318

### DIFF
--- a/src/components/select-project/select-project.controller.ts
+++ b/src/components/select-project/select-project.controller.ts
@@ -158,18 +158,19 @@ export class SelectProjectController implements angular.IController {
     return this.KeywordService.filterForKeywords(candidates, searchFields, keywords);
   };
 
-  private canIAddToProject(): boolean {
+  private canIAddToProject() {
     let canIAddToProject: boolean = true;
 
-    // TODO AuthorizationService.canIAddToProject assumes that the ProjectsService.get has already completed performed
-    // but currently this happens after project change events fire and it happens in the various dialog controllers, not
-    // as part of this component.  For now always return true until we can resolve this.
-    //
-    // if (!this.isNewProject()) {
-    //   canIAddToProject = this.AuthorizationService.canIAddToProject(_.get(this.ctrl.selectedProject, 'metadata.name'));
-    // }
+    var projectName = _.get(this.ctrl.selectedProject, 'metadata.name');
 
-    return this.ctrl.forms.selectProjectForm.selectProject.$setValidity('cannotAddToProject', canIAddToProject);
+    if (!this.isNewProject()) {
+      this.AuthorizationService.getProjectRules(projectName).then( () => {
+        canIAddToProject = this.AuthorizationService.canIAddToProject(projectName);
+        this.ctrl.forms.selectProjectForm.selectProject.$setValidity('cannotAddToProject', canIAddToProject);
+      });
+    }
+
+    this.ctrl.forms.selectProjectForm.selectProject.$setValidity('cannotAddToProject', canIAddToProject);
   }
 
   private updateProjects(projects: any) {
@@ -219,8 +220,6 @@ export class SelectProjectController implements angular.IController {
         this.onSelectProjectChange();
       }
     }
-
-    this.canIAddToProject();
   }
 
   private listProjects() {
--- a/dist/origin-web-catalogs.js
+++ b/dist/origin-web-catalogs.js
@@ -1787,7 +1787,10 @@ webpackJsonp([ 0, 1 ], [ function(e, t) {
             var r = [ "metadata.name", 'metadata.annotations["openshift.io/display-name"]' ], n = this.KeywordService.generateKeywords(e);
             return this.KeywordService.filterForKeywords(t, r, n);
         }, e.prototype.canIAddToProject = function() {
-            return this.ctrl.forms.selectProjectForm.selectProject.$setValidity("cannotAddToProject", !0);
+            var e = this, t = !0, r = i.get(this.ctrl.selectedProject, "metadata.name");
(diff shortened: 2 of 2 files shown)


src/components/select-project/select-project.controller.ts
```diff
@@ -83,13 +83,18 @@ export class SelectProjectController implements angular.IController {
...
+    if (!this.isNewProject()) {
+      if (_.isUndefined(this.AuthorizationService.getRulesForProject(projectName))) {
+        this.AuthorizationService.getProjectRules(projectName).then( () => {
+          canIAddToProject = this.AuthorizationService.canIAddToProject(projectName);
+          return this.ctrl.forms.selectProjectForm.selectProject.$setValidity('cannotAddToProject', canIAddToProject);
+        });
+      } else {
+        canIAddToProject = this.AuthorizationService.canIAddToProject(projectName);
+      }
+    }
```
spadgett: I believe this can just be:

```ts
if (!this.isNewProject()) {
  this.AuthorizationService.getProjectRules(projectName).then( () => {
    canIAddToProject = this.AuthorizationService.canIAddToProject(projectName);
    this.ctrl.forms.selectProjectForm.selectProject.$setValidity('cannotAddToProject', canIAddToProject);
  });
}
```

It will resolve immediately without requesting the rules again if they're already cached.

src/components/select-project/select-project.controller.ts
```diff
@@ -83,13 +83,18 @@ export class SelectProjectController implements angular.IController {
...
-    // if (!this.isNewProject()) {
-    //   canIAddToProject = this.AuthorizationService.canIAddToProject(_.get(this.ctrl.selectedProject, 'metadata.name'));
-    // }
+    var projectName = _.get(this.ctrl.selectedProject, 'metadata.name');
+
+    if (!this.isNewProject()) {
+      if (_.isUndefined(this.AuthorizationService.getRulesForProject(projectName))) {
+        this.AuthorizationService.getProjectRules(projectName).then( () => {
+          canIAddToProject = this.AuthorizationService.canIAddToProject(projectName);
+          return this.ctrl.forms.selectProjectForm.selectProject.$setValidity('cannotAddToProject', canIAddToProject);
```
spadgett: I don't think the error gets cleared if I select "Create Project" after selecting a readonly project.

src/components/select-project/select-project.controller.ts
dtaylor113: Ok, should I always call `AuthorizationService.getProjectRules(projectName, true)' to always update the cache?

src/components/select-project/select-project.controller.ts
dtaylor113: It does get cleared, `canIAddToProject` is always initially set to true.

src/components/select-project/select-project.controller.ts
spadgett: > Ok, should I always call AuthorizationService.getProjectRules(projectName, true) to always update the cache?

No, I wouldn't. I don't see a reason not to use the cached results. It doesn't change often, and the user can refresh the page if needed. I believe it's a fairly expensive request.

dtaylor113: Yes, that worked.

src/components/select-project/select-project.controller.ts
spadgett: Maybe I'm missing it, but I don't see where `$setValidity` is called again to clear the error even if `canIAddToProject` is set to true.

dtaylor113: The last line of the function, the return, will be called for a new project.

spadgett: Yup I see now, thanks

src/components/select-project/select-project.controller.ts
```diff
@@ -161,13 +161,14 @@ export class SelectProjectController implements angular.IController {
...
-    //
-    // if (!this.isNewProject()) {
-    //   canIAddToProject = this.AuthorizationService.canIAddToProject(_.get(this.ctrl.selectedProject, 'metadata.name'));
-    // }
+    var projectName = _.get(this.ctrl.selectedProject, 'metadata.name');
+
+    if (!this.isNewProject()) {
+      this.AuthorizationService.getProjectRules(projectName).then( () => {
+        canIAddToProject = this.AuthorizationService.canIAddToProject(projectName);
+        return this.ctrl.forms.selectProjectForm.selectProject.$setValidity('cannotAddToProject', canIAddToProject);
```
spadgett: This value is not actually returned to anyone. If you need `canIAddToProject()` to always return the result, it needs to return a promise that's resolved. I don't see anyone actually using the return value from `canIAddToProject()`, however, so I'd just remove the return here and below.

Note that this line also changes the return value for anyone else sharing this promise from AuthorizationService!

dtaylor113: Ok, done, thanks!

PR #448, opened 2017-09-20
Adopt service catalog API changes

spadgett:
https://trello.com/c/y1lbFrbp

Requires https://github.com/openshift/origin-web-common/pull/199
Bumps origin-web-common 0.0.63 -> 0.0.64

@jeff-phillips-18 FYI

PR #469, opened 2017-10-05
Set form definition for binding parameters.

cfchase:
Form definition for bind parameters were not being set, rendering input fields out of order set by brokers.


cfchase:
[(8) Bind Parameters](https://trello.com/c/DXrXuPnQ/525-8-bind-parameters)

spadgett:
@jeff-phillips-18 FYI, we should also use the form definition when we show bind parameters

src/components/order-service/order-service.controller.ts
```diff
@@ -372,6 +372,7 @@ export class OrderServiceController implements angular.IController {
     this.ctrl.parameterSchema = _.get(plan, 'instanceCreateParameterSchema');
     this.ctrl.parameterFormDefinition = _.get(this, 'ctrl.selectedPlan.externalMetadata.schemas.service_instance.create.openshift_form_definition');
     this.ctrl.bindParameterSchema = _.get(plan, 'serviceInstanceCredentialCreateParameterSchema');
+    this.ctrl.bindParameterFormDefinition = _.get(this, 'ctrl.selectedPlan.externalMetadata.schemas.service_binding.create.openshift_form_definition');
```
jeff-phillips-18: Note: we will need to update this with the new api changes (ctrl.selectedPlan.**spec**.externalMeatadata...)

spadgett: > Note: we will need to update this with the new api changes 

Updated in #448

PR #474, opened 2017-10-09
Add update service wizard, support readonly catalog parameters.

jeff-phillips-18:
Adding an update service wizard as well as updating the catalog-parameters component to provide for read-only form and horizontal layout and the ability to hide the current values.

TODO:
- [x]  Get update service instance to work correctly, currently setting the spec.updateRequests value is not working as expected
- [x]  Test with service classes with multiple plans 
- [x]  Test with instanceUpdateParameterSchema set.


dist/less/catalogParameters.less
```diff
@@ -0,0 +1,9 @@
+.catalog-parameters {
+  input.form-control[disabled] {
+    color: @text-color;
+    cursor: default;
+    background-color: transparent;
```
rhamilto: Nit: alpha

dist/less/catalogParameters.less
```diff
@@ -0,0 +1,9 @@
+.catalog-parameters {
+  input.form-control[disabled] {
```
rhamilto: Nit:  nesting is unnecessary since there is only one rule

src/components/update-service/update-service-review.html
```diff
@@ -0,0 +1,46 @@
...
+      </div>
+    </div>
+  </div>
+  <div class="results-failure" ng-if="$ctrl.error">
+    <div class="results-status">
+      <span class="pficon pficon-error-circle-o text-danger" aria-hidden="true"></span>
+      <span class="sr-only">Error</span>
+      <div class="results-message">
+        <h3>
+          Failed to updaate <strong>{{$ctrl.serviceName}}</strong> failed to provision in <strong>{{$ctrl.project | displayName}}</strong>.
```
rhamilto: Typo: "updaate"

src/components/update-service/update-service.controller.ts
```diff
@@ -0,0 +1,338 @@
...
+      this.DataService.create('secrets', null, secret, this.context).then(_.noop, (e: any) => {
+        this.ctrl.error = _.get(e, 'data');
+      });
+    }
+
+    // Update the instance
+    var serviceInstancesVersion: any = this.APIService.getPreferredVersion('serviceinstances');
+    // TODO: This is not working
+    updateInstance.spec.updateRequests = updateInstance.spec.updateRequests ? updateInstance.spec.updateRequests + 1 : 1;
+    this.DataService.update(serviceInstancesVersion, _.get(this.ctrl.serviceInstance, 'metadata.name'), updateInstance, this.context).then((updated: any) => {
```
spadgett: We shouldn't continue updating the instance if creating the secret failed.

src/components/update-service/update-service.controller.ts
```diff
@@ -0,0 +1,338 @@
...
+    this.context = {
+      namespace: this.ctrl.project.metadata.name
+    };
+
+    this.origParameterData = angular.copy(_.get(this.ctrl.serviceInstance, 'spec.parameters', {}));
+
+    var promises: any = [];
+
+    _.each(_.get(this.ctrl.serviceInstance, 'spec.parametersFrom'), (parametersSource) => {
+      promises.push(this.DataService.get("secrets", _.get(parametersSource, 'secretKeyRef.name'), this.context).then((secret: any) => {
```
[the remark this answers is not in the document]
jeff-phillips-18: Would there ever be? While it would do it twice, it would only override its own values.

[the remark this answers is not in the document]
spadgett: I think there can be more than on as long as the keys of the JSON objects are unique in each.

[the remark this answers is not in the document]
spadgett: Could they reference two different keys in the same secret, though?

jeff-phillips-18:
I believe I have addressed @rhamilto and @spadgett comments.

This still has not been fully tested with multiple plans and/or editable plans.

jeff-phillips-18:
Tested updating the service plan. 
Fixed issue where plans were not showing as radios. 
Fixed watching for the update success/fail condition.
Fixed title for the wizard.

jeff-phillips-18:
All Todo's completed. Ready for final reviews.

src/components/select-plan/selectPlan.html
```diff
@@ -0,0 +1,25 @@
...
+  </div>
+  <div ng-if="!$ctrl.availablePlans.length" class="blank-slate-pf">
+    <div class="blank-slate-pf-icon">
+      <span class="pficon pficon-info"></span>
+    </div>
+    <h1>
+      No Plans Available
+    </h1>
+    <p>
+      There are no plans currently available for this service class.
```
spadgett: nit: "service class" is the k8s name, but we don't use it in the UI. So maybe just, "There are no plans currently available for this service."

src/components/update-service/update-service-results.html
```diff
@@ -0,0 +1,45 @@
+<div class="order-service-config">
+  <div ng-if="!$ctrl.error">
+    <div ng-if="!$ctrl.orderComplete">
```
spadgett: One div?

```html
<div ng-if="!$ctrl.error && !ctrl.orderComplete">
```

src/components/update-service/update-service.controller.ts
```diff
@@ -0,0 +1,355 @@
...
+    }
+    this.ctrl.reviewStep.allowed = false;
+  };
+
+  public showPlan = () => {
+    this.clearValidityWatcher();
+    this.ctrl.configPageShown = false;
+
+    var updateProperties: any = _.get(this.ctrl.parameterSchema, 'properties');
+    this.configStep.hidden = _.size(updateProperties) < 1;
```
spadgett: Just confirming it's not possible to show the dialog without plans or parameter schema. Are we checking for schema `properties` in origin-web-console?

src/components/update-service/update-service.controller.ts
```diff
@@ -0,0 +1,355 @@
...
+    if (this.configStep.hidden) {
+      this.ctrl.nextTitle = 'Update';
+    } else {
+      this.ctrl.nextTitle = "Next >";
+    }
+    this.planStep.valid =  (this.ctrl.selectedPlan !== this.originalPlan) || !this.configStep.hidden;
+
+    this.validityWatcher = this.$scope.$watch("$ctrl.parameterSchema", () => {
+      var updateProperties: any = _.get(this.ctrl.parameterSchema, 'properties');
+      this.configStep.hidden = _.size(updateProperties) < 1;
```
spadgett: Might pull these two lines into a helper since it's repeated from above.

src/components/update-service/update-service.controller.ts
```diff
@@ -0,0 +1,355 @@
...
+        angular.extend(updatedSecretParameters, updateSecretParameters);
+
+        // Remove the reference to this secret
+        updateInstance.spec.parametersFrom = _.reject(updateInstance.spec.parametersFrom, function (parameters: any) {
+          return _.get(parameters, 'secretKeyRef.name') === secret.metadata.name;
+        });
+      }
+    });
+
+    // Add a new secret if any secret parameters were updated
```
spadgett: What happens if a new parameter is added wasn't already set in a secret or in the instance?

spadgett: Also are we handling the case where we're clearing an existing value properly?

src/components/catalog-parameters/catalog-parameters.controller.ts
```diff
@@ -35,22 +63,72 @@ export class CatalogParametersController implements angular.IController {
         success: true
       }
     };
+    if (this.ctrl.isHorizontal) {
+      angular.extend(this.ctrl.parameterFormDefaults.formDefaults, {
```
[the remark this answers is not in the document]
jeff-phillips-18: If isHorizontal changes `setupFormDefaults` is re-called and it would go back to stacked.

src/components/catalog-parameters/catalog-parameters.controller.ts
```diff
@@ -35,22 +63,72 @@ export class CatalogParametersController implements angular.IController {
...
   }
 
-  public $onChanges(onChangesObj: angular.IOnChangesObject) {
-    if (onChangesObj.parameterFormDefinition) {
-      // https://github.com/json-schema-form/angular-schema-form/blob/development/docs/index.md
-      // If no form definition is supplied, show all fields in the schema using '*'.
-      this.ctrl.parameterForm = this.cloneParameterForm(this.ctrl.parameterFormDefinition) || ['*'];
+  private setupReadonlySchema() {
+    if (!this.ctrl.parameterSchema || !this.ctrl.readOnly) {
+      return;
```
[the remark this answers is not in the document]
jeff-phillips-18: When readonly it uses the readonlyParameterSchema. so this is covered. However, we do need to update the parameter form. I will add the readonly changed check for that.

src/components/update-service/update-service.controller.ts
jeff-phillips-18: We are. Should I add a safe guard here? What should we do? Close the panel? Display a message?

src/components/update-service/update-service.controller.ts
jeff-phillips-18: Actually all of this can just be done whenever we update the parameterSchema at that time (no need to watch). Thanks.

src/components/catalog-parameters/catalog-parameters.controller.ts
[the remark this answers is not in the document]
spadgett: OK yeah I see now. Thanks

src/components/update-service/update-service.controller.ts
spadgett: I think it's OK, just wanted to confirm.

src/components/update-service/update-service.controller.ts
jeff-phillips-18: Clearing works properly, if it was in the parameters it will be removed. if it was in a secret, the secret gets removed from its list and a new secret is added.

Still working on checking for new parameters

spadgett:
Were you able to test other read-only input types like checkboxes?

jeff-phillips-18:
Everything but array types, thanks for the reminder. I'll fake it and test it.

jeff-phillips-18:
Addressed all comments thus far, still testing different types in readonly forms

src/components/update-service/update-service.controller.ts
jeff-phillips-18: Added code to add any new parameters to a new secret.

src/components/update-service/update-service.controller.ts
```diff
@@ -0,0 +1,348 @@
...
+      _.set(updateInstance, 'spec.parameters', updateParameters);
+    }
+
+    // Update the secret parameters
+    var updatedSecretParameters: any = {};
+
+    _.each(this.secrets, (secret: any) => {
+      var originalSecretParameters: any = JSON.parse(this.SecretsService.decodeSecretData(secret.data).parameters);
+      var updateSecretParameters: any = _.pick(parameterData, _.keys(originalSecretParameters));
+      newParameters = _.omit(newParameters, _.keys(originalSecretParameters));
```
spadgett: I don't think this handles property names with a `.` in them correctly. It will treat them like JSON object paths.

Maybe (untested)

```js
var keys = _.map(originalSecretParameters, (value: any, key: string) => {
  // Make sure properties with a `.` in the name aren't treated as paths.
  return [ key ];
};
var updateSecretParameters: any = _.pick(parameterData, keys);
newParameters = _.omit(newParameters, keys);
```

src/components/update-service/update-service.controller.ts
```diff
@@ -0,0 +1,348 @@
...
+
+      if (!angular.equals(updateSecretParameters, originalSecretParameters)) {
+        // Add the parameters to the new secret parameters
+        angular.extend(updatedSecretParameters, updateSecretParameters);
+
+        // Remove the reference to this secret
+        updateInstance.spec.parametersFrom = _.reject(updateInstance.spec.parametersFrom, function (parameters: any) {
+          return _.get(parameters, 'secretKeyRef.name') === secret.metadata.name;
+        });
```
spadgett: nit: could use shorthand here:

```js
_.reject(updateInstance.spec.parametersFrom, { secretKeyRef: { name: secret.metadata.name } });
```

src/components/update-service/update-service.controller.ts
```diff
@@ -0,0 +1,348 @@
...
+    } catch (e) {
+      this.Logger.warn('Unable to load parameters from secret ' + _.get(parametersSource, 'secretKeyRef.name'), e);
+    }
+  }
+
+  private updateServiceInstance(updateInstance: any) {
+    // Update the instance
+    var serviceInstancesVersion: any = this.APIService.getPreferredVersion('serviceinstances');
+    var isServiceInstanceReady = this.$filter('isServiceInstanceReady');
+    var isBindingFailed = this.$filter('isBindingFailed');
```
spadgett: Should use `isServiceInstanceFailed`

src/components/update-service/update-service.controller.ts
```diff
@@ -0,0 +1,348 @@
...
+    var bindingFailedMessage = this.$filter('bindingFailedMessage');
```
spadgett: serviceInstanceFailedMessage

src/components/update-service/update-service.controller.ts
jeff-phillips-18: Tested and this is working as is when there is a . in the key. I can update if you aren't comfortable though.

jeff-phillips-18:
Tested with different types (including array and object) found some necessary modifications.

src/components/update-service/update-service.controller.ts
jeff-phillips-18: I went with your suggestion, cleaner to get the keys just once anyhow.

jeff-phillips-18:
Rebased. @rhamilto Any further comments?

jeff-phillips-18:
Updated the update service wizard to use the new wizard-title parameter.

PR #475, opened 2017-10-09
Adopt more service catalog API changes

spadgett:
Requires https://github.com/openshift/origin-web-common/pull/223
Updates origin-web-common to 0.0.66

* Cluster prefix for service classes and plans
* ServiceInstanceCredential -> ServiceBinding
* check removedFromBrokerCatalog

TODO:

- [x] Write test for removedFromBrokerCatalog

cc @jeff-phillips-18

PR #489, opened 2017-10-11
Bump version to 0.0.54

spadgett:


jeff-phillips-18:
Do we want to bump the common version?

spadgett:
@jeff-phillips-18 It was bumped in #475 since it was required for that PR

Issue #484, opened 2017-10-10
Add blank state message when no description


spadgett:
This shouldn't be common, but I'm assuming we want a blank state message here. 

@openshift/team-ux-review ?

![openshift web console 2017-10-10 12-30-35](https://user-images.githubusercontent.com/1167259/31398420-dd03e53c-adb6-11e7-87d5-8ba32bedf61c.png)

ncameronbritt:
I would think something like "No description provided for this service," so that it doesn't look like something is wrong.

rhamilto:
I eliminated "for this service" because the item could not be a service.  Yea/nay?

![screen shot 2017-10-11 at 11 02 22 am](https://user-images.githubusercontent.com/895728/31449713-a55693a4-ae75-11e7-9fe8-242a8b910149.PNG)
![screen shot 2017-10-11 at 11 03 01 am](https://user-images.githubusercontent.com/895728/31449714-a5849d94-ae75-11e7-93a6-3bcfd938c2b6.PNG)
![screen shot 2017-10-11 at 11 03 19 am](https://user-images.githubusercontent.com/895728/31449715-a58f4c12-ae75-11e7-8976-c58a4aa05dfa.PNG)
![screen shot 2017-10-11 at 11 14 43 am](https://user-images.githubusercontent.com/895728/31449716-a59a4ffe-ae75-11e7-9046-674e2f85197b.PNG)
![screen shot 2017-10-11 at 11 15 00 am](https://user-images.githubusercontent.com/895728/31449717-a5a88182-ae75-11e7-847a-672c8265aeee.PNG)

spadgett:
Looks good to me. 3rd screenshot seems like a bug, though, or were you just putting it in to test?

beanh66:
@rhamilto Looks good to me!

rhamilto:
> Looks good to me. 3rd screenshot seems like a bug, though, or were you just putting it in to test?

Yup.  Good eye, @spadgett!

rhamilto:
> Looks good to me. 3rd screenshot seems like a bug, though, or were you just putting it in to test?

I should add that third screenshot is the scenario where the service has a longDescription but not a "short" one.

PR #490, opened 2017-10-11
Adding default message if no description is provided

rhamilto:
Supports https://github.com/openshift/origin-web-catalog/issues/484

screen shot 2017-10-11 at 11 02 22 am
screen shot 2017-10-11 at 11 03 01 am
screen shot 2017-10-11 at 11 14 43 am
screen shot 2017-10-11 at 11 15 00 am


src/components/order-service/order-service-info.html
```diff
@@ -31,7 +31,7 @@
       </span>
       <span ng-if="$ctrl.selectedPlan.spec.description">{{$ctrl.selectedPlan.spec.description}}</span>
     </p>
-    <p ng-if="$ctrl.description" ng-bind-html="$ctrl.description | linky : '_blank'" class="description"></p>
+    <p ng-bind-html="($ctrl.description | linky : '_blank') || 'No description provided.'" class="description"></p>
```
spadgett: I don't think we should show "No description provided." if there's a long description. Only if there's no description at all (short or long).

rhamilto:
Updated to only show "No description provided" if neither description nor longDescription is provided for the service.

PR #493, opened 2017-10-12
Correct handling of select-project dropdown based on number of projects and can create privileges

### LINKED ISSUE #477: Hide the add to project dropdown if no project
I stumbled on this for a second, probably indicating that it could be clearer. Usually if there isn't a choice we make a dropdown into a heading.

I have no projects, should I add to project or create a new? Both? Having no projects (but not realizing it atm since I create & delete them all the time) I found this odd:
screen shot 2017-10-09 at 2 55 32 pm

Hmm, no options, so why is it a drop down?
screen shot 2017-10-09 at 2 59 37 pm

Kind of clear, but is still a bit abnormal.
screen shot 2017-10-09 at 2 59 46 pm

I'm open to disagreement here, just pointing it out as we usually don't show a drop-down menu if there isn't actually an item to select.

@spadgett @jeff-phillips-18 @rhamilto

### LINKED ISSUE #494: Add custom msg for when there are no projects and user cannot create projects
This is for the `<select-projects>` component.

**No existing projects and cannot create projects**
image

Seems like a situation where it would be good to let the admin put a url or email address or something in the message.  We support custom messages and add it to the projects sidebar. We could show it here as well.

dtaylor113:
**Todo:**
- [x] Add all usecases to `<select-project>` component
- [x] Move `<select-project>` to top of form and add no-projects-cant-create empty state to *create-from-builder*
- [x] Add no-projects-cant-create empty state to *order-service*

Fixes #477
Fixes #494

**Latest Screenshots**
*Captions below screenshots*

image
---
image
---
image
---
image
---
image
---
image

### DIFF
--- a/src/components/order-service/order-service.controller.ts
+++ b/src/components/order-service/order-service.controller.ts
@@ -34,6 +34,7 @@ export class OrderServiceController implements angular.IController {
   private selectedProjectWatch: any;
   private bindTypeWatch: any;
   private validityWatcher: any;
+  private noProjectsCantCreateWatch: any;
   private DNS1123_SUBDOMAIN_VALIDATION: any;
 
   constructor($scope: any,
@@ -64,6 +65,7 @@ export class OrderServiceController implements angular.IController {
     this.ctrl.longDescription = this.ctrl.serviceClass.longDescription;
     this.ctrl.docUrl = _.get(this.ctrl.serviceClass, 'resource.spec.externalMetadata.documentationUrl');
     this.ctrl.supportUrl = _.get(this.ctrl.serviceClass, 'resource.spec.externalMetadata.supportUrl');
+    this.ctrl.noProjectsCantCreate = false;
     this.ctrl.applications = [];
     this.ctrl.parameterData = {};
     this.ctrl.bindParameterData = {};
@@ -165,6 +167,10 @@ export class OrderServiceController implements angular.IController {
       this.ctrl.nextTitle = this.bindParametersStep.hidden ? 'Create' : 'Next >';
       this.reviewStep.allowed = this.bindParametersStep.hidden && this.bindStep.valid;
     });
+
+    this.noProjectsCantCreateWatch = this.$scope.$on('no-projects-cannot-create', () => {
+      this.ctrl.noProjectsCantCreate = true;
+    });
   }
 
   public clearValidityWatcher = () => {
@@ -194,7 +200,7 @@ export class OrderServiceController implements angular.IController {
     this.updateBindability();
 
     this.validityWatcher = this.$scope.$watch("$ctrl.forms.orderConfigureForm.$valid", (isValid: any, lastValue: any) => {
-      this.configStep.valid = isValid;
+      this.configStep.valid = isValid && !this.ctrl.noProjectsCantCreate;
       this.bindStep.allowed = this.configStep.valid;
       this.reviewStep.allowed = this.bindStep.hidden && this.configStep.valid;
     });
@@ -325,6 +331,7 @@ export class OrderServiceController implements angular.IController {
   public $onDestroy() {
     this.DataService.unwatchAll(this.watches);
     this.selectedProjectWatch();
+    this.noProjectsCantCreateWatch();
     this.bindTypeWatch();
     this.clearValidityWatcher();
   }
--- a/src/components/select-project/select-project.controller.ts
+++ b/src/components/select-project/select-project.controller.ts
@@ -4,6 +4,7 @@ import * as _ from 'lodash';
 export class SelectProjectController implements angular.IController {
   static $inject = [
     '$filter',
+    '$scope',
     'AuthService',
     'AuthorizationService',
     'KeywordService',
@@ -17,6 +18,7 @@ export class SelectProjectController implements angular.IController {
   public ctrl: any = this;
 
   private $filter: any;
+  private $scope: any;
   private ProjectsService: any;
   private Logger: any;
   private AuthService: any;
@@ -30,6 +32,7 @@ export class SelectProjectController implements angular.IController {
   private lastResults: any;
 
   constructor($filter: any,
+              $scope: any,
               AuthService: any,
               AuthorizationService: any,
               KeywordService: any,
@@ -37,6 +40,7 @@ export class SelectProjectController implements angular.IController {
               ProjectsService: any,
               RecentlyViewedProjectsService: any) {
     this.$filter = $filter;
+    this.$scope = $scope;
     this.AuthService = AuthService;
     this.AuthorizationService = AuthorizationService;
     this.KeywordService = KeywordService;
@@ -51,6 +55,16 @@ export class SelectProjectController implements angular.IController {
 
   public $onInit() {
     this.ctrl.nameTaken = false;
+    this.ctrl.noProjectsCantCreate = false;
+
+    this.ctrl.noProjectsConfig = {
+      title: 'No Projects Found',
+      info: "Services cannot be provisioned without a project."
+    };
+
+    if (this.ctrl.showDivider === undefined) {
+      this.ctrl.showDivider = true;
+    }
 
     this.ProjectsService.canCreate().then(() => {
       this.ctrl.canCreate = true;
@@ -64,6 +78,20 @@ export class SelectProjectController implements angular.IController {
           msg += " (" + result.status + ")";
         }
         this.Logger.warn(msg);
+      }
+
+      var data = result.data || {};
+
+      if (data.details) {
+        var messages = [];
+        _.forEach(data.details.causes || [], function (cause: any) {
+          if (cause.message) {
+            messages.push(cause.message);
+          }
+        });
+        if (messages.length > 0) {
+          this.ctrl.noProjectsCantCreateMessage = messages.join("\n");
+        }
       }
     }).finally(() => {
       this.listProjects();
@@ -93,6 +121,10 @@ export class SelectProjectController implements angular.IController {
 
   public isNewProject(): boolean {
     return this.projects && this.ctrl.selectedProject && !_.has(this.ctrl.selectedProject, 'metadata.uid');
+  }
+
+  public canOnlyCreateProject(): boolean {
+    return this.ctrl.numProjectChoices === 1 && this.ctrl.canCreate;
   }
 
   public getProjectChoices = () => {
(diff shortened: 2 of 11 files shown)


beanh66:
@dtaylor113 these look awesome, thanks for sharing all the screenshots. My only comment would be on the final use case (No projects and can't create a project, default message). I think the code snippet should be styled differently if possible. On the Homepage it is styled that way because of the dark background of the project panel, but on a white background I think we would show code with the light grey background and normal font color. Something like this:
 `oc adm new-project <projectname> --admin=<yourUsername>`

@jennyhaines can you confirm?

jennyhaines:
@dtaylor113 @beanh66 - Yes, I agree with Colleen about styling this code snippet differently. Right now it's too stark/contrasty. I would actually use the same style that PatternFly uses for [inline code](http://www.patternfly.org/pattern-library/widgets/#cnt-remaining-chars). (Blue on light blue.) 

Here's an example of that style implemented in OpenShift Container Registry (circa June 2017) 
oscontainerregistry

spadgett:
If you put it in a `<code>` tag, it should use this style. It's what we use elsewhere.

https://github.com/openshift/origin-web-console/blob/master/app/views/_cannot-create-project.html#L2

beanh66:
Perfect, thanks for the update @jennyhaines!

jennyhaines:
@dtaylor113 In that case, I would default to using the <code> tag that @spadgett mentioned, since you're using that elsewhere (rather than using the blue style in PF).

dtaylor113:
@beanh66, ok changed to:

image

dtaylor113:
@beanh66, changed no projects/empty state msg to "Services cannot be provisioned without a project."

beanh66:
@dtaylor113 LGTM!

spadgett:
@dtaylor113 you can test large project lists using this option with mock data:

https://github.com/openshift/origin-web-catalog/blob/master/app/config.js#L46-L47

src/components/select-project/selectProject.html
```diff
@@ -1,9 +1,23 @@
-<ng-form name="$ctrl.forms.selectProjectForm">
+<span ng-if="$ctrl.noProjectsCantCreate" class="no-projects-cant-create">
+  <pf-empty-state config="$ctrl.noProjectsConfig"></pf-empty-state>
+  <div ng-if="!$ctrl.noProjectsCantCreateMessage">
+    A cluster admin can create a project for you by running the command:
+    <div class="code-block">
+      <code>oc adm <span class="command-arg">new-project</span> &lt;projectname&gt; <span class="command-arg">--admin={{$ctrl.user.metadata.name || '&lt;YourUsername&gt;'}}</span></code>
```
spadgett: Is `$ctrl.user.metadata.name` ever set?

dtaylor113:
> @dtaylor113 you can test large project lists using this option with mock data:

Hi @spadgett, I did have to make a one line change to get the large project list/filter to work.   I noticed that if you have a large project list and you can create projects, you do not get a change to create a project -the 'Create Project' option is not added to the dropdown.  So you can only select a project from the large list, you cannot create one.  Is this intentional?

spadgett:
It's intentional for now. The problem is that if we added "Create Project" to the dropdown, it's not obvious you can search for other projects when you have large project lists.

src/components/select-project/selectProject.html
dtaylor113: Nice catch!  Thanks, fixed:
image

dtaylor113:
> @dtaylor113 you can test large project lists using this option with mock data:

Fixed, tested

src/components/select-project/select-project.controller.ts
```diff
@@ -95,6 +120,10 @@ export class SelectProjectController implements angular.IController {
     return this.projects && this.ctrl.selectedProject && !_.has(this.ctrl.selectedProject, 'metadata.uid');
   }
 
+  public canOnlyCreateProject(): boolean {
```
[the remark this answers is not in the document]
dtaylor113: I call `canOnlyCreateProject()` three times from the HTML, I could replace each call with `(this.ctrl.numProjectChoices === 1 && this.ctrl.canCreate)`, or make canOnlyCreateProject a variable.  Though a helper method was warranted :-)

src/components/select-project/select-project.controller.ts
```diff
@@ -214,12 +247,18 @@ export class SelectProjectController implements angular.IController {
...
       if (_.size(this.projects) === 1) {
         this.ctrl.selectedProject = createProject;
         this.onSelectProjectChange();
       }
+    } else if (_.size(this.projects) === 0) {
+      this.ctrl.noProjectsCantCreate = true;
+      this.$scope.$emit('no-projects-cannot-create');
     }
+
+    this.ctrl.numProjectChoices = _.size(this.projects);
```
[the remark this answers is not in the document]
dtaylor113: Fixed, tested

dtaylor113:
> the line is visible even when there are no other configuration parameters.
Will there ever be a usecase where `<select-project>` is the only config options?  To fix this I'd have to move the line out of the `<select-project>` component and move it to the 5 wizards (create-from-builder, deploy-image, etc..), and only show it if the wizard has other config params/settings.   -which I can do if we feel it is necessary.

spadgett:
> Will there ever be a usecase where <select-project> is the only config options?

Yes. Not all services take parameters.

Maybe just add an optional flag to select-project saying there are no other fields?

dtaylor113:
> the line is visible even when there are no other configuration parameters.

Fixed:
image

src/styles/order-service.less
```diff
@@ -116,13 +116,25 @@
...
+    margin-left: -10px;
+  }
   h3 {
     line-height: 1.4;
     margin-top: 0;
     + .alert {
       margin-top: 20px;
     }
   }
+  .select-project-divider {
```
rhamilto: Nit: alpha

src/styles/order-service.less
```diff
@@ -116,13 +116,25 @@
...
+    border-bottom: 1px solid @color-pf-black-400;
+
```
rhamilto: Nit:  unnecessary line

src/styles/order-service.less
```diff
@@ -116,13 +116,25 @@
...
+    margin-top: 22px;
```
rhamilto: Nit: alpha

src/components/select-project/selectProject.html
```diff
@@ -1,9 +1,23 @@
-<ng-form name="$ctrl.forms.selectProjectForm">
+<span ng-if="$ctrl.noProjectsCantCreate" class="no-projects-cant-create">
+  <pf-empty-state config="$ctrl.noProjectsConfig"></pf-empty-state>
+  <div ng-if="!$ctrl.noProjectsCantCreateMessage">
```
rhamilto: A &lt;p&gt; is a better choice here as it adds semantic meaning.

src/components/select-project/selectProject.html
```diff
@@ -1,9 +1,23 @@
-<ng-form name="$ctrl.forms.selectProjectForm">
+<span ng-if="$ctrl.noProjectsCantCreate" class="no-projects-cant-create">
```
rhamilto: &lt;div&gt; is probably a better choice here since this is a decent-sized block of markup.

dtaylor113:
@rhamilto, thanks for review, I believe I have addressed all your issues.

src/styles/order-service.less
```diff
@@ -116,13 +116,19 @@
...
+    margin-left: -10px;
+  }
   h3 {
     line-height: 1.4;
     margin-top: 0;
     + .alert {
       margin-top: 20px;
     }
   }
+  .no-projects-cant-create p, .no-projects-cant-create div  {
```
rhamilto: Nit:  would be great to nest here:

```
.no-projects-cant-create {
  div,
  p {

```

rhamilto:
The use of pf-empty-state here and https://github.com/openshift/origin-web-catalog/pull/501 gives me pause as this introduces a new empty state style to the console.  The console had an empty state pattern before pf-empty-state existed and the existing pattern is being used in all contexts that need it.  Adding pf-empty-state makes the aforementioned instances inconsistent with the rest of the console.  I understand the desire to use the official PatternFly components, but I think it's actually a negative in this case because it's introducing inconsistency in the UI.

serenamarie125:
@rhamilto there are already inconsistencies.  See the associated design story identifying this issue https://patternfly.atlassian.net/browse/OSUX-155

spadgett:
> there are already inconsistencies. See the associated design story identifying this issue https://patternfly.atlassian.net/browse/OSUX-155

2 of the 4 screenshots in that issue are from new designs in 3.7 that didn't account for what we had done elsewhere in the console. And I'd argue the other two from before 3.7 are more similar than different. This is mostly a problem we've introduced in 3.7. Adding yet another empty state style seems like a step backwards.

I'm not opposed to changing the style, but I really value consistency. I'd rather change them all to be consistent when we are able to.

serenamarie125:
@spadgett I'm not arguing that we shouldn't make them consistent, just pointing out that the consistency already exists, and we are taking a look at it

dtaylor113:
Hi @spadgett, which empty state would you like me to use in this case?  Could you please send a link to an example in the code base.  -thanks

dtaylor113:
Hi @serenamarie125, @spadgett,  any chance of letting this PR merge and addressing all the empty state versions in a follow up PR when we address https://patternfly.atlassian.net/browse/OSUX-155?

dtaylor113:
**Removed the icon from the empty state:**

image

src/components/order-service/order-service.controller.ts
```diff
@@ -194,7 +200,7 @@ export class OrderServiceController implements angular.IController {
     this.updateBindability();
 
     this.validityWatcher = this.$scope.$watch("$ctrl.forms.orderConfigureForm.$valid", (isValid: any, lastValue: any) => {
-      this.configStep.valid = isValid;
+      this.configStep.valid = this.ctrl.noProjectsCantCreate === true ? false : isValid;
```
spadgett:
```js
this.configStep.valid = isValid && !this.ctrl.noProjectsCantCreate;
```

src/components/select-project/select-project.controller.ts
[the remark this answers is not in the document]
spadgett: I see now. I think this is fine.

spadgett:
@dtaylor113 LGTM except for my one nit. @rhamilto had one or two comments pending as well

src/components/order-service/order-service.controller.ts
dtaylor113: Fixed

src/styles/order-service.less
dtaylor113: Fixed

PR #498, opened 2017-10-13
Fix to scroll when overlay panel is too tall for viewport

jeff-phillips-18:
Simple fix for https://github.com/openshift/origin-web-console/issues/2095

### DIFF
--- a/dist/less/overlay-panel.less
+++ b/dist/less/overlay-panel.less
@@ -115,6 +115,8 @@ body.overlay-open {
 .catalogs-overlay-panel-wrapper {
   bottom: 0;
   left: 0;
+  overflow-x: hidden;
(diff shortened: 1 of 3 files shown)


spadgett:
Does the dialog still need a max height?

jeff-phillips-18:
I don't think that is necessary. We could set one, but the user can scroll the view to see the dialog contents that are below the viewport. We could set one on a per dialog basis and handle it per dialog.

If we want a more complex panel that does header, body, footer (either via parameters or transcludes) we can then do some settings such that the body scrolls.

spadgett:
Ah OK, this scrolls the entire page? I think that makes sense and is consistent with Bootstrap modal.

PR #500, opened 2017-10-16
Close the overlay panel on ESC

### LINKED ISSUE #326: Accessibility: Escape key should close the overlay
Suggested by @ncameronbritt, and I've had the same thought. I've tried to close the overlay with the escape key before, and it doesn't work. It's potentially an accessibility issue and inconsistent with our other Bootstrap modals.

Maybe we can optionally prompt if the user has typed something.

cc @openshift/team-ux-review

jeff-phillips-18:
Fixes #326

### DIFF
--- a/src/components/overlay-panel/overlay-panel.controller.ts
+++ b/src/components/overlay-panel/overlay-panel.controller.ts
@@ -2,9 +2,15 @@ import * as angular from 'angular';
 import * as $ from 'jquery';
 
 export class OverlayPanelController implements angular.IController {
-  public ctrl: any = this;
+  static $inject = ['$document', '$scope'];
 
-  constructor () {
+  public ctrl: any = this;
+  private $document: any;
+  private $scope: any;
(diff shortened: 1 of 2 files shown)


src/components/overlay-panel/overlay-panel.controller.ts
```diff
@@ -37,10 +43,22 @@ export class OverlayPanelController implements angular.IController {
...
 
   private hideDialog = () => {
     this.ctrl.shown = false;
     $('body').removeClass('overlay-open');
-  }
+    this.$document.unbind('keydown keypress', this.closeOnEsc);
+  };
+
+  private closeOnEsc = (event: any) => {
+    console.log("Keydown: " + event.which);
```
spadgett: remove console log

jeff-phillips-18:
Removed the console.log.  @spadgett tested and it does not close when esc on a ui-select.

PR #501, opened 2017-10-17
Adding empty state component when there are no filter matches

### LINKED ISSUE #461: Page feels empty when no filters match
@openshift/team-ux-review Should we have a blank state message or something below the filters saying there are no matches?

cc @dtaylor113 @jeff-phillips-18 

openshift catalog components 2017-10-02 13-48-06

amarie401:
An empty state message when the filters result in no matches while also allowing users to clear the applied filters through the url [by adding an optional urlAction property to the angular-pf emptystate component](https://github.com/patternfly/angular-patternfly/pull/648)

NOTE: A-PF will need to be updated to introduce this new config property in order for the Clear Filters link to work.

https://github.com/patternfly/angular-patternfly/pull/650

This would close #461 

screen shot 2017-10-19 at 1 28 54 pm

@spadgett @dtaylor113 @beanh66


beanh66:
@amarie401 The styling looks great! After looking through more OpenShift and PatternFly empty state examples, wondering if we can we make a slight adjustment to the copy to simplify? Suggestion below:

No results match. (title)
The active filters are hiding all catalog items. (description text) 
Clear Filters (link)

src/components/services-view/services-view.controller.ts
```diff
@@ -63,6 +63,17 @@ export class ServicesViewController implements angular.IController {
       resultsLabel: "Items",
       appliedFilters: []
     };
+
+    this.ctrl.emptyFilterConfig = {
+      icon: 'pficon-info',
+      title: 'No results match the filter criteria',
```
spadgett: Per @beanh66 `'No results match.'`

src/components/services-view/services-view.controller.ts
```diff
@@ -63,6 +63,17 @@ export class ServicesViewController implements angular.IController {
...
+      info: 'The active filters are currently hiding all catalog items.',
+      helpLink: {
+        label: 'To view all items',
```
spadgett: I would leave `label` blank so you just see "Clear Filters"

spadgett:
@amarie401 Sorry, just realized you probably haven't updated the PR yet since @beanh66's comment!

amarie401:
I have made requested changes @spadgett @dtaylor113 @beanh66

amarie401:
I've tried a few different combinations and they seem to pass in karma so not sure why the test is failing in Travis..

I've tried:

```1) var filterLink = jQuery(element).find('pf-empty-state');
      filterLink.find('.blank-state-pf-helpLink').click();
      expect(jQuery(element).find('.active-filter.label.label-info').length).toBe(0);

      2) var filterLink = jQuery(element).find('.blank-slate-pf');
      filterLink.find('.blank-state-pf-helpLink').click();
      expect(jQuery(element).find('.active-filter').length).toBe(0);

      3) componentTest.eventFire(element.querySelector('.blank-state-pf-helpLink'), 'click');
      expect(jQuery(element).find('.active-filter.label.label-info').length).toBe(0);

**Comment by jeff-phillips-18** [comment]
Try  adding `componentTest.scope.$digest();` after the click.

**Comment by amarie401** [comment]
I've added that to a few of the combinations I've tried but still failing :/

**Comment by dtaylor113** [comment]
Hi, looking into this...

**Comment by amarie401** [comment]
After looking into it, it seems like firefox is causing the unit test issues that I am experiencing. When changing the test config to use firefox it then fails. I noticed this was brought up in issue #433 . What would be the best approach?

@dtaylor113 @spadgett @jeff-phillips-18

**Comment by dtaylor113** [comment]
Hi @amarie401,
I'm using FireFox v49 and the unit tests pass when run locally.  I suspect it is an event timing issue.  When the 'Clear Filters' link is clicked it asynchronously broadcasts an event, which maybe happening after the synchronous unit test has run.

In the failing unit test, you have two ".toBe(0)" expectations.  I'd add a log message to each one.  For example: 
`expect(jQuery(element).find('pf-empty-state').length).toEqual(0, 'empty state not shown);` 
...and...
`expect(jQuery(element).find('.active-filter.label.label-info').length).toBe(0, 'no filter tags');`.
...just to verify which expect 0 is actually failing.

Next, you can have the unit test spit out the HTML it is seeing by adding:
`console.log(jQuery(element)[0].innerHTML);`
...right before the last expect(..) in the unit test.

I was going to suggest that it maybe better to test that the `clearAppliedFilters` method is being called via:
```
var ctrl = componentTest.isoScope.$ctrl;
spyOn(ctrl, 'clearAppliedFilters');
//Click clear filter link
...
expect(ctrl.clearAppliedFilters).toHaveBeenCalled();
```
However, I get a failure that clearAppliedFilters wasn't called.   In general, if you can't get this unit test to run, I'd just remove it for this PR and perhaps follow up with another PR which adds in this unit test.

File: src/components/services-view/services-view.html [hunk 3]
```diff
@@ -100,6 +100,9 @@
                           filter-on-keyword="$ctrl.keywordFilterValue"
                           apply-filters="$ctrl.applyFilters($event)">
           </catalog-filter>
+          <div class="col-xs-12" ng-if="$ctrl.filterConfig.appliedFilters.length > 0 && $ctrl.filterConfig.resultsCount === 0">
```
rhamilto: class="col-xs-12" should not be needed.

amarie401: I added that because without it then the pfEmptyState doesn't span the full width. Any alternatives that you suggest?

amarie401: screen shot 2017-10-18 at 3 43 08 pm

rhamilto: Ah.  I'd be inclined to write a rule for this case.  col-xs-12 works to get the appearance you're looking for, but isn't really intended for this application.

amarie401: Could you clarify what you mean by write a rule for this case?

amarie401: I see what you're saying. I'll go back and fix.

rhamilto: At https://github.com/openshift/origin-web-catalog/blob/master/src/styles/services-view.less#L13, add:

```
.pf-empty-state {
  margin: 0 (@grid-gutter-width / 2);
  width: 100%;
}
```

and then change your [markup class](https://github.com/amarie401/origin-web-catalog/blob/02c97bc365b6d9ad77ce76c16205328a0300eded/src/components/services-view/services-view.html#L103) from `col-xs-12` to `pf-empty-state`.

You also need to include &lt;pf-empty-state&gt; for [subcategories](https://github.com/amarie401/origin-web-catalog/blob/02c97bc365b6d9ad77ce76c16205328a0300eded/src/components/services-view/services-view.html#L78)?

screen shot 2017-10-18 at 4 25 13 pm

rhamilto:
The "Clear Filters" link inside pf-empty-state doesn't appear to work.

amarie401:
@rhamilto I made your suggested changes to the less. Also I'm not sure why the Clear Filters link isn't working for you. It is working on my local. Have you pulled down the emptystate-filter branch?

rhamilto:
> Have you pulled down the emptystate-filter branch?

Yep.  And I updated the clearAppliedFilters function to console log each time it is called, which is only when I navigate categories and subcategories.  :(

rhamilto:
https://github.com/patternfly/angular-patternfly/commit/23e37790 is why yours works on mine doesn't.  Please note the dependency in your PR.  This PR should not merge without the dependency update!

amarie401:
@rhamilto ahhh yes. In this PR I had linked to the a-pf PR where I introduced the urlAction property to the empty-state config to be used for this issue. Will make sure to note that more clearly in this PR. Thanks!

jeff-phillips-18:
@amarie401 Can you update the version of angular-patternfly to 4.11.5 as part of this PR?

amarie401:
@jeff-phillips-18 I can 👍

jeff-phillips-18:
https://github.com/openshift/origin-web-catalog/pull/501
image

IMO, either the background should be white or the entire area should be grey.

@beanh66

serenamarie125:
Given the PatternFly guidelines I would make it grey.

amarie401:
@serenamarie125 I agree

amarie401:
With the grey background on ```.service-items``` & border on ```.pf-blank-slate```

screen shot 2017-10-18 at 6 11 15 pm

With the grey background on ```.service-items``` &  border removed from ```.pf-blank-slate```

screen shot 2017-10-18 at 6 14 58 pm

serenamarie125:
@amarie401 I'd expect that the toolbar would still have a white background, and that the area below the separator would be grey.  Does that make sense ?

dtaylor113:
> You also need to include <pf-empty-state> for subcategories?

Hi @amarie401, just want to make sure this issue gets addressed.  You need to add a pf-empty-state [here](https://github.com/amarie401/origin-web-catalog/blob/48b80a3f3cca1bdc4c22e93e9ea693c0cf2b97c8/src/components/services-view/services-view.html#L78)  -thanks

amarie401:
@rhamilto how could we go about making the empty state component span the full height. If we want to keep the toolbar and space around the empty state white?

rhamilto:
> @rhamilto how could we go about making the empty state component span the full height. If we want to keep the toolbar area white?

It's going to take some CSS hackin'.  We're doing a number of things to manipulate that box already (note the presence of "sass offerings" below the search bar impacts the way that white box renders).  I think the simplest approach is to not use the gray background as it alleviates the need to solve this problem and removes the inconsistency using the gray background introduces).

rhamilto:
This will turn off the background color and border for any pf-empty-state directives inside the service area.

```
.services-view .blank-slate-pf {
  background-color: transparent;
  border: none;
}
```

amarie401:
I have updated the styling to use a transparent background as @rhamilto suggested and added the pfEmptyState component to the subcategory filters as well @dtaylor113

dtaylor113:
LGTM, thanks!

spadgett:
@amarie401 I think you need to rebase to pick up the angular-patternfly version update. The "Clear Filters" link isn't working for me otherwise.

src/styles/services-view.less
```diff
@@ -10,6 +10,14 @@ services-view,
 }
 
 .services-view {
+  .pf-empty-state {
```
rhamilto: Nit:  we alphabetize rules by selector, so this should go after .blank-slate-pf

amarie401: Okie dok will do!

amarie401:
@spadgett I believe I had the updated version already on this current [branch](https://github.com/amarie401/origin-web-catalog/blob/emptystate-filter/package.json#L33).

spadgett:
Thanks @amarie401

PR #487, opened 2017-10-10
Look for provider display name annotation on image streams

spadgett:
Waiting for consensus on the name in https://github.com/openshift/origin/issues/16755

@sspeiche

PR #508, opened 2017-10-18
Bump version to v0.0.56

rhamilto:


spadgett:
LGTM but we might wait for #474 since it's close.

cc @jeff-phillips-18

spadgett:
I'd like to include #487 too if possible, but that depends on https://github.com/openshift/origin/pull/16925 (currently merging)

PR #510, opened 2017-10-18
Fix bug where plan info displayed on info screen but shouldn't

### LINKED ISSUE #499: Should not show plan description on info page when there's more than one plan
We show the plan description on the info page when there's only one so that the description isn't lost if there is important information. This isn't a problem if there is more than one plan, so we don't need to show it here. It might be misleading in some cases. Here there is a plan option that uses persistent storage instead of ephemeral, but the info page makes it sound like it's only ephemeral.

openshift web console 2017-10-16 10-41-10

rhamilto:
Fixes #499 

screen shot 2017-10-18 at 10 48 18 am


src/components/order-service/order-service.controller.ts
```diff
@@ -71,6 +71,7 @@ export class OrderServiceController implements angular.IController {
 
     this.ctrl.appToBind = null;
     this.ctrl.configStepValid = true;
+    this.ctrl.lessThanTwoServicePlans = _.size(this.ctrl.servicePlans) < 2;
```
spadgett: Maybe call it `hasMultiplePlans` and flip the logic? I feel like that reads better.

rhamilto:
Now with reversed logic for better readability.  Thanks, @spadgett.

src/components/order-service/order-service-info.html
```diff
@@ -27,7 +27,7 @@
     </div>
   </div>
   <div class="order-service-description-block">
-    <p ng-if="$ctrl.currentStep.id !== 'plans' && ($ctrl.selectedPlan.spec.externalMetadata.displayName || $ctrl.selectedPlan.spec.description)">
+    <p ng-if="$ctrl.currentStep.id !== 'plans' && !$ctrl.multipleServicePlans && ($ctrl.selectedPlan.spec.externalMetadata.displayName || $ctrl.selectedPlan.spec.description)">
```
spadgett: Just noticed this, but can't be remove the `$ctrl.currentStep.id !== 'plans'` check? This is always the info step.

rhamilto: Yep.  Good catch!  Fixed.

PR #518, opened 2017-10-20
Fix readonly parameters margins

jeff-phillips-18:
Also fixes next button title going back to the plan page of the update service wizard.

### DIFF
--- a/dist/less/catalogParameters.less
+++ b/dist/less/catalogParameters.less
@@ -8,6 +8,19 @@
     padding-left: 0;
   }
   &.readonly {
+    .control-label {
+      padding-right: 0;
+      word-wrap: break-word;
+    }
+
+    form {
+      margin-top: 10px;
+    }
+
+    .form-group {
+      margin-bottom: 0;
+    }
+
     fieldset {
       padding-left: 20px;
       > div {
--- a/dist/origin-web-catalogs.css
+++ b/dist/origin-web-catalogs.css
@@ -248,6 +248,16 @@ landingbody,
   margin-top: -3px;
   padding-left: 0;
 }
+.catalog-parameters.readonly .control-label {
+  padding-right: 0;
+  word-wrap: break-word;
+}
+.catalog-parameters.readonly form {
+  margin-top: 10px;
+}
+.catalog-parameters.readonly .form-group {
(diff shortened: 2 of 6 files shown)


dist/less/catalogParameters.less
```diff
@@ -8,6 +8,19 @@
     padding-left: 0;
   }
   &.readonly {
+    .control-label {
+      padding-right: 0;
+      word-break: break-word;
```
spadgett: I don't think this is a standard `word-break` value. Do you mean `word-wrap: break-word;`?

https://developer.mozilla.org/en-US/docs/Web/CSS/word-break

spadgett:
@jeff-phillips-18 Do you mind pasting a screenshot?

dist/less/catalogParameters.less
jeff-phillips-18: Yes, sorry.

jeff-phillips-18:
image

spadgett:
I feel like the parameters need some top margin. It seems crowded.

Should we reuse the same font-style as the overview for the "PARAMETERS" header? It looks a bit bigger here.

The left indentation looks strange to me still.

@jwforres @beanh66 opinion?

jwforres:
+1 to needing some top margin
+1 to indention looking strange

beanh66:
@spadgett After seeing this, agreed all around. 

+1 to adding more space above PARAMETERS
+1 to left aligning, I think we will be fine without the indent since we have the PARAMETERS header
+1 the header should be the same font we use on the overview for things like DEPLOYMENT

jeff-phillips-18:
Those changes:
image

jeff-phillips-18:
For what it's worth, that styling is all in console and will not be part of this PR.

serenamarie125:
PARAMETERS looks to be a smaller font than Reveal Values now @spadgett @jeff-phillips-18

jeff-phillips-18:
Yes, 11px vs 13px

jeff-phillips-18:
A few more tweaks (again not in this PR):
image

serenamarie125:
not as noticeable now that you've changed the spacing.  Looks better to me!

PR #522, opened 2017-10-21
[WIP] Support `addToProject` param to preselect project

### LINKED ISSUE #521: Can't select project with mock data
AuthorizationService has not been properly mocked.

```
Error: _this.AuthorizationService.getProjectRules is not a function
```

spadgett:
Needed to preselect the last project when clicking "Browse Catalog" from within a project.

Closes #521

### DIFF
--- a/app/mockServices/mockAuthorization.service.ts
+++ b/app/mockServices/mockAuthorization.service.ts
@@ -1,8 +1,21 @@
 interface IAuthorizationService {
   canIAddToProject(projectName: string) : boolean;
 }
+import * as angular from 'angular';
 
 export class AuthorizationService implements IAuthorizationService {
+  public static $inject = ['$q'];
+
+  private $q : any;
+
+  constructor ($q: angular.IQService) {
+    this.$q = $q;
+  }
+
+  public getProjectRules() : angular.IPromise < any > {
+    return this.$q.when({});
+  }
+
   public canIAddToProject (projectName: string) {
     // can add to any project except 'my-proj-a'
     return !(projectName === 'my-proj-a');
--- a/dist/origin-web-catalogs.js
+++ b/dist/origin-web-catalogs.js
@@ -1859,24 +1859,25 @@ webpackJsonp([ 0, 1 ], [ function(e, t) {
     "use strict";
     t.__esModule = !0;
     var n = r(1), i = r(0), a = function() {
-        function e(t, r, n, a, s, c, o, l) {
-            var d = this;
+        function e(t, r, n, a, s, c, o, l, d) {
+            var p = this;
             this.ctrl = this, this.getProjectChoices = function() {
-                return d.ctrl.matchingProjects ? d.ctrl.matchingProjects : d.largeProjectList ? [] : d.projects;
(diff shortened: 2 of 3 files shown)


jeff-phillips-18:
Seems to me the select project component isn't the right place to be checking for location parameters. Should be the parent controller passing along the pre-selected project.

spadgett:
Yeah, I'm trying to find a clean way to do this. I'm a little worried we might have bugs if we try to pass it  to 5 different dialogs through multiple layers of components. Maybe it's not as bad as I think.

dtaylor113:
LGTM but I do second Jeff's comment.

spadgett:
I have the changes locally to pass it down through all the components instead. I had to update 9 different components to pass it through several layers. The change is pretty large, and I'm worried this is going to be error prone and hard to maintain.

I'm not crazy about reading the query param directly either.

Any thoughts on a clean way to handle this? Maybe adding`CatalogService.preselectProject()` / `CatalogService.getPreselectedProject()` methods that the select-project component could use? Then it's not dealing with the URL params directly.

If possible, I'd really like to keep the code that handles this in just a few places so that the behavior is always consistent across all dialogs and it's maintainable. For what it's worth, we have a few other components that read and set query parameters like the tab set, which sets the `?tab=` param. That has worked well.

cc @jwforres

jeff-phillips-18:
I'm actually not sure I agree with the premise here. If I am on the landing page, I don't know that I have a project selected at all. I'm not sure I would remember that I hit 'Browse Catalog' from the project overview page. There is no indication which project I came from therefore, I don't think we should pre-select a project at all.

The real issue is that 'Browse Catalog' takes me to the landing page which is really much more than just browsing the catalog.

jwforres:
Sounds like too large / risky of a change to plumb it down from the controllers.  I'm not opposed to reading the query param since its something we do today.  But I'm also fine with the two methods on the Service.

Even using the service methods you could still have a query param, if the controller sees the query param it calls the method, as opposed to plumbing it through.

jwforres:
agreed @jeff-phillips-18 , the flow is a bit weird.  We are planning to fix that long term though correct? With the designs @cshinn is working on for catalog viewing within the project?

spadgett:
> There is no indication which project I came from therefore, I don't think we should pre-select a project at all.

It's not really a problem if you only have a few projects. If you are cluster-reader on a large cluster, you have to find your project again among thousands of others, which is difficult and annoying.

Maybe that's uncommon enough that we punt on it.

> The real issue is that 'Browse Catalog' takes me to the landing page which is really much more than just browsing the catalog.

I agree this is the real problem. Is there anything we should do in 3.7 to improve the experience?

spadgett:
OK I think there's  enough disagreement about both user experience and implementation that we punt on this for 3.7 and look at implementing the design from @cshinn in a future release.

Quick add (https://trello.com/c/UWfRPJwE) would also help.

jeff-phillips-18:
With some small styling changes and very little extra code, we could probably add a Browse Catalog navigation item to the project page to get something like this:

image

Just sayin... 😉

jwforres:
waaaaay too late in the release for that

PR #524, opened 2017-10-23
Adopt renamed service external name properties

spadgett:
Bumps origin-web-common to 0.0.70

https://github.com/openshift/origin-web-common/releases/tag/v0.0.70

Requires https://github.com/openshift/origin-web-common/pull/235


spadgett:
Rebased and added the common version bump. Tested again with the service catalog API changes.

PR #527, opened 2017-10-24
Bug 1504973 - Confirm password values

spadgett:
Let the user confirm the entered password values when provisioning a service.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1504973

screencast 2017-10-24 11-58-07


spadgett:
@cfchase

jwforres:
LGTM but hold till @spadgett is done testing

spadgett:
OK I was able to validate the fix with a real APB that uses password fields

PR #529, opened 2017-10-24
Fix planUpdatable check

spadgett:
The logic was reversed deciding to hide the plans step when updating a service instance.

See https://github.com/openshift/origin-web-console/issues/2350

### DIFF
--- a/dist/origin-web-catalogs.js
+++ b/dist/origin-web-catalogs.js
@@ -2158,7 +2158,7 @@ webpackJsonp([ 0, 1 ], [ function(e, t) {
                 id: "plans",
                 label: "Plan",
                 view: "update-service/update-service-plans.html",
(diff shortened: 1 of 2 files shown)


spadgett:
@jmontleon

jmontleon:
This allows me to update the plan successfully now.

Maybe better to open a separate issue for this one, but what I see now is:
We specify in the dev plan that it can be updated to prod in the updates_to array in the dev plan and I can do that. (see comment at: https://github.com/openshift/origin-web-console/issues/2350 for catalog output.)

However we don't specify that prod can be 'downgraded' to dev (the updates_to array is empty and ommited in the catalog output, but the UI still lets me try to do that.

spadgett:
@jmontleon Can you open a separate issue for `updates_to`? I don't see it mentioned in the spec.

https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md

cc @pmorie

PR #530, opened 2017-10-24
Add bottom margin to filter panel when under sub-category

### LINKED ISSUE #505: Filter dropdown ul gets different bottom margin depending on whether its in a subcategory
When looking at All then the filter is not in a nested UL, but if you pick any subcategory then the filter checkboxes are in a nested UL and bottom margin is removed due to base patternfly styles.

The UL that needs to be fixed is `pf-filter-panel .filter-panel-container ul`

screen shot 2017-10-17 at 3 44 03 pm
screen shot 2017-10-17 at 3 44 27 pm

dtaylor113:
Fixes #505 

image

### DIFF
--- a/src/styles/services-view.less
+++ b/src/styles/services-view.less
@@ -14,6 +14,9 @@ services-view,
     background-color: transparent;
     border: none;
   }
+  .filter-panel-container ul {
+    margin-bottom: 10px;
+  }
   .pf-empty-state {
     margin: 0 (@grid-gutter-width / 2);
     width: 100%;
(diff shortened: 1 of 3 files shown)


src/styles/services-view.less
```diff
@@ -7,6 +7,9 @@ services-view,
     padding-left: 2px;
     vertical-align: middle;
   }
+  pf-filter-panel .filter-panel-container ul {
```
rhamilto: This is a bit of overkill because of where the rule is nested and the specificity.   Try

```
.filter-panel-container ul {
  margin-bottom: 10px;
}
```
on line 20.  Should produce the same result, but without duplication and overspecificity.

dtaylor113: Thanks, Fixed

PR #535, opened 2017-10-24
Provide ability to get service plans for a given service class

jeff-phillips-18:


src/services/catalog.service.ts
```diff
@@ -70,6 +70,30 @@ export class CatalogService { static $inject = ['$filter', '$q', 'Constants', 'A
...
   }
 
+  // Get service plans for a service class, takes either the serviceClass object or it's name
+  public getServicePlansForServiceClass(serviceClass: any) : angular.IPromise < any > {
+    let plansVersion = this.apiService.getPreferredVersion('clusterserviceplans');
+    var opts: any = undefined;
+    let serviceClassName = _.get(serviceClass, 'metadata.name', serviceClass);
+
+    // Only request service plans if the resource is available.
+    if (serviceClassName) {
```
spadgett: I think it's better to return null here than return all. I'm worried the dialog might show all plans for all service classes if we accidentally pass a null service class.

src/services/catalog.service.ts
```diff
@@ -70,6 +70,30 @@ export class CatalogService { static $inject = ['$filter', '$q', 'Constants', 'A
     return deferred.promise;
   }
 
+  // Get service plans for a service class, takes either the serviceClass object or it's name
+  public getServicePlansForServiceClass(serviceClass: any) : angular.IPromise < any > {
+    let plansVersion = this.apiService.getPreferredVersion('clusterserviceplans');
+    var opts: any = undefined;
+    let serviceClassName = _.get(serviceClass, 'metadata.name', serviceClass);
```
spadgett: I'd prefer to test `_.isString` explicitly so we don't accidentally handle values like `{}`

spadgett: Maybe

```ts
let serviceClassName = _.isString(serviceClass) ? serviceClass : _.get(serviceClass, 'metadata.name');
```

PR #538, opened 2017-10-25
Don't move "Try Sample Repository" link when clicked

### LINKED ISSUE #497: Incorrect validation error when adding to project
If you click in the "Application name" input then click "try sample" without typing, you get a validation error.

spadgett:
Only show the name error message in the create from builder dialog if
the user has typed something into the field. This avoids a problem where
clicking the "Try Sample Repository" link doesn't work because the error
message on name input blur causes the link to move.

Fixes #497

/assign @jwforres


jwforres:
lgtm but travis is failing

spadgett:
Thanks, I updated the dist

spadgett:
Passing now.

PR #541, opened 2017-10-26
Change landing page bg image

rhamilto:
screen shot 2017-10-26 at 4 28 59 pm

Assuming we get the all clear to use this image in the upstream, this will replace the product-specific PRs I opened elsewhere.


rhamilto:
@spadgett, we got the all clear from @jennyhaines to use this image.  Review, please.

jennyhaines:
FYI @spadgett @rhamilto : I finally heard back from legal, and they recommend that a copy of the Unsplash License https://unsplash.com/license also be placed in the public repo. (It is not actually required by that license, but it would put users on notice about the restriction contained in the last sentence in the license.)

rhamilto:
> FYI @spadgett @rhamilto : I finally heard back from legal, and they recommend that a copy of the Unsplash License https://unsplash.com/license also be placed in the public repo. (It is not actually required by that license, but it would put users on notice about the restriction contained in the last sentence in the license.)

Ideas on how to incorporate the license so it's clear the license covers the image (which is encoded and not a standalone file) and not the code?

PR #542, opened 2017-10-30
Preserve linebreaks in broker error messages

spadgett:
See https://github.com/openshift/origin-web-console/issues/2217

openshift web console 2017-10-30 14-45-25


spadgett:
@jim-minter FYI, thanks for improving the message from TSB

PR #544, opened 2017-11-01
Improve display of provision call failure message

jhadvig:
When creating a service that already exists we are not formatting the failure message:
failure2

So I've reused the format that we are using in [create-from-builder-results.html](https://github.com/openshift/origin-web-catalog/blob/master/src/components/create-from-builder/create-from-builder-results.html#L47-L60) and created a new `result-error` component.
Unfortunately there is a difference between how those two error types (one from `create-from-builder.controller.ts` and the second from order-service.ts) are structured, since one is returned from the origin and the other one from service-catalog. So I've created `formatError` filter which  I'm using in the *order-service.controller.ts* which parses the error message into the same format as the error from the `create-from-builder.controller.ts`.

Formatted error now looks like:
failure

@spadgett @rhamilto PTAL

### DIFF
--- a/src/components/order-service/order-service.controller.ts
+++ b/src/components/order-service/order-service.controller.ts
@@ -277,7 +277,7 @@ export class OrderServiceController implements angular.IController {
           this.ctrl.projectDisplayName = this.$filter('displayName')(project);
           this.createService();
         }, (result: any) => {
-          this.ctrl.error = result.data;
+          this.ctrl.error = this.$filter('formatError')(result.data);
         });
     } else {
       this.ctrl.projectDisplayName = this.$filter('displayName')(this.ctrl.selectedProject);
@@ -305,7 +305,7 @@ export class OrderServiceController implements angular.IController {
       if (secretName) {
         let secret = this.BindingService.makeParametersSecret(secretName, parameters, serviceInstance);
         this.DataService.create('secrets', null, secret, context).then(_.noop, (e: any) => {
-          this.ctrl.error = _.get(e, 'data');
+          this.ctrl.error = this.$filter('formatError')(_.get(e, 'data'));
         });
       }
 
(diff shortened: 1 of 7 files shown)


jhadvig:
Will generate the `dist` after review

src/components/order-service/order-service.controller.ts
```diff
@@ -473,7 +473,7 @@ export class OrderServiceController implements angular.IController {
       var readyCondition: any = _.find(conditions, {type: 'Ready'});
 
       this.ctrl.orderComplete = readyCondition && readyCondition.status === 'True';
-      this.ctrl.error = _.find(conditions, {type: 'Failed', status: 'True'});
+      this.ctrl.error = this.$filter('formatError')(_.find(conditions, {type: 'Failed', status: 'True'}));
```
jeff-phillips-18: the filter could be done once in the html rather than 4 places in the js.

src/components/result-error/result-error.component.ts
```diff
@@ -0,0 +1,10 @@
+
+export const resultError: angular.IComponentOptions = {
+  bindings: {
+    error: '=',
+    actionType: '@',
+    serviceName: '=',
+    projectName: '='
```
jeff-phillips-18: error, serviceName and projectName should be one way bindings.

src/components/result-error/result-error.component.ts
```diff
@@ -0,0 +1,10 @@
+
+export const resultError: angular.IComponentOptions = {
+  bindings: {
+    error: '=',
+    actionType: '@',
+    serviceName: '=',
```
jeff-phillips-18: Is this component only for use with service errors? Could this be any object's name?

spadgett:
@jhadvig Sorry, I think we crossed wires a bit here. #542 already adds the formatting of the error on provision failures. I think in general it's better to use CSS to preserve the newlines than try to parse messages in JavaScript or TypeScript.

We could add a component for displaying results messages in our dialogs so that we always use the same markup structure, but I'd like it to be a bit more flexible. This assumes a certain formatting of the message.

Since the original problem is already fixed, I'm inclined to leave well enough alone.

cc @jwforres

spadgett:
Talked to @jhadvig about this, and he is OK closing the PR since #542 fixed the original problem.

PR #519, opened 2017-10-20
Fix checkbox filter spacing

spadgett:
Fixes #504

screen shot 2017-10-20 at 6 24 32 pm

PR #546, opened 2017-11-02
Fix styling of pf-filter-panel input by adding Bootstrap class

rhamilto:
Before in mobile Safari where it's very noticeably different:
screen shot 2017-11-02 at 4 08 21 pm

After:
screen shot 2017-11-02 at 4 17 27 pm


spadgett:
The checkboxes should have been fixed in #519. @rhamilto Are you on current master?

spadgett:
Looks like there might still be an issue for checkboxes on Safari

rhamilto:
> The checkboxes should have been fixed in #519. @rhamilto Are you on current master?

Yep.  The checkboxes issue is a Bootstrap one.
screen shot 2017-11-02 at 5 17 31 pm

pf-filter-panel looks worse because the font size is smaller (12px vs 14px).

PR #547, opened 2017-11-03
Fall back to external name when no display name is set

spadgett:
For service classes, use the external name when there's no display name set.
This prevents the internal UID from being shown in the UI.

https://github.com/kubernetes-incubator/service-catalog/blob/master/pkg/apis/servicecatalog/v1beta1/types.go#L244-L246

Fixes:

image

@ncameronbritt @jwforres


spadgett:
Added mock data and a unit test for this. @jwforres another look?

PR #550, opened 2017-11-06
In CreateFromBuilder and OrderService wizards, correct error handling of ProjectNameTaken

dtaylor113:
Partial fix for [console#1955](https://github.com/openshift/origin-web-console/issues/1955)

image

- Immediate form validation error if project name already exists for user.  Form validation error when 'Project Name already exists' error returned from server (project name not unique across cluster).
- Form validation error cleared when user starts to type in new project name


src/components/select-project/select-project.component.ts
```diff
@@ -4,6 +4,7 @@ export const selectProject: angular.IComponentOptions = {
   bindings: {
     selectedProject: '=',
     nameTaken: '<',
+    onNewProjectNameChanged: '&',
```
spadgett: I don't adding a new callback is necessary... The embedding page can just watch `selectedProject.metadata.name`.

```ts
this.$scope.$watch('$ctrl.selectedProject.metadata.name', () => {
  this.ctrl.projectNameTaken = false;
});
```

src/components/select-project/selectProject.html
```diff
@@ -58,7 +58,8 @@ <h3 ng-if="$ctrl.canOnlyCreateProject()">Create Project</h3>
     ng-if="$ctrl.isNewProject()">
   <div class="form-group">
     <label for="name" class="control-label required">Project Name</label>
-    <div ng-class="{'has-error': ($ctrl.forms.createProjectForm.name.$error.pattern && $ctrl.forms.createProjectForm.name.$touched) || $ctrl.nameTaken}">
+    <div ng-class="{'has-error': ($ctrl.forms.createProjectForm.name.$error.pattern && $ctrl.forms.createProjectForm.name.$touched) ||
+                                 ($ctrl.nameTaken || $ctrl.forms.createProjectForm.name.$error.oscUnique)}">
```
spadgett: The parens around the last two conditions are unnecessary.

```html
 <div ng-class="{'has-error': ($ctrl.forms.createProjectForm.name.$error.pattern && $ctrl.forms.createProjectForm.name.$touched) ||
                               $ctrl.nameTaken || $ctrl.forms.createProjectForm.name.$error.oscUnique}">
```

src/components/create-from-builder/create-from-builder.controller.ts
```diff
@@ -106,6 +106,7 @@ export class CreateFromBuilderController implements angular.IController {
       onShow: this.showResults
     };
     this.ctrl.steps = [this.infoStep, this.configStep, this.bindStep, this.reviewStep];
+    this.ctrl.currentStep = "Information";
```
spadgett: Do we need to set this everywhere? I'd think it would only be necessary in the error handler when we change steps back to Configuration.

spadgett:
Is the "Create" button disabled until the user changes the name after an `AlreadyExists` error?

src/components/create-from-builder/create-from-builder.controller.ts
```diff
@@ -383,8 +390,14 @@ export class CreateFromBuilderController implements angular.IController {
       }, (e: any) => {
         this.ctrl.error = e;
       });
-    }, (result) => {
-      this.ctrl.error = result;
+    }, (e) => {
+      if (e.data.reason === 'AlreadyExists') {
+        this.ctrl.projectNameTaken = true;
+        this.ctrl.wizardDone = false;
+        this.ctrl.currentStep = "Configuration";
```
spadgett: This does this result in any flash / flicker when it changes to the results step and back?

dtaylor113: Yes it does, if that is an issue I can look into wiring up `nextCallback()` and return `false' to not navigate, not sure if that'll do the trick though.

src/components/create-from-builder/create-from-builder.controller.ts
dtaylor113: We need to set it so that when the error happens, it navigates correctly, else the currentStep will always be "Configuration".

> Is the "Create" button disabled until the user changes the name after an AlreadyExists error?

Yes

src/components/create-from-builder/create-from-builder.controller.ts
spadgett: I think that's what we do in the process-template-dialog.

dtaylor113: yes, and then we use eventing to navigate to Results upon success, just more overhead than what I'm doing at the moment, but if we want to avoid the flicker...

jeff-phillips-18: If we do this, then there should be no need to set currentStep at all.

src/components/select-project/select-project.component.ts
dtaylor113: Fixed.

src/components/select-project/selectProject.html
dtaylor113: Fixed

src/components/create-from-builder/create-from-builder.controller.ts
dtaylor113: Ok, then I believe I have addressed all review comments.  Should I create an issue about the proposed "+" next to the select project dropdown?

src/components/select-project/select-project.controller.ts
```diff
@@ -114,11 +113,6 @@ export class SelectProjectController implements angular.IController {
     }
   }
 
-  public onNewProjectNameChange() {
-    this.ctrl.nameTaken = false;
-    this.ctrl.forms.createProjectForm.name.$setValidity('nameTaken', !this.ctrl.nameTaken);
```
spadgett: Is there a reason not to set form validity? I suspect at least one of the dialogs relies on this.

PR #553, opened 2017-11-07
Change subcategory styling

rhamilto:
Supports https://github.com/openshift/origin-web-console/issues/2454

Note:  @jennyhaines requested some tweaks that make this implementation different from [her mock](https://github.com/openshift/origin-web-console/issues/2454#issuecomment-342304860).

screen shot 2017-11-07 at 2 15 09 pm
screen shot 2017-11-07 at 2 15 25 pm
screen shot 2017-11-07 at 2 15 42 pm

as seen in the console:
screen shot 2017-11-07 at 2 15 57 pm

cc: @openshift/team-ux-review

### DIFF
--- a/dist/less/services-view.less
+++ b/dist/less/services-view.less
@@ -145,8 +145,10 @@ services-view,
   }
 
   .services-items-filter {
+    border-bottom: 1px solid @color-pf-black-200;
     flex: 1 1 100%;
-    margin: (@grid-gutter-width / 2) (@grid-gutter-width / 2) (@grid-gutter-width / 4);
+    margin: (@grid-gutter-width / 4) (@grid-gutter-width / 2);
+    padding: 0 0 10px;
     div.checkbox {
       margin-bottom: 0;
       margin-top: 0;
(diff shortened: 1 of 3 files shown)


rhamilto:
/hold
found a bug

spadgett:
/hold

for @jwforres review

jennyhaines:
@rhamilto - Why are some of the subcategory icons black, and others full color? I'd recommend the full color.

spadgett:
@jennyhaines They'll be in full color. Some of the mock test data we use is only black and white.

jwforres:
My only concern is the thin blue bar at the top of the active tile. Its really hard to distinguish that bar from the background. What ends up happening is it makes the active tile look like it is not aligned correctly with the other tiles.  I've walked by @spadgett and @rhamilto 's screens a few times and every time it bothers me that it looks like it is not aligned.

It's not as bad looking at the zoomed in screenshots, but on an actual screen it just looks off.

jennyhaines:
@rhamilto  - I'm personally fine reverting back to having the blue bar down one pixel so it appears within the white card, rather than hovering 1px above. I know that you said it could look like that tile is receding, but I see what @jwforres is saying about it appearing misaligned, and visually that might be a more important problem to fix.

serenamarie125:
@jennyhaines @rhamilto what about tightening up margin above/below filter?  Wondering if that will also save a bit of space?  Unsure if it would look crammed though.  Thoughts?

serenamarie125:
@rhamilto were you able to implement the hover suggestion as well?

jennyhaines:
@serenamarie125 - the space is tightened up from the original implementation. It's now 15px above instead of (I think) 27px. I think any less padding, and it would look a little tood snug (I tried it out as small as 10px, and that didn't work.)

serenamarie125:
ok great thank @jennyhaines !

rhamilto:
> @rhamilto were you able to implement the hover suggestion as well?

Yep.  Here Java is hovered:

screen shot 2017-11-07 at 1 47 52 pm

jwforres:
we fiddled with the blue bar on @rhamilto 's screen and nothing really looks right on an actual monitor in context of the whole page, @smarterclayton @spadgett and I all agreed to nix the blue bar on the active state. Its pretty obvious what is active vs what is hovered based on which tile is connected.

jennyhaines:
@jwforres Okay, that works especially since this design may only be used for this upcoming release. The slight 2px top corners of the unselected state tiles help to differentiate them even more.

PR #568, opened 2017-11-14
Bug 1512858 - Don't request service classes and plans for builders

spadgett:
Don't try to request service classes or plans in the create from builder
dialogs when pod presets are disabled.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1512858

/kind bug
/assign @jwforres

### DIFF
--- a/src/components/create-from-builder/create-from-builder.controller.ts
+++ b/src/components/create-from-builder/create-from-builder.controller.ts
@@ -144,9 +144,6 @@ export class CreateFromBuilderController implements angular.IController {
     this.noProjectsCantCreateWatch = this.$scope.$on('no-projects-cannot-create', () => {
       this.ctrl.noProjectsCantCreate = true;
     });
-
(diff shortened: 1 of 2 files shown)


src/components/create-from-builder/create-from-builder.controller.ts
```diff
@@ -144,9 +144,6 @@ export class CreateFromBuilderController implements angular.IController {
     this.noProjectsCantCreateWatch = this.$scope.$on('no-projects-cannot-create', () => {
       this.ctrl.noProjectsCantCreate = true;
     });
-
-    this.getServiceClassesAndPlans();
-    this.instancesSupported = !!this.APIService.apiInfo(this.APIService.getPreferredVersion('serviceinstances'));
```
spadgett: There is a `if (this.ctrl.showPodPresets)` block above (line 136) that does this when pod presets are enabled. I think this was an oversight that it was left here, but it's not needed.

PR #569, opened 2017-11-27
Add ability to set project to add ordered items to.

jeff-phillips-18:


jeff-phillips-18:
Correct @dtaylor113

jeff-phillips-18:
@spadgett just to be clear, there are no required console changes for this PR. Things should still function as they currently do. Once this PR is merged, we can proceed with the console changes necessary to order items in the context of a project.

PR #572, opened 2017-11-28
Updates to catalog search to prep for the project page

jeff-phillips-18:
This is to allow the search component to be re-used in the secondary masthead in the project context pages.


src/styles/landing-page.less
```diff
@@ -36,6 +37,55 @@
     flex: 1 1 0%; // for .word-break()
     .word-break();
   }
+  .dropdown-menu {
+    margin-top: 0;
+    padding: 0;
+    width: 100%;
+    > li > a.catalog-search-match {
```
rhamilto: Could do some nesting here to avoid repeating selectors:

```
> li > a {
  &.catalog-search-match {
    ...
  &.catalog-search-show-all {
```

src/styles/landing-page.less
```diff
@@ -36,6 +37,55 @@
...
+    .catalog-search-show-none {
+      background-color: @color-pf-black-200;
+      border-color: @color-pf-black-300;
+      border-width: 0 0 1px 0;
+      padding: 5px 0;
+      text-align: center;
+    }
+  }
+  .search-pf-input-group {
+    .form-control{
```
rhamilto: Nit: space between l and {

src/styles/landing-page.less
```diff
@@ -36,6 +37,55 @@
...
+    }
+    > li > a.catalog-search-show-all {
+      background-color: @color-pf-black-200;
+      color: @link-color;
+      padding: 5px 0;
+      text-align: center;
+    }
+
+    // important needed to override styles declared important in PatternFly
+    .active > a.catalog-search-match, :focus{
```
rhamilto: I'm confused by the :focus that's not attached to an element.  Are you sure you want to override :focus styles for any descendent of .dropdown-menu?

Nit: need a space between s and {

src/styles/landing-page.less
```diff
@@ -36,6 +37,55 @@
...
+      background-color: @color-pf-blue-50 !important;
+      border-color: @color-pf-blue-100 !important;
+      color: inherit !important;
+    }
+    .active > a.catalog-search-show-all, :focus {
```
rhamilto: I'm confused by the :focus that's not attached to an element.  Are you sure you want to override :focus styles for any descendent of .dropdown-menu?  This :focus is overriding the one on line 60, no?

src/styles/landing-page.less
```diff
@@ -36,6 +37,55 @@
...
+      background-color: @color-pf-black-200;
+      border-color: @color-pf-black-300;
+      border-width: 0 0 1px 0;
+      padding: 5px 0;
+      text-align: center;
+    }
+  }
+  .search-pf-input-group {
+    .form-control{
+      font-size: (@font-size-base + 2);
```
rhamilto: @font-size-base is currently 13px in console, so the result of this calculation is going to differ between the demo app and the console.  Double check this isn't problematic.

src/styles/landing-page.less
```diff
@@ -36,6 +37,55 @@
...
+      height: 2.2em;
+    }
+    .pficon-close {
+      font-size: (@font-size-base + 4);
```
rhamilto: @font-size-base is currently 13px in console, so the result of this calculation is going to differ between the demo app and the console.  Double check this isn't problematic.

src/components/catalog-search/catalog-search.html
```diff
@@ -1,5 +1,13 @@
-<div class="catalog-search">
-  <form role="form" class="landing-search-form search-pf has-button">
+<div class="catalog-search" ng-class="{'mobile-shown': $ctrl.mobileSearchInputShown}">
+  <a href=""
```
spadgett: If we use the `btn-link` class, can this switch to a button element without too much work?

For better accessibility and to avoid problems like https://bugzilla.redhat.com/show_bug.cgi?id=1470998, @jwforres wanted to try to switch to using buttons when we can for new code.

src/components/catalog-search/catalog-search.controller.ts
```diff
@@ -56,6 +57,10 @@ export class CatalogSearchController implements angular.IController {
       this.$scope.$emit('open-overlay-panel', item);
     }
     this.ctrl.searchText = '';
+    this.ctrl.mobileSearchInputShown = false;
+    if (angular.isFunction(this.ctrl.searchToggleCallback)) {
```
spadgett: Nit: We typically use lodash when angular and lodash have the similar utilities (`_.isFunction`) unless there's a reason not to.

src/components/catalog-search/catalog-search-result.html
```diff
@@ -13,10 +13,6 @@
...
     </div>
   </div>
-  <span ng-if="match.model.id === 'viewNone'" class="catalog-search-show-none">
-    {{match.model.text}}
-  </span>
-  <span ng-if="match.model.id === 'viewAll'" class="catalog-search-show-all">
-    {{match.model.text}}  <span class="fa fa-angle-right"></span>
-  </span>
 </a>
+<a ng-if="match.model.id === 'viewAll'" href="" class="catalog-search-show-all">{{match.model.text}}  <span class="fa fa-angle-right"></span></a>
```
spadgett: Not new, but do you mind adding an aria-hidden while in the code?

```html
<span class="fa fa-angle-right" aria-hidden="true"></span>
```

src/styles/landing-page.less
jeff-phillips-18: This code was moved from below such that it only applies to item under  catalog-search not necessarily on the landing page. I questioned it as well but decided to leave it as it was. I'll investigate further.

jeff-phillips-18: Turns out the :focus is not necessary at all. It doesn't change a thing.

src/styles/landing-page.less
jeff-phillips-18: In the PR to pull this into console, we over-ride this setting to get it back to 13px. Should we just set it to 13px here rather than using the base?

src/styles/landing-page.less
jeff-phillips-18: Same here, over-ridden in console.

src/styles/landing-page.less
rhamilto: Perhaps it would be easier just to set @font-size-base to 13px in the demo app?

jeff-phillips-18: Good call.

jeff-phillips-18: These settings were correct. The override in console is for having it in the masthead in project context.

Still think its worth updating the base font size in the catalog demo app.

src/components/catalog-search/catalog-search.html
jeff-phillips-18: Updated to a button. This shows a focus border which was a bit confusing to me, so I added some code to give focus to the search input when it is toggled open.

jeff-phillips-18:
Updated to address issues above.

app/styles/variables.less
```diff
@@ -1,3 +1,4 @@
 @navbar-height: 60px;
 @landing-offset-top: (@landing-search-area-height + @navbar-height);
 @oli-font-path: "../fonts";
+@font-size-base: 13px;
```
spadgett: Roll back this change because of https://github.com/openshift/origin-web-console/pull/2540 ?

src/components/catalog-search/catalog-search.controller.ts
```diff
@@ -72,6 +80,29 @@ export class CatalogSearchController implements angular.IController {
...
+    this.ctrl.searchText = '';
+    if (_.isFunction(this.ctrl.searchToggleCallback)) {
+      this.ctrl.searchToggleCallback(this.ctrl.mobileSearchInputShown);
+    }
+    if (this.ctrl.mobileSearchInputShown) {
+      this.setSearchInputFocus(0);
+    }
+  }
+
+  private setSearchInputFocus(timeSoFar: number) {
```
spadgett: Does the `take-focus` directive work?

https://github.com/openshift/origin-web-common/blob/7b835d9f76f3ade1f0366213e8b7564fee2238e7/src/components/take-focus/takeFocus.js

jeff-phillips-18: No, I tried that first but it doesn't.

jeff-phillips-18: take-focus works when you first add a component. Since this is just hidden and shown based on the size and the toggle it won't work here.

We could use ng-if to hide and show it but that would mean having the input (with all 15 parameters) duplicated for mobile mode and also watch for window resizes to determine which input to show.

Or, we could add this kind of functionality to the take-focus directive so that a component gets focus when it is added or when it becomes visible.

app/styles/variables.less
jeff-phillips-18: done.

jeff-phillips-18:
Found one more issue that may need updating to support search in project context. 

Please hold.

jeff-phillips-18:
Thanks @spadgett wasn't sure if that was enabled for this repo

spadgett:
@jeff-phillips-18 Sure, it just adds the label for now and doesn't actually block merging. But it's nice to have the label.

jeff-phillips-18:
OK, updated and ready for final reviews/merge. 

Changing the landing page to use cached catalog items changed when the filter count for the items got computed in the servicesView component. This caused the count to get reset and was showing as 0 though there were 31 items. Changed where the filterConfig gets initialized so that the value does not get reset.

PR #575, opened 2017-11-30
<select-project> component updates

dtaylor113:
To support "Show Home Page" in http://openshift.github.io/openshift-origin-design/web-console/patterns/masthead 

image

Setting `hideCreateProject` & `hideAddToProjectLabel` to `true` results in:

image

**Also added** the ability to preselect a project in the list by specifying a `selectedProjectName` string


src/components/select-project/select-project.controller.ts
```diff
@@ -113,6 +113,12 @@ export class SelectProjectController implements angular.IController {
     }
   }
 
+  public onOpenClose (isOpen: boolean) {
+    if (isOpen) {
+      this.$scope.$emit('select-project-openned');
```
spadgett: This feels more like it should be an optional callback passed into the component instead of `$scope.$emit` (similar to `onProjectSelected`).

Curious how it's used. To select the radio when the dropdown is opened?

src/components/select-project/select-project.component.ts
```diff
@@ -2,11 +2,14 @@ import {SelectProjectController} from './select-project.controller.ts';
 
 export const selectProject: angular.IComponentOptions = {
   bindings: {
-    selectedProject: '=',
+    selectedProject: '=?',
+    selectedProjectName: '@?',
```
spadgett: I suggest `preselectProjectName` so it's clear that the value is not updated on changes.

src/components/select-project/select-project.controller.ts
```diff
@@ -277,4 +288,10 @@ export class SelectProjectController implements angular.IController {
       });
     }
   }
+
+  private getProjectByName() {
+    return _.find(this.projects, (project: any) => {
```
spadgett: You could use the shorthand for `_.find` here.

Seems strange to have a method called `getProjectByName` that doesn't take a name as a parameter. I'd probably make it general and take in the name.

```ts
private getProjectByName(name) {
  return _.find(this.projects, { metadata: { name: name } });
}
```

dtaylor113: I know, but it's already on the $scope, but I understand.  I tried to use the shorthand, with a nested obj, didn't work, so I went long hand.

spadgett: > I tried to use the shorthand, with a nested obj, didn't work, so I went long hand.

It should work: we do it in a lot of places. Can you paste the code?

spadgett: I was thinking something like `preselectProject()` would be a better abstraction with lines 253-259. Something like

```ts
private preselectProject() {
  if (this.ctrl.selectedProject) {
    return;
  }

  // if one project, default to it, else default to selected project by name (if available)
  if (_.size(this.projects) === 1) {
    this.ctrl.selectedProject = this.projects[0];
  } else if (this.ctrl.selectedProjectName) {
    this.ctrl.selectedProject = _.find(this.projects, {
      metadata: {
        name: this.ctrl.selectedProjectName
      }
    });
  }
  this.onSelectProjectChange();
}
```

src/components/select-project/select-project.controller.ts
dtaylor113: Yes, to select the radio when the dropdown is opened.  I can switch to a callback.

dtaylor113:
Hi @spadgett, I believe I have addressed all your review comments.
Thanks,
Dave

PR #587, opened 2017-12-06
Do not store categories or vendors in Catalog service

### LINKED ISSUE #586: Catalog Service returns stale values for categories and vendors
The public objects for categories and vendors are created when the getCatalogItems or the getProjectCatalogItems methods are called. If one expects the categories to be those for one or the other but the alternate had been called since retrieval, the values returned will be incorrect.

jeff-phillips-18:
Remove the determination of categories and vendors from the calls to get catalog items (either project or all). Rather, add methods to determine and return these based on a list of catalog items.

Fixes #586

### DIFF
--- a/dist/origin-web-catalogs.js
+++ b/dist/origin-web-catalogs.js
@@ -49,6 +49,7 @@ webpackJsonp([ 0, 1 ], [ function(e, t) {
     t.catalogFilter = {
         bindings: {
             config: "<",
+            vendors: "<",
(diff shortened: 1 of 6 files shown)


jeff-phillips-18:
Added method to sort catalog items to the catalog service.

jeff-phillips-18:
Required by https://github.com/openshift/origin-web-console/pull/2563

PR #590, opened 2017-12-08
Version bump to 3.9.0-alpha.0

jeff-phillips-18:


bower.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "origin-web-catalog",
-  "version": "0.0.68",
+  "version": "0.0.69",
```
spadgett: We're switching to `v3.9.0-alpha.0` (1, 2, 3...) for releases as in common

https://github.com/openshift/origin-web-common/releases/tag/v3.9.0-alpha.0

jeff-phillips-18: OK, thanks. I'll update

jeff-phillips-18:
Should I bump the common version while I'm in there?

spadgett:
Sure. I think @benjaminapetersen is about to cut a v3.9.0-alpha.1 for common

bower.json
```diff
@@ -40,7 +40,7 @@
     "patternfly": ">=3.29.3 <4.0.0",
     "jquery": "~3.2.1",
     "lodash": "~4.17.4",
-    "origin-web-common": "~0.0.70",
+    "origin-web-common": "v3.9.0-alpha.1",
```
spadgett: I wonder if we can still specify a range for prerelease versions. Does something like `>=v3.9.0-alpha.1 <3.10.0` work?

spadgett: Seems to work for npm at least testing with https://semver.npmjs.com/

jeff-phillips-18:
Updated to use a range.

bower.json
```diff
@@ -40,7 +40,7 @@
     "patternfly": ">=3.29.3 <4.0.0",
     "jquery": "~3.2.1",
     "lodash": "~4.17.4",
-    "origin-web-common": "~0.0.70",
+    "origin-web-common": ">=v3.9.0-alpha.1 <3.10.0",
```
benjaminapetersen: +1 def prefer the range, thx!

PR #592, opened 2017-12-11
Fixes for build warnings, failure with node 6.9.0

### LINKED ISSUE #591: npm run start fails with error message
The following error is displayed:

```bash 
ERROR in ./src/components/update-service/update-service.controller.ts
(246,63): error TS2345: Argument of type 'string[][]' is not assignable to parameter of type 'Many<PropertyName>'.
  Type 'string[][]' is not assignable to type 'PropertyName[]'.
    Type 'string[]' is not assignable to type 'PropertyName'.
      Type 'string[]' is not assignable to type 'symbol'.

ERROR in ./src/components/update-service/update-service.controller.ts
(247,45): error TS2345: Argument of type 'string[][]' is not assignable to parameter of type 'Many<PropertyName>'.
  Type 'string[][]' is not assignable to type 'PropertyName[]'.
```

I haven't tried to run the catalog standalone in a good while, so I'm not sure if this is a legitimate error or if standalone serve has fallen by the wayside (ie, works fine if running with `bower link` to console).  

@jeff-phillips-18 @spadgett @jwforres

jeff-phillips-18:
Fixes #591

### DIFF
--- a/app/app.ts
+++ b/app/app.ts
@@ -67,7 +67,7 @@ angular
   .component('logoutpage', logoutPage)
   .component('navigation', navigation)
   .constant('amTimeAgoConfig', {titleFormat: 'LLL'})
-  .run(function(IS_IOS) {
+  .run(function(IS_IOS: boolean) {
     if (IS_IOS) {
       // Add a class for iOS devices. This lets us disable some hover effects
       // since iOS will treat the first tap as a hover if it changes the DOM
--- a/app/mockServices/mockBinding.service.ts
+++ b/app/mockServices/mockBinding.service.ts
@@ -30,7 +30,7 @@ export class BindingService implements IBindingService {
   }
 
   public makeParametersSecret(secretName: any, parameters: any, owner: any) {
-    var secret = {
+    var secret: any = {
       apiVersion: 'v1',
       kind: 'Secret',
       metadata: {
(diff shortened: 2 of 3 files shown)


benjaminapetersen:
ha! good ol' any 😄  

i need to get into TS more.

app/app.ts
```diff
@@ -67,7 +67,7 @@ angular
   .component('logoutpage', logoutPage)
   .component('navigation', navigation)
   .constant('amTimeAgoConfig', {titleFormat: 'LLL'})
-  .run(function(IS_IOS) {
+  .run(function(IS_IOS: any) {
```
benjaminapetersen: Ignorance question, but is there a better annotation than `any` for things like this? It appears that `IS_IOS` is a `bool` so annotating with type `bool` would make sense.  I could be wrong, tho.

I've been wondering about similar things like `DataService: any`.  What is best practice for an annotation that should resolve to a specific service singleton?

jeff-phillips-18: I could use boolean here.

jeff-phillips-18: Largely we have ignored typing in this repo. I've added issue #593 to address this concern.

benjaminapetersen: Yeah, there is a lot of `any`.  I'd def like to see a file or two that somewhat follows best practices, so that if we wanted to fix it we could.

benjaminapetersen:
Nice. Need to cut a release now?

jeff-phillips-18:
I don't believe this warrants a release.

benjaminapetersen:
Yup, pulled the change fine w/o it.

PR #594, opened 2017-12-12
Update create-from-builder component to use getPreferredVersion

benjaminapetersen:


benjaminapetersen:
First of a small handful of these to update in catalog.

PR #596, opened 2017-12-12
[WIP] Update order-service component use getPreferredVersion

benjaminapetersen:


benjaminapetersen:
wups, did this once 2x, before and after lunch.  closing in favor of #597 because it has the update to line `313`.

PR #602, opened 2017-12-15
Refactor services-view styles for within a project

rhamilto:
Supports https://github.com/openshift/origin-web-console/issues/2592

localhost_9000_dev-console_project_myproject_catalog ipad pro
localhost_9000_dev-console_project_myproject_catalog iphone 8 plus
localhost_9000_dev-console_project_myproject_catalog iphone 8 plus 1
localhost_9000_dev-console_project_myproject_catalog iphone 8 plus 2

cc: @jennyhaines

### DIFF
--- a/dist/less/services-view.less
+++ b/dist/less/services-view.less
@@ -254,18 +254,25 @@ services-view,
           position: absolute;
           right: 0;
           top: 0;
+          // so that SVGs are sized correctly in IE
+          &[src$=".svg"] {
+            width: 36px;
+          }
         }
       }
     }
   }
 
   .services-view-container {
-    background-color: fade(@color-pf-blue-700, 60%);
     display: flex;
     flex: 1 1 auto;
     flex-direction: column;
-    width: 100%; // IE11
     position: relative;
+    width: 100%; // IE11
+
+    .landing & {
+      background-color: fade(@color-pf-blue-700, 60%);
+    }
 
     .add-methods {
       align-items: center;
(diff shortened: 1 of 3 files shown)


jennyhaines:
@rhamilto - I think this looks great, except I just noticed that the "< Back" link on the 3rd and 4th image don't align to anything. I would think that it would be aligned directly under the words "My Project", above it. I'm seeing the same thing on the second image with "Select an item...". I would think that would be aligned under "My Project".

rhamilto:
> @rhamilto - I think this looks great, except I just noticed that the "< Back" link on the 3rd and 4th image don't align to anything. I would think that it would be aligned directly under the words "My Project", above it. I'm seeing the same thing on the second image with "Select an item...". I would think that would be aligned under "My Project".

Good eye, @jennyhaines!  I wrestled with that and couldn't find a good option.  I'll make your proposed change and update screenshots.

sg00dwin:
@rhamilto to get rid of the transition position movement adjust these rules 

```
.services-inner-container {
  position: relative;
}
.services-view .services-view-container .services-no-sub-categories.ng-enter, .services-view .services-view-container .services-no-sub-categories.ng-enter-prepare, .services-view .services-view-container .services-sub-categories.ng-enter, .services-view .services-view-container .services-sub-categories.ng-enter-prepare {
  top: 0;
}

.services-view .services-no-sub-categories {
    // make top padding 15px to match => .services-view .services-no-sub-categories 
}

```

rhamilto:
Now incorporates fixes from @sg00dwin.  @sg00dwin, mind taking another look?

rhamilto:
@sg00dwin is working on an animation fix, but I'm gonna merge this now and handle the animation fix as a follow-on.

PR #603, opened 2017-12-15
Fix bug where pficons do not display

rhamilto:
Requires https://github.com/openshift/origin-web-common/pull/270
Partial fix for https://github.com/openshift/origin/issues/17745


spadgett:
@rhamilto See my comment in the other PR... I think we need a common filter for this since we use it in so many places (at least 7?)

rhamilto:
Now utilizes normalizeIconClass filter from origin-web-common#v3.9.0-alpha.3.

rhamilto:
@benjaminapetersen, mind giving this a review?

benjaminapetersen:
@rhamilto yup, pulling it down to take a look

PR #605, opened 2017-12-15
fix(overlay-panel): Fix overlay-panel escape key event handler.

### LINKED ISSUE #604: overlay-panel is dismissed by escape key when another modal is on top of it
When an overlay-panel is open and something like a confirmation modal is open on top of it, pressing the escape key causes both to be dismissed.

TheRealJon:
## Description
Remove `keypress` event binding from overlay panel controller and add a check in `keydown` event handler to ensure preventDefault() has not already been called on the event. This ensures that if another handler has already handled the escape keydown event and explicitly prevented default behavior, then the overlay panel will not be dismissed.

Fixes #604

/kind bug

### DIFF
--- a/src/components/overlay-panel/overlay-panel.controller.ts
+++ b/src/components/overlay-panel/overlay-panel.controller.ts
@@ -43,21 +43,26 @@ export class OverlayPanelController implements angular.IController {
   private showDialog = () => {
     this.ctrl.shown = true;
     $('body').addClass('overlay-open');
-    this.$document.bind('keydown keypress', this.closeOnEsc);
+    this.$document.on('keydown', this.closeOnEsc);
   };
 
   private hideDialog = () => {
     this.ctrl.shown = false;
     $('body').removeClass('overlay-open');
-    this.$document.unbind('keydown keypress', this.closeOnEsc);
+    this.$document.off('keydown', this.closeOnEsc);
   };
 
   private closeOnEsc = (event: any) => {
     if (event.which === 27) {
-      event.preventDefault();
-      this.$scope.$evalAsync(() => {
-        this.closePanel();
-      });
+      // Only close this overlay if another handler has not called 
+      // preventDefault() on this event. This means only one overlay/modal will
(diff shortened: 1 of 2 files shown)


src/components/overlay-panel/overlay-panel.controller.ts
```diff
@@ -43,21 +43,26 @@ export class OverlayPanelController implements angular.IController {
...
   private closeOnEsc = (event: any) => {
     if (event.which === 27) {
-      event.preventDefault();
-      this.$scope.$evalAsync(() => {
-        this.closePanel();
-      });
+      // Only close this overlay if another handler has not called 
+      // preventDefault() on this event. This means only one overlay/modal will
+      // be closed per escape keydown event.
+      if(!event.isDefaultPrevented()) {
```
spadgett: We should double check browser support here. Is this a jQuery event object? If so, it should be OK

https://api.jquery.com/event.isDefaultPrevented/#event-isDefaultPrevented

The standard is `event.defaultPrevented`, though.

https://developer.mozilla.org/en-US/docs/Web/API/Event/defaultPrevented

I'm assuming `$document.bind` will pass a jQuery event. `bind()` seems to be deprecated in jQuery 3.0 incidentally.

http://api.jquery.com/bind/

nit: space between `if` and `(`

spadgett (commented):
Thanks @TheRealJon, just one question

TheRealJon:
Regarding the `event.isDefaultPrevented()` it looks like jQuery just uses the javascript event's `defaultPrevented` property.

https://github.com/jquery/jquery/blob/3.2.1/src/event.js#L527

TheRealJon:
@spadgett I removed the deprecated jQuery `bind()` and `unbind()` functions and replaced them with the equivalent jQuery `on()` and `off()` functions. They do pass a jQuery event object to the handler, so using `event.isDefaultPrevented()` is probably the best bet. It actually looks like jQuery provides more cross-browser compatibility than the standard js flag.

spadgett (approved):

PR #607, opened 2017-12-19
Add `npm run watch` command to watch /src

benjaminapetersen:
@rhamilto @sg00dwin this should fix the lack of a `npm run watch` for the catalog.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,8 @@
     "test:coverage": "npm run test && npm run coverage",
     "test:watch": "karma start --auto-watch --no-single-run",
     "docs": "jsdoc src/app -r -d docs",
-    "docs:launch": "npm run docs && live-server --port=3002 --open=docs/"
+    "docs:launch": "npm run docs && live-server --port=3002 --open=docs/",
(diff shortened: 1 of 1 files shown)


rhamilto:
This does not appear to result in new dist.  I ran `npm run watch` and changed src/styles/services-view.less and src/components/services-view/services-view.html, but no new dist was generated.

benjaminapetersen:
I tested with .just, .html, .less. it took a while at times and the cli was
not immediately responsive. Also, I deleted /dist often to be sure it was
working.

I can give it another look

benjaminapetersen:
@rhamilto tested changing `src/styles/services-view.less`, which did cause a `/dist` change, however, it takes about 20+ seconds to show.  

Per [this issue with webpack 2](https://github.com/webpack/webpack/issues/1922), messing around with setting `resolve.unsafeCache: true` in the config to speed things up a bit.  So far, doesn't seem to help.

Tried recording a `.gif` to show you, but recording fails due to the stated slowness.

benjaminapetersen:
Guessing there are other optimizations that need to happen to our `webpack.conf.js` in order to speed this up.  Will take some research.

PR #608, opened 2017-12-19
Fine tuning of positioning and alignment during transitions.

sg00dwin:
- removes ng-leave transition to prevent 'services-no-sub-categories' from briefly showing when markup is present.

### DIFF
--- a/src/styles/services-view.less
+++ b/src/styles/services-view.less
@@ -65,14 +65,6 @@ services-view,
       }
     }
 
-    &.ng-leave {
-      opacity: 1;
-      transition: 200ms ease-in opacity;
-      &.ng-leave-active {
-        opacity: 0;
-      }
-    }
-
     .services-item-icon {
       &:after {
         border: 2px solid transparent;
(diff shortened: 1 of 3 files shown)


sg00dwin:
@rhamilto ready for review

rhamilto:
Now the landing page at resolutions less than 768px has the bug when navigating between categories.  :(

http://recordit.co/lBtPr2OyIZ

rhamilto:
And the landing page at resolutions above 768px has expansion bug of items with no subcategories.
http://recordit.co/lAaJSeaOyq

sg00dwin:
@rhamilto ptal

src/styles/services-view.less
```diff
@@ -156,11 +148,18 @@ services-view,
     }
   }
 
+  .services-no-sub-categories {
+  	background:  @color-pf-white; // so background extends when full height when default to show all
```
rhamilto: Nit:  spacing

src/styles/services-view.less
```diff
@@ -189,6 +191,7 @@ services-view,
       background-color: rgba(255,255,255,.9);
       border-radius: 2px 2px 0 0;
       display: inline-block;
+      float: left;
```
rhamilto: Might be good to add a comment here as the combo of display: inline-block and float makes me suspicious.  Why is the float needed?

PR #610, opened 2018-01-02
Close panel when navigating to advanced options

spadgett:
Don't assume changing locations will cause the panel to close.

See https://github.com/openshift/origin-web-console/issues/2624

/kind bug
/assign @jwforres


spadgett:
Travis failed on npm install. Closing and reopening to kick the tests again.

PR #612, opened 2018-01-03
Services View: Adjusted 'Clear Filters' link in filter empty state to not redirect

### LINKED ISSUE #601: "Clear filters" link within empty state results always redirects to landing page, even if you're not on the landing page
When clicking the "Clear filters" link within the empty state results, redirection to the landing page always occurs.  If I'm clicking the link when in a project, I would expect to stay within the project.

screen shot 2017-12-14 at 4 16 15 pm

dtaylor113:
Fixes #601


spadgett:
@dtaylor113 Should this be fixed in angular-patternfly? I don't think angular-patternfly should treat a missing URL as if it's a link.

dtaylor113:
Ok, I'll handle it in angular-patternfly

PR #618, opened 2018-01-12
[WIP] Fix duplicate keyword filter bug

### LINKED ISSUE #617: Same keyword can be set multiple times
Search for "ruby" from the search box in the nav.  Then filter by "ruby" using the filter by keyword input in the filter bar.  These are the same filter and should only show one badge.  Note clearing one of the badges clears both.

screen shot 2018-01-10 at 10 19 22 am

TheRealJon:
- Transform all keywords to lowercase before they are added to filters (from filter panel or top search bar)
- Check for duplicates before adding a new keyword filter

Fix #617


TheRealJon:
Closing PR. Issue will be fixed in angular-patternfly. See https://github.com/patternfly/angular-patternfly/issues/704

PR #619, opened 2018-01-16
Made ng-required for select-project dropdown configurable

dtaylor113:
Added new `projectSelectionRequired` property which determines if the `<select-project>` dropdown is required or not.  Allows parent component to determine if the `<select-project>` dropdown/form is required or not.
Needed for Issue: https://bugzilla.redhat.com/show_bug.cgi?id=1534858


src/components/select-project/select-project.component.ts
```diff
@@ -10,7 +10,8 @@ export const selectProject: angular.IComponentOptions = {
     availableProjects: '<?',
     showDivider: '<?',
     hideCreateProject: '<?',
-    hideLabel: '<?'
+    hideLabel: '<?',
+    projectSelectionRequired: '<?'
```
spadgett: Can we use `isRequired` or something shorter?

dtaylor113: Changed to `isRequired`

spadgett:
OK to merge when travis passes

PR #625, opened 2018-01-25
Added pfEmptyState when no catalog items in Services View

dtaylor113:
Previously, when there are no catalog items, we showed a simple text message:

image

**Implemented Empty State:**

_Catalog Home/Landing Page:_
image

_Project's Browse Catalog Page:_
image

Related Issue:  https://github.com/openshift/origin-web-console/issues/2496


src/components/services-view/services-view.controller.ts
```diff
@@ -71,6 +71,15 @@ export class ServicesViewController implements angular.IController {
       }
     };
 
+    this.ctrl.noItemsConfig = {
+        title: 'No items.',
+        info: 'There are no catalog items. Please see:',
```
rhamilto: To be consistent with other empty state text, I propose we say:
"No catalog items have been loaded."

And the helpLink:
"See Loading the Default Image Streams and Templates".

src/styles/services-view.less
```diff
@@ -134,6 +134,12 @@ services-view,
     padding-bottom: 30px; // give equal space below last row of service-items
     width: 100%; // IE11 needs width associated with flex
     z-index: 9; // elevate so that toggling tiles that cause repositioning do not overlay main content block
+    &.no-items-available {
+      display: block;
+      .pf-empty-state {
+        margin: 0;
```
rhamilto: This is good at all screen sizes (e.g., 320px)?

src/components/services-view/services-view.controller.ts
dtaylor113: Updated text and link

src/styles/services-view.less
dtaylor113: LGTM
image

rhamilto: What's adding the period after "Templates"?  pf-empty-state?  The period looks like a mistake.

src/components/services-view/services-view.controller.ts
```diff
@@ -71,6 +71,14 @@ export class ServicesViewController implements angular.IController {
       }
     };
 
+    this.ctrl.noItemsConfig = {
+        title: 'No catalog items have been loaded.',
```
rhamilto: Sorry.  I think you still want the title to be "No items." with the info being "No catalog items have been loaded." followed by the help link.  

Another example for reference:  https://user-images.githubusercontent.com/895728/31667665-20f0674a-b31e-11e7-8e83-8b0a047220ba.png

dtaylor113: Ok, updated to:
image

src/styles/services-view.less
dtaylor113: The period is being added in a-pf repo, I'll fix this upstream.

rhamilto:
lgtm assuming Travis passes.

PR #627, opened 2018-01-29
Catalog Filter: prevent duplicate keyword filters

dtaylor113:
Prevent duplicate catalog keyword filters from being applied.   Case insensitive.   Upon entering a duplicate keyword, the existing keyword filter tag/chip "flashes":

dup-keywords

Related Issue: https://github.com/openshift/origin-web-catalog/issues/617

### DIFF
--- a/src/components/catalog-filter/catalog-filter.controller.ts
+++ b/src/components/catalog-filter/catalog-filter.controller.ts
@@ -2,15 +2,17 @@ import * as angular from 'angular';
 import * as _ from 'lodash';
 
 export class CatalogFilterController implements angular.IController {
-  static $inject = ['$scope', 'Catalog'];
+  static $inject = ['$scope', '$timeout', 'Catalog'];
 
   public ctrl: any = this;
   private $scope: any;
+  private $timeout: any;
   private Catalog: any;
   private removeClearFilterListener: any;
 
-  constructor($scope: any, Catalog: any) {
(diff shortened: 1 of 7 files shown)


src/components/catalog-filter/catalog-filter.controller.ts
```diff
@@ -91,6 +98,29 @@ export class CatalogFilterController implements angular.IController {
...
+  private highlightKeywordFilter = (keyword: string) => {
+      let filterTags: any = document.querySelectorAll('.active-filter.label.label-info.single-label');
+
+      let existingKeyword: any = _.find(filterTags, (tag: any) => {
+        return tag.innerText.trim() === ("Keyword: " + keyword.toLowerCase());
+      });
+
+      if (existingKeyword) {
+          this.$timeout(() => {
+              existingKeyword.style.opacity = 0;
```
rhamilto: We should probably add and remove classes here to trigger the animation rather than manipulating the style object directly in a manner similar to the way Angular does.

dtaylor113: Hi, changed to add/remove animation class.

src/styles/animations.less
```diff
@@ -34,3 +34,8 @@
     transform: translateY(-50px);
   }
 }
+
+@keyframes flash {
```
rhamilto: Nit:  alpha.  This animation should come first in the file.

PR #631, opened 2018-01-31
Catalog Filter Panel: better logic to find duplicate filter result tag to flash

dtaylor113:
So, previous iteration of this worked in the Catalog demo app., but not in Console  -I did the newbie mistake of not 'bower linking' my catalog dev code to console and making sure it actually ran in Console.  

This update makes the `querySelectorAll` more specific, while at the same time making the `_find the dup keyword filter tag/result_' matching logic more generic.

### DIFF
--- a/src/components/catalog-filter/catalog-filter.controller.ts
+++ b/src/components/catalog-filter/catalog-filter.controller.ts
@@ -105,10 +105,10 @@ export class CatalogFilterController implements angular.IController {
   };
 
   private highlightKeywordFilter = (keyword: string) => {
-      let filterTags: any = document.querySelectorAll('.active-filter.label.label-info.single-label');
+      let filterTags: any = document.querySelectorAll('pf-filter-panel-results .label-info');
 
       let existingKeyword: any = _.find(filterTags, (tag: any) => {
-        return tag.innerText.trim() === ("Keyword: " + keyword.toLowerCase());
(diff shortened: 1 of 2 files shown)


src/components/catalog-filter/catalog-filter.controller.ts
```diff
@@ -105,10 +105,12 @@ export class CatalogFilterController implements angular.IController {
...
 
   private highlightKeywordFilter = (keyword: string) => {
-      let filterTags: any = document.querySelectorAll('.active-filter.label.label-info.single-label');
+      let filterTags: any = document.querySelectorAll('pf-filter-panel-results .label-info');
 
       let existingKeyword: any = _.find(filterTags, (tag: any) => {
-        return tag.innerText.trim() === ("Keyword: " + keyword.toLowerCase());
+        let filterTagText: string = tag.innerText.trim();
+        return _.includes(filterTagText, "Keyword:") &&
+               _.includes(filterTagText, keyword.toLowerCase());
```
spadgett: If `filterTagText` is a string, won't this match substrings? For instance, "Keyword:foobar" includes "foo"

dtaylor113: Ok, changed to be more restrictive.

dtaylor113: Hi @spadgett, think this is ready to go :-)

PR #634, opened 2018-02-05
Resolve bug where overlay wizard scrolling is sluggish

rhamilto:
Supports https://github.com/openshift/origin-web-console/issues/2761

Note:  .landing and .landing-side-bar benefit from this same fix;
however, the application of the fix causes .landing-side-bar to
disappear as -webkit-overflow-scrolling: touch changes the behavior of
position: fixed.  I will work to see if I can find a workaround and
open a separate PR.


spadgett:
Feel free to merge when travis passes

PR #638, opened 2018-02-05
Bug 1542669 - Don't show builder version number on first wizard page

spadgett:
This partially addresses a usability problem where it's not clear on the first page that you can select different versions. The problem is compounded because https://github.com/openshift/origin/issues/17552 means the oldest is now selected by default, so it looks like you're stuck with an old version of something. This PR removes the version and just shows the image stream display name on the info step and in the wizard title. 

Note there is still an issue because the descriptions are different per version (image stream tag) and might contain a version number. Open to input on how to address that.

/assign @rhamilto 
/cc @serenamarie125 @beanh66 @sspeiche @bparees 

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1542669


rhamilto:
Yuck.  *facepalm*

spadgett:
Fixing the sort order in #639 so the newest version is selected by default.

spadgett:
/assign @dtaylor113 

@dtaylor113 PTAL

spadgett:
/unassign @rhamilto 

Robb is on PTO.

@dtaylor113 Quick review? This is only a couple lines.

PR #643, opened 2018-02-12
Sort image stream tags by semantic version when possible

spadgett:
/assign @benjaminapetersen


spadgett:
Updated tests for Patternfly filter changes. @dtaylor113 fyi

PR #645, opened 2018-02-15
Show APB Version in Wizard Info Step

### LINKED ISSUE #624: Show APB version when clicking on item
It would be very useful to be able to see the version of APBs before deploying a service. Somewhere here:

screen shot 2018-01-24 at 11 19 11

It is very annoying to deploy the whole thing just to see that I don't have the APB I expected.

alecmerdler:
Fixes #624

### DIFF
--- a/src/components/create-from-builder/create-from-builder-info.html
+++ b/src/components/create-from-builder/create-from-builder-info.html
@@ -16,6 +16,7 @@
           {{tag}}
         </span>
       </div>
+      <span>Version {{$ctrl.istag.annotations.version}}</span>
       <ul ng-if="$ctrl.documentationUrl || $ctrl.supportUrl" class="list-inline order-service-documentation-url">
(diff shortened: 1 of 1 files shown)


src/components/create-from-builder/create-from-builder-info.html
```diff
@@ -16,6 +16,7 @@
           {{tag}}
         </span>
       </div>
+      <span>Version {{$ctrl.istag.annotations.version}}</span>
```
spadgett: Most image streams won't have a version annotation, so this will just show up as "Version " with no value in most cases.

Did you mean to add this to the create from builder form? It looks like the issue has a screenshot of a service class, not a builder image.

spadgett:
@alecmerdler Thanks for the PR. We probably need to check with @jwmatthews to know how the ansible service broker advertises the APB version.

PR #646, opened 2018-02-19
Resolve bug where repaint bug occurs in Chrome on retina displays

rhamilto:
Before:
http://g.recordit.co/TsJrOYg2b8.gif

After:
http://g.recordit.co/jq2toEwUpk.gif


spadgett:
OK to merge when travis passes

PR #657, opened 2018-03-05
Bug 1549450 - Better to use the same name for Clear filters, Clear Filter, Clear All Filters

jhadvig:
Catalog version of https://github.com/openshift/origin-web-console/pull/2870


spadgett:
Thanks, @jhadvig. Will need dist files.

PR #658, opened 2018-03-06
introducing 3scale in the middleware integration list

maneta:
Adding 3scale to the middleware list.

I'm assuming here that the `tags` array will match with the OS template tags.

cc @spadgett

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,6 +110,7 @@
     "url-loader": "^0.5.7",
     "webpack": "2.2.0",
     "webpack-dev-server": "^2.9.2",
-    "webpack-merge": "^3.0.0"
+    "webpack-merge": "^3.0.0",
+    "bower": "^1.8.2"
   }
 }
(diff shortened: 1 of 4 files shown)


spadgett:
> I'm assuming here that the tags array will match with the OS template tags.

That's right.

spadgett:
Thanks, you'll need to build the dist files and amend your commit to add them. See

https://github.com/openshift/origin-web-catalog#contributing

maneta:
@spadgett done the rebuild. thanks.

package.json
```diff
@@ -36,6 +36,7 @@
     "angular-schema-form-bootstrap": "^0.2.0",
     "angular-ui-bootstrap": "0.14.x",
     "bootstrap": "^3.3.6",
+    "bower": "^1.8.2",
```
spadgett: This should only be under `devDependencies`.

maneta: done in f56c3e74a924c61f4d2df236f5d00915cb971dac

PR #660, opened 2018-03-08
Bump npm dependencies

spadgett:
/assign @benjaminapetersen


spadgett:
> Just the package-lock.json file, no change to package.json?

Right. We're just picking up the latest for our `^` and `~` dependencies.

PR #661, opened 2018-03-12
Remove hawtio-core dependency

TheRealJon:
Removes hawtio-core dependency. Dependent on https://github.com/openshift/origin-web-common/pull/316.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -40,7 +40,7 @@
     "patternfly": ">=3.29.3 <4.0.0",
     "jquery": "~3.2.1",
     "lodash": "~4.17.4",
-    "origin-web-common": ">=v3.10.0-alpha.0 <3.11.0",
+    "origin-web-common": ">=3.10.0-alpha.1 <3.11.0",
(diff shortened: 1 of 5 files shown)


bower.json
```diff
@@ -40,7 +40,7 @@
     "patternfly": ">=3.29.3 <4.0.0",
     "jquery": "~3.2.1",
     "lodash": "~4.17.4",
-    "origin-web-common": ">=v3.10.0-alpha.0 <3.11.0",
+    "origin-web-common": "^3.10.0-alpha.1",
```
spadgett: Caret will match 3.11.x, which I don't think we want. I'd prefer to go with the previous range we had and just updating the last number.

spadgett (commented):
LGTM except the one comment. Please squash

bower.json
benjaminapetersen: @TheRealJon the range lets us specify an exact version in `origin-web-console`, `origin-web-catalog` should defer rather than compete.

spadgett (approved):

Issue #593, opened 2017-12-11
The web catalog project should use real types in typescript


jeff-phillips-18:
Lots and lots of definitions that simply use 'any' rather than the appropriate types. It would be beneficial to go through and replace as many 'any' declarations with the correct type.

benjaminapetersen:
Just adding a comment here as well, I'd love to see a file or two that follow best practices for `bool`, `sting`, etc, as well as what to do about `interfaces` or functions that require a specific singleton, such as `DataService`.  

I don't mind working on tech debt items with mini-PRs on the side, but I'm pretty fresh to TS yet so if anyone has a good idea about how to do this & could kick off a small PR, that would def be helpful.

dtaylor113:
I don't know about the more advanced topics you mentioned, but this is a good reference for the basic types:   https://www.typescriptlang.org/docs/handbook/basic-types.html.

dtaylor113:
The Catalog app. uses some interesting types  `public scope: ng.IScope;`:
https://github.com/openshift/origin-web-catalog/blob/master/test/utils/ComponentTest.ts#L8

benjaminapetersen:
Yeah thats helpful.  `ICompileService` implies you prob have to do an interface `IDataService` to require `DataService`, I wonder if there is away to get around boilerplate interfaces for singletons.

PR #662, opened 2018-03-12
Bug 1449038 - Improve catalog search

spadgett:
https://trello.com/c/2fFSJMX1
https://bugzilla.redhat.com/show_bug.cgi?id=1449038

Requires https://github.com/openshift/origin-web-common/pull/317

### DIFF
--- a/src/components/catalog-search/catalog-search.controller.ts
+++ b/src/components/catalog-search/catalog-search.controller.ts
@@ -3,11 +3,12 @@ import * as _ from 'lodash';
 import * as $ from 'jquery';
 
 export class CatalogSearchController implements angular.IController {
-  static $inject = ['$rootScope', '$scope', '$timeout', '$q', 'Catalog', 'KeywordService'];
+  static $inject = ['$rootScope', '$scope', '$timeout', '$q', 'Catalog', 'Constants', 'KeywordService'];
 
   public ctrl: any = this;
 
   private Catalog: any;
+  private Constants: any;
   private KeywordService: any;
   private $rootScope: any;
   private $scope: any;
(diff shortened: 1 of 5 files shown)


spadgett:
cc @rhamilto

spadgett:
/assign @benjaminapetersen 

@benjaminapetersen PTAL, the common bump will happen in #661

src/components/catalog-search/catalog-search.controller.ts
```diff
@@ -103,9 +105,9 @@ export class CatalogSearchController implements angular.IController {
     }
   }
```
benjaminapetersen: Just thinking as I read this, if we aim to avoid `:any` for the type & try to be more specific, should we also try to use `const` as much as possible (instead of `let`), for variables that should not be mutated?  

Nothing to change in this PR, necessarily.

spadgett: +1 to using const when we can. I don't think we take good advantage of ES6 / Typscript. Let me know if there are places you see I should have done that.

src/components/catalog-search/catalog-search.controller.ts
```diff
@@ -19,12 +20,13 @@ export class CatalogSearchController implements angular.IController {
   // Used when the user starts typing before the items have loaded.
   private searchDeferred: ng.IDeferred<any[]>;
 
-  constructor($rootScope: any, $scope: any, $timeout: any, $q: any, Catalog: any, KeywordService: any) {
+  constructor($rootScope: any, $scope: any, $timeout: any, $q: any, Catalog: any, Constants: any, KeywordService: any) {
```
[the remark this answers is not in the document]
spadgett: I think the basic problem here and elsewhere is KeywordService comes from common, which is not Typescript. So we don't have the interface (except for the mock service, which is incomplete and also unfortunately duplicates a lot of the same code from common).

I agree we're not using types well. This is a general problem in the catalog repo. We have #593 to track.

IMO, we should decide if we really want Typescript or not and go all-in. We're in a weird space in between where we're not seeing the advantages. We're at an inflection point where we could look at doing that since we're talking about moving to React.

PR #668, opened 2018-03-28
Don't return catalog items from shared `openshift` namespace

spadgett:
Items in the openshift namespace are already returned from the
getCatalogItems() method. Don't return them in getProjectCatalogItems(),
otherwise they will appear in the catalog twice.

See https://github.com/openshift/origin-web-console/issues/2928
/assign @benjaminapetersen

### DIFF
--- a/src/services/catalog.service.ts
+++ b/src/services/catalog.service.ts
@@ -117,6 +117,15 @@ export class CatalogService {
   }
 
   public getProjectCatalogItems(projectName: string, includeImages: boolean = true, includeTemplates: boolean = true, partialObjectMetadataList: boolean = false ) {
+    // Never fetch items from the `openshift` namespace as project items since
+    // they are already included in the `getCatalogItems()` response. Otherwise
+    // they'll show up twice in the catalog.
+    if (projectName === 'openshift') {
+      let items: any[] = [];
+      let errors = '';
+      return this.$q.when([ items, errors ]);
+    }
+
     let deferred = this.$q.defer();
     let catalogItems: any = {
       imageStreams: [],
--- a/dist/origin-web-catalogs.js
+++ b/dist/origin-web-catalogs.js
@@ -394,11 +394,11 @@ webpackJsonp([ 0, 1 ], [ function(e, t) {
             } ]
         } ];
         n.set(window, "OPENSHIFT_CONSTANTS.SERVICE_CATALOG_CATEGORIES", s), n.set(window, "OPENSHIFT_CONSTANTS.SAAS_OFFERINGS", a);
-        var c = {
+        var o = {
             pod_presets: !1
         };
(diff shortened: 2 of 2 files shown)


src/services/catalog.service.ts
```diff
@@ -117,6 +117,15 @@ export class CatalogService {
   }
 
   public getProjectCatalogItems(projectName: string, includeImages: boolean = true, includeTemplates: boolean = true, partialObjectMetadataList: boolean = false ) {
+    // Never fetch items from the `openshift` namespace as project items since
```
benjaminapetersen: Is this the right place for this logic?  If its `getProjectCatalogItems(projectName)`, I expect it to return the catalog items for whatever project, even if `openshift`.  Should calling code instead be wrapped in the `if` statement?

I don't see this method used elsewhere in `/src`, but in `origin-web-catalog` I find:

```
./scripts/directives/processTemplateDialog.js:      Catalog.getProjectCatalogItems(ctrl.templateProjectName, false, true).then( _.spread(function(catalogServiceItems, errorMessage) {
./scripts/directives/nav.js:                var projectItemsPromise = Catalog.getProjectCatalogItems(currentProjectName).then(_.spread(function(catalogServiceItems) {
./scripts/controllers/projectBrowseCatalog.js:          var projectItemsPromise = Catalog.getProjectCatalogItems(currentProjectName).then( _.spread(function(catalogServiceItems, errorMessage) {
```

I'm assuming this guard would be in place then for someone who is in the `openshift` namespace & happens to browse the catalog?

spadgett: Yeah, I debated this. The problem is we're adding namespaced service classes, and we *should* fetch those for the openshift namespace (just not templates and image streams). So this isn't always a no-op. The only place we can have that logic is here the way the code is structured.

My first implementation had a bug as well where I fixed it in one place and not the other. And I found myself copying code between the two (it was a bit more than just an `if` since we use `$q.all`).

`processTemplateDialog` probably shouldn't be using this method :/

> I'm assuming this guard would be in place then for someone who is in the openshift namespace & happens to browse the catalog?

Yes. Today you get a runtime error when you try to do that since there are multiple items with the same UID.

benjaminapetersen: Gotcha.  Thoughts on something like:

```JavaScript 
// a little more verbose, but only if overrdiging defaults. 
getProjectCatalogItems('my-project', {
  templates: true, 
  images: true,
  partialObjectMetadata: true, 
  openshiftNamespace: true
});
getProjectCatalogItems('my-project', {templates: false});
// vs, having to go look up what the booleans mean when using:
getProjectCatalogItems('my-project', false, false, true);
getProjectCatalogItems('my-project', false, true, false);
getProjectCatalogItems('my-project', false, false);
```
Maybe adding openshift namespace to that as a config object option, though default to `false` as well.  

I don't have super strong feelings, but it could be nice especially considering the need to look up the positional args across our repos.

Issue #437, opened 2017-09-18
Allow direct linking to the catalog with a specific category visible/open


david-martin:
Related to #423.

This feature will allow a user to arrive on the service catalog screen with a specific category/sub-category selected.
For example, if a user goes to https://mycluster.example.com/?category=databases&subCategory=mysql, the catalog will render as below

![image](https://user-images.githubusercontent.com/878251/30544887-dc4c2890-9c7f-11e7-8012-0288de5b5813.png)

david-martin:
@spadgett Similar to #423, does this seem like a reasonable feature request?

I can put a PR together if it's likely to be accepted

spadgett:
It makes sense to me. We might want to update those query parameters when the selected category changes like we do for other tabs in the web console.

I have a slight preference for `subcategory` without the camel casing.

david-martin:
:+1:  will do

david-martin:
@spadgett Would it make sense to add listeners for 'select-category' and 'select-subcategory' in the catalog https://github.com/openshift/origin-web-catalog/blob/f11248b99ae5d879190e02665d3f50e4a03f21dc/src/components/services-view/services-view.controller.ts#L51-L66,
 and emit those events from the console when data is loaded and if there is a 'category' and 'subcategory' query param?
https://github.com/openshift/origin-web-console/blob/6f40623444af501083edb4979a9b833ccfc5f026/app/scripts/controllers/landingPage.js#L113

Or should that logic be isolated to the catalog only?

david-martin:
@spadgett Any thoughts or points on above comment?

PR #669, opened 2018-04-04
Add search params

sedroche:
Following on from [issue 437](https://github.com/openshift/origin-web-catalog/issues/437) we want to add the ability to link to a filtered view of the catalog.

- I have added query params of `category` and `subcategory` to allow this filtering.
- Query parameters are updated when the selected category changes

### DIFF
--- a/src/components/services-view/services-view.controller.ts
+++ b/src/components/services-view/services-view.controller.ts
@@ -3,7 +3,7 @@ import * as _ from 'lodash';
 import * as $ from 'jquery';
 
 export class ServicesViewController implements angular.IController {
-  static $inject = ['Constants', 'Catalog', 'KeywordService', 'Logger', 'HTMLService', '$element', '$filter', '$rootScope', '$scope', '$timeout'];
+  static $inject = ['Constants', 'Catalog', 'KeywordService', 'Logger', 'HTMLService', '$element', '$filter', '$location', '$rootScope', '$scope', '$timeout'];
 
   static readonly MAX_RESIZE_RETRIES: number = 20;
 
@@ -16,6 +16,7 @@ export class ServicesViewController implements angular.IController {
   private element: any;
   private $filter: any;
   private $rootScope: any;
+  private $location: any;
   private $scope: any;
   private $timeout: any;
   private scrollParent: any;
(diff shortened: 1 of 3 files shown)


dtaylor113:
Seems this PR broke some existing unit tests. 

Tested landing page URLs as well as project catalog urls, ex:  https://localhost:9000/dev-console/project/my-proj/catalog?category=languages&subcategory=java, in combination with filtering, searching and changing categories.  LGTM.

src/components/services-view/services-view.controller.ts
```diff
@@ -55,7 +57,8 @@ export class ServicesViewController implements angular.IController {
     this.debounceResize = _.debounce(() => this.resizeExpansion(false), 50, { maxWait: 250 });
     $(window).on('resize.services', this.debounceResize);
 
-    this.ctrl.currentFilter = this.ctrl.currentSubFilter = 'all';
+    this.ctrl.currentFilter = this.$location.$$search.category || 'all';
+    this.ctrl.currentSubFilter = this.$location.$$search.category && this.$location.$$search.subcategory;
```
spadgett: We should use `$location.search()` to get the current values instead of accessing `$$search` directly.

$location.search()

spadgett (changes requested):
What happens if the category in the query param doesn't exist?

sedroche:
@spadgett If a category or subcategory doesn't exist then no results are shown.

## no category
no-category
## no subcategory
no-subcategory

src/components/services-view/services-view.controller.ts
sedroche: Gotcha thanks :+1:

sedroche:
@dtaylor113  Thanks for the review. I will fix the tests once the filtering is agreed upon

maleck13:
@spadgett Wondering is this ok to tag for 3.11

src/components/services-view/services-view.controller.ts
```diff
@@ -203,6 +212,8 @@ export class ServicesViewController implements angular.IController {
 
     this.ctrl.currentFilter = category;
     this.ctrl.currentSubFilter = subCategory;
+    subCategory = category === 'all' ? null : subCategory;
+    this.$location.path(this.$location.path()).search({category: this.ctrl.currentFilter, subcategory: subCategory});
```
spadgett: We should make adding the query param doesn't cause the view to reload. I think we'll need to add `reloadOnSearch: false` here:

https://github.com/openshift/origin-web-console/blob/master/app/scripts/app.js#L67

spadgett (approved):
/lgtm
/hold

Thanks. Holding for 3.11

src/components/services-view/services-view.controller.ts
sedroche: PR for this is here -> https://github.com/openshift/origin-web-console/pull/3011

PR #675, opened 2018-04-23
Clone conditional values on form fields

PhilipGough:
This change allows hiding form fields based on specific conditions as per https://github.com/json-schema-form/angular-schema-form/blob/0.8.13/docs/index.md#condition

### DIFF
--- a/src/components/catalog-parameters/catalog-parameters.controller.ts
+++ b/src/components/catalog-parameters/catalog-parameters.controller.ts
@@ -164,6 +164,10 @@ export class CatalogParametersController implements angular.IController {
         newObject.key = original.key;
       }
 
+      if (original.condition) {
+        newObject.condition = original.condition;
+      }
+
       if (CatalogParametersController.ALLOWED_FORM_INPUT_TYPES[original.type]) {
         newObject.type = original.type;
       }


spadgett:
@PhilipGough We're migrating to React in the near future, so I'm hesitant to rely on more specific features of angular-schema-form. Particularly since it relies specifically on angular expressions.

> The condition option lets you hide or show a field depending on an angular expression.

I think we really push more on parameter presentation at the OSB API level.

cc @pmorie

PhilipGough:
Hey @spadgett, I understand your concern. I was tasked with finding the minimum required changes to get this working as is.

One observation is that the react library that I have looked (in linked pr) offers the same conditional features. So while the implementation might change, if when moving to react you guys still plan on using a json schema forms lib, then I don't think we would introduce anything here that could not be implemented later on.

cc @maleck13

spadgett:
@PhilipGough Thanks, which library were you looking at? I'm not sure where we link is. We had done some early investigation into react-jsonschema-form if that's the one.

I'm OK adding this if we do it in a way that we can support after we migrate to React. We should be careful not to do anything too specific to angular-schema-form, and try to align whatever metadata we're looking for to align with the react-jsonschema-form implementation.

I also want to be careful that we're not letting brokers execute arbitrary JS as that's a security exposure. The conditional needs to be declarative. Angular expressions can run code, which is a problem due to XSS concerns.

I do think we need to push this at the OSB API level since we keep adding these one-off fixes and not address the real problem: no good way to specify presentation. I want to be careful how many band aids we keep adding.

spadgett:
It looks like the way to do it in react-jsonschema-form is using schema dependencies? This seems like the best approach since it's part of JSON schema itself.

https://github.com/mozilla-services/react-jsonschema-form#schema-dependencies

https://spacetelescope.github.io/understanding-json-schema/reference/object.html#dependencies

Regardless, we can't use angular-schema-form conditions because they let brokers run arbitrary script code.

PhilipGough:
> Regardless, we can't use angular-schema-form conditions because they let brokers run arbitrary script code.

I had just verified this myself actually prior to your comment so for security reasons that's obviously a non-runner. 

In regards to the dependencies, you are correct that is what is used in the react lib however in `angular-schema-form` I don't see support for this. If that is the case then I'm happy to close this and we can look for alternatives.

david-martin:
@spadgett What are you thoughts on us adding support for (a subset of) schema dependecies features for now, until the move to react.
i.e.
* identify which schema dependencies we need (I think a boolean check for a checkbox, and a string comparison for a select)
* validate the schema returned from the broker is one of these
* construct the angular-schema-form `condition` string dynamically (rather than just incjecting what the broker gave us)

spadgett:
@david-martin That seems reasonable to me.

It looks like the JSON spec only lets you add a dependency based on the presence of a value. Are you thinking then of implementing dynamic dependencies from react-jsonschema-form as described here?

https://github.com/mozilla-services/react-jsonschema-form#dynamic

david-martin:
@spadgett Yes, I was thinking about that, but I'm not sure where they would go (in the schema or in the openshift_form_defintion)

The 'oneOf' feature in dynamic dependencies would give what we need for both our current use cases.
Instead of a checkbox (true or false), we'd change this to be a select with 2 specific options, and show specific fields depending on which option is selected.

As dynamic dependencies is not part of JSON schema, I feel like it shouldn't go in the schema of the plan, but in the openshift_form_defitniion instead.
Is this OK?

We'd then add catalog UI code to set the appropriate `condition` string for angular, and in future copy the dependencies into the schema (again in the catalog UI code) for react.
How does that sound?

PhilipGough:
*Re-opened to make discussion more visible*

@spadgett @david-martin 

Hey guys, I've done some more investigation on this. The keywords to allow the dynamic provisioning in the react library we are discussing, specifically the `anyOf` keyword,  are in fact part of the schema. [See reference here](https://spacetelescope.github.io/understanding-json-schema/reference/combining.html#combining-schemas).

However there have been numerous failed attempts at implementing these keywords in a way that was acceptable in the library. [See docs FAQ ](https://github.com/mozilla-services/react-jsonschema-form#q-does-rjsf-support-oneof-anyof-multiple-types-in-an-array-etc).

So it seems like everything we would need is in the schema but hasn't made it to the lib. 

@maleck13 @david-martin Maybe we could spend some time trying to flesh out an approach based on our current use cases to get the support we will need going forward from the point of view of the broker? So my suggestion is as follows:

*Gather a set of requirements and use cases from others involved here and see what the community needs now as a minimum
*Investigate how far away the existing code in the react library is from allowing us to provide this
*Assuming the lib is workable, or we feel that we can make the required changes we could allow the user to define the dependency schema within the apb. I feel this is reasonable since it is part of the spec but if its seen as too complex from a dev perspective we could attempt to simplify it and provide some mapping.
*Intermediately we would need to provide a mechanism to translate a dependency schema to an angular condition and this work would need to go directly into the web-catalog for security reasons

What do others think of this potential approach?

PR #677, opened 2018-04-30
Add limited 'oneOf' json schema dependencies support for plan schemas.

david-martin:
This change examines the json schema object in the service plan,
looking for a `dependencies` object.
If this exists, any `oneOf` property dependencies defined will be
examined and a corresponding angular-schema-form `condition` will
be added to the form definition so specified fields will only be
shown when the corresponding option is selected/entered in another
field.

Short clip of how it looks
https://youtu.be/MX-G0TjaP7E

Example plan
```json
{
  "metadata": {
    "name": "5f04a3f508d7615a422863814f58ad98"
  },
  "spec": {
    "clusterServiceBrokerName": "ansible-service-broker", // some fields omitted
    "externalMetadata": {
      "schemas": {
        "service_binding": {},
        "service_instance": {
          "create": {
            "openshift_form_definition": [
              "eg_enum_param",
              "eg_conditional_one",
              "eg_conditional_two",
              "eg_boolean_param",
              "eg_conditional_boolean_one",
              "eg_conditional_boolean_two"
            ]
          },
          "update": {}
        }
      }
    },
    "instanceCreateParameterSchema": {
      "$schema": "http://json-schema.org/draft-04/schema",
      "additionalProperties": false,
      "dependencies": {
        "eg_boolean_param": {
          "properties": {
            "eg_conditional_boolean_one": {
              "title": "Example Shown When True",
              "type": "string"
            },
            "eg_conditional_boolean_two": {
              "title": "Example Also Shown When True",
              "type": "string"
            }
          },
          "required": [
            "eg_conditional_boolean_one",
            "eg_conditional_boolean_two"
          ]
        },
        "eg_enum_param": {
          "oneOf": [
            {
              "properties": {
                "eg_conditional_one": {
                  "title": "Example Shown If 'Yes' or 'Maybe'",
                  "type": "string"
                },
                "eg_enum_param": {
                  "enum": [
                    "Yes",
                    "Maybe"
                  ]
                }
              },
              "required": [
                "eg_conditional_one"
              ]
            },
            {
              "properties": {
                "eg_conditional_two": {
                  "title": "Example Shown if 'No'",
                  "type": "string"
                },
                "eg_enum_param": {
                  "enum": [
                    "No"
                  ]
                }
              },
              "required": [
                "eg_conditional_two"
              ]
            }
          ]
        }
      },
      "properties": {
        "eg_boolean_param": {
          "title": "Conditional Example Boolean",
          "type": "boolean"
        },
        "eg_enum_param": {
          "default": "Yes",
          "enum": [
            "Yes",
            "No",
            "Maybe"
          ],
          "title": "Conditional Example",
          "type": "string"
        }
      },
      "type": "object"
    } // some fields omitted
  }
}
```

Example parameters in an apb.yml (pending related broker changes in https://github.com/openshift/ansible-service-broker/pull/928)
```yaml
plans:
  - name: default
    description: __DEPENDENCIES
    free: True
    metadata: {}
    parameters:
    - name: eg_enum_param
      type: enum
      title: Conditional Example
      default: "Yes"
      enum: ["Yes", "No", "Maybe"]
    - name: eg_conditional_one
      title: Example Shown If 'Yes' or 'Maybe'
      type: string
      dependencies:
      - key: eg_enum_param
        value: ["Yes", "Maybe"]
    - name: eg_conditional_two
      title: Example Shown if 'No'
      type: string
      dependencies:
      - key: eg_enum_param
        value: "No"
    - name: eg_boolean_param
      title: Conditional Example Boolean
      type: boolean
    - name: eg_conditional_boolean_one
      title: Example Shown When True
      type: string
      dependencies:
      - key: eg_boolean_param
    - name: eg_conditional_boolean_two
      title: Example Also Shown When True
      type: string
      dependencies:
      - key: eg_boolean_param
```

Changes based on initial integration with broker (with @philipgough)
* [x] Allow for multiple condition matches i.e. the property enum array has more than 1 string
* [x] Auto add fields to the properties if they aren't already there i.e. it should be OK to define a property in a dependency only, rather than having to define it in the `dependencies` and in the `properties`

### DIFF
--- a/src/components/catalog-parameters/catalog-parameters.controller.ts
+++ b/src/components/catalog-parameters/catalog-parameters.controller.ts
@@ -12,11 +12,18 @@ export class CatalogParametersController implements angular.IController {
     'select': true
   };
 
+  private static CONDITIONS = {
+    oneOf: " || ",
+    fieldEqual: _.template('model["${key}"] === ${test}'),
+    fieldTruthy: _.template('model["${key}"]')
+  };
+
   public ctrl: any = this;
 
   public $onInit() {
     this.setupFormDefaults();
     this.ctrl.parameterForm = this.cloneParameterForm(this.ctrl.parameterFormDefinition) || ['*'];
+    this.setupFormDependencies();
     this.updateHiddenModel();
     this.setupReadonlySchema();
   }
@@ -72,6 +79,77 @@ export class CatalogParametersController implements angular.IController {
         checkboxHelpHtmlClass: 'col-sm-8 col-sm-offset-4'
       });
     }
+  }
+
+  private setupFormDependencies() {
+    _.each(_.get(this.ctrl.parameterSchema, 'dependencies'), (dependency: any, key: string) => {
+      if (_.isArray(dependency.oneOf)) {
+        // Only support 'oneOf' at this time, not 'allOf' or 'anyOf'
+        // See https://github.com/mozilla-services/react-jsonschema-form#dynamic for example 'dependencies' object
+        _.each(dependency.oneOf, (schema) => {
+          this.mapDependencyCondition(schema, key);
+        });
+      } else {
+        // simple nested dependency
+        this.mapDependencyCondition(dependency, key);
+      }
+    });
+  }
+
+  private mapDependencyCondition(schema, key) {
+    // Get the fields to show if the dependency is valid
+    const requiredFields = _.get(schema, 'required');
+    if (!_.isArray(requiredFields)) {
+      return;
+    }
+    _.each(requiredFields, (requiredField: any) => {
+      let fieldIndex = _.findIndex(this.ctrl.parameterForm, (field: any) => {
+        if (_.isObject(field)) {
+          return requiredField === field.key;
+        } else {
+          return requiredField === field;
+        }
+      });
+
+      // Only continue if the field is in the form defintion
+      if (fieldIndex > -1) {
+        // Normalise the field object
+        let field = this.ctrl.parameterForm[fieldIndex];
+        if (_.isString(field)) {
+          field = {
+            key: field
+          };
+        }
+
+        // Check if the required field is already in the schema properties.
+        // Add it if not.
+        if (!_.has(this.ctrl.parameterSchema, ['properties', field.key])) {
+          this.ctrl.parameterSchema.properties[field.key] = schema.properties[field.key];
+        }
+
+        // Set the angular 'condition'
+        const enumMatches = _.get(schema, ['properties', key, 'enum']);
+        if (!enumMatches) {
+          // Nothing specific to match against, just add a truthy condition
+          field.condition = CatalogParametersController.CONDITIONS.fieldTruthy({
+            key: key
+          });
+        } else if (enumMatches.length > 0) {
+          field.condition = CatalogParametersController.CONDITIONS.fieldEqual({
+            key: key,
+            test: JSON.stringify(enumMatches[0])
+          });
+          // Allow matching with any other strings in the array (if there are multiple)
+          for (let oi=1, ol=enumMatches.length; oi<ol; oi++) {
+            field.condition += CatalogParametersController.CONDITIONS.oneOf + CatalogParametersController.CONDITIONS.fieldEqual({
+              key: key,
+              test: JSON.stringify(enumMatches[oi])
+            });
+          }
+        }
+        this.ctrl.parameterForm.splice(fieldIndex, 1, field);
+      }
+    });
   }
 
   private setupReadonlySchema() {
--- a/dist/origin-web-catalogs.js
+++ b/dist/origin-web-catalogs.js
@@ -394,11 +394,11 @@ webpackJsonp([ 0, 1 ], [ function(e, t) {
             } ]
         } ];
         n.set(window, "OPENSHIFT_CONSTANTS.SERVICE_CATALOG_CATEGORIES", s), n.set(window, "OPENSHIFT_CONSTANTS.SAAS_OFFERINGS", a);
-        var c = {
+        var o = {
             pod_presets: !1
         };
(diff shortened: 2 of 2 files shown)


david-martin:
ping @spadgett @PhilipGough @pb82.

Related to UI & Broker changes discussed in https://github.com/openshift/ansible-service-broker/issues/859 & https://github.com/openshift/origin-web-catalog/pull/675

maleck13:
In the example I see a field always shows. Is it possible to not show any fields also? excuse my ignorance just not sure

david-martin:
@maleck13 do you mean "don't show this field if another field has a specific value" ?

If yes, I don't think that's part of json schema dependencies. i.e. you define fields that are dependencies of others, but *not* fields that are negated dependencies.
You can workaround this though quite easily e.g

* field 1 gives 3 options.
* only show field 2 if option 1 is chosen
* only show field 3 if option 2 is chosen
* only show field 4 if option 3 is chosen

In this case, field 3 & 4 won't be shown if option 1 is selected on field 1.

david-martin:
@philipgough I've added support for below if you want to try it again:

* auto add properties defined in dependencies to the schema properties if they are not already there
* allow matching from multiple strings i.e. the `oneOf` bit

david-martin:
After discussion on https://github.com/openshift/ansible-service-broker/issues/859#issuecomment-387359151 I'll look to round out this PR by allowing `oneOf` conditions with booleans and numbers (not just string).

david-martin:
@spadgett This is ready for review now.
Where would be an appropriate place for documenting this catalog feature?

src/components/catalog-parameters/catalog-parameters.controller.ts
```diff
@@ -74,6 +81,84 @@ export class CatalogParametersController implements angular.IController {
...
+
+        // Check if the required field is already in the schema properties.
+        // Add it if not.
+        if (_.isUndefined(this.ctrl.parameterSchema.properties[field.key])) {
+          this.ctrl.parameterSchema.properties[field.key] = schema.properties[field.key];
+        }
+
+        // Set the angular 'condition'
+        let enumMatches = _.get(schema, 'properties.' + key + '.enum');
+        if (_.isUndefined(enumMatches)) {
```
spadgett: Should this just be...?

```js
if (enumMatches) {
```

src/components/catalog-parameters/catalog-parameters.controller.ts
```diff
@@ -74,6 +81,84 @@ export class CatalogParametersController implements angular.IController {
...
+          }
+        }
+        this.ctrl.parameterForm.splice(fieldIndex, 1, field);
+      }
+    });
+  }
+
+  private escapeIfString(value) {
+    if (_.isString(value)) {
+      return '"' + value + '"';
```
spadgett: OK, so if I understand correctly this is escaping the string to use in an Angular expression. If the string contains an `"` character, could I break out an execute arbitrary expressions?

Should we just use `JSON.stringify` on all values?

What types can `value` be?

src/components/catalog-parameters/catalog-parameters.controller.ts
```diff
@@ -74,6 +81,84 @@ export class CatalogParametersController implements angular.IController {
...
+          field.condition = CatalogParametersController.CONDITIONS.fieldTruthy({
+            key: key
+          });
+        } else if (enumMatches.length > 0) {
+          field.condition = CatalogParametersController.CONDITIONS.fieldEqual({
+            key: key,
+            test: this.escapeIfString(enumMatches[0])
+          });
+          // Allow matching with any other strings in the array (if there are multiple)
+          for (var oi=1, ol=enumMatches.length; oi<ol; oi++) {
```
spadgett: let

src/components/catalog-parameters/catalog-parameters.controller.ts
david-martin: > OK, so if I understand correctly this is escaping the string to use in an Angular expression. If the string contains an " character, could I break out an execute arbitrary expressions?

I have tested a value with quotes in it (in the serviceplan definition), and they are escaped before they get to this point.
However, ...

> Should we just use JSON.stringify on all values?

... this makes a lot of sense and will definitely prevent arbitrary expressions.
I'll change it to use this.

> 
> What types can value be?

String, number, boolean.

src/components/catalog-parameters/catalog-parameters.controller.ts
```diff
@@ -74,6 +81,84 @@ export class CatalogParametersController implements angular.IController {
...
+  private mapDependencyCondition(schema, key) {
+    // Get the fields to show if the dependency is valid
+    let requiredFields = _.get(schema, 'required');
+    if (!_.isArray(requiredFields)) {
+      return;
+    }
+    _.each(requiredFields, (requiredField: any) => {
+      let fieldIndex = _.findIndex(this.ctrl.parameterForm, (field: any) => {
+        if (_.isObject(field)) {
+          return requiredFields.indexOf(field.key) > -1;
```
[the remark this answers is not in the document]
david-martin: This code is actually wrong.
It should be:

```
        if (_.isObject(field)) {
          return requiredField === field.key;
        } else {
          return requiredField === field;
        }
```

src/components/catalog-parameters/catalog-parameters.controller.ts
david-martin: This is an undefined check, so do you mean `if (!enumMatches) {` ?

david-martin:
@spadgett Thanks for the feedback. My typescript skills are obviously lacking :)

I've pushed up changes for re-review

spadgett (approved):
LGTM, but hold for 3.11

/hold

maleck13:
@spadgett ok to merge this now as it is generically useful

david-martin:
Out of date, no longer required

Issue #673, opened 2018-04-18
The modal wizard utilized in the Service Catalog does not support keyboard shortcuts


serenamarie125:
The modal wizards utilized when selecting an item in the Service Catalog should support keyboard shortcuts.  Currently when the modal wizard is issued, nothing in the wizard has focus by default.

For each page in the wizard, some element should have focus.  The @openshift/team-ux-review should provide input for this.

Once you get to the configuration step, if you select an item, you can then traverse through the form, continue to Tab to the Next button & hit the space key to proceed to the next page.  But once you get to the next page, nothing has default focus.

serenamarie125:
When on the Information page, the Next page should have focus.
When on the Configuration page, the first field should have focus.

Notes:
Move to the next field on a page. | Tab
Return to the previous field on a page. | Shift+Tab
Perform a button action | Space
Move to the next selection in a list. | Down arrow
Move to the previous selection in a list. | Up arrow

serenamarie125:
@spadgett FYI

serenamarie125:
@jeff-phillips-18 can you take a look at this?

jeff-phillips-18:
will do

jeff-phillips-18:
@serenamarie125 For the binding step, since there is a default (do not bind), should focus should be on the `Create` button?

serenamarie125:
That's a good question @jeff-phillips-18.  As a general rule, I would suggest the following:
- When no input fields exist, default focus should be on the primary action button of the wizard/modal
- When one or more input fields exist, regardless if there are default values, default focus should be on the primary action button

When tabbing through input fields, the order should be depending on the field placement.  (top to bottom, left to right )

Once the last input field has focus, tabbing will bring you to the primary action of the wizard button bar (Next in this case), then to the Back button, and then the Cancel button.  Tabbing once more will loop back to the first input field on the modal wizard.  (Note that the behavior would be the same for any modal, the action buttons would just differ).

<img width="857" alt="screen shot 2018-04-20 at 8 35 17 pm" src="https://user-images.githubusercontent.com/12504403/39078645-70441e88-44da-11e8-8cf7-cdb7d07c741c.png">

Note that this also means if the focus is on the first input field and all fields have default values, <Shift>Tab will bring them to the Cancel button, then the Back button, and then the Next button, which will be more efficient than tabbing through a form with a lot of input fields.

Also note that if an input field or button  is disabled, it will be skipped and not receive focus.

[...]

PR #678, opened 2018-05-08
Keep focus inside overlay panel when displayed

jeff-phillips-18:
Set initial focus appropriately in wizards
Requires upgrade to Angular-Patternfly 4.18

This fixes most of #673. Will need follow up work on some of the initial focus components and possibly adjust the tab order for the wizard buttons (might need to be done in angular-patternfly) if we want the tab order to be the next button then the cancel button rather than left to right ordering.

This PR gets us most of the way there and allows for keyboard access.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "angular-animate": "1.5.11",
     "angular-drag-and-drop-lists": "2.0.0",
     "angular-moment": "1.0.0",
-    "angular-patternfly": ">=4.11.8 <5.0.0",
+    "angular-patternfly": ">=4.18.0 <5.0.0",
     "angular-sanitize": "1.5.11",
     "angular-schema-form": "^0.8.13",
     "angular-schema-form-bootstrap": "^0.2.0",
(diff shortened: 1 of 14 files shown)


package.json
```diff
@@ -30,7 +30,7 @@
     "angular-animate": "1.5.11",
     "angular-drag-and-drop-lists": "2.0.0",
     "angular-moment": "1.0.0",
-    "angular-patternfly": ">=4.11.8 <5.0.0",
+    "angular-patternfly": ">=4.18.0 <5.0.0",
```
dtaylor113: Looks like Console is at "angular-patternfly": "4.14.9":  https://github.com/openshift/origin-web-console/blob/master/bower.json#L16

1) I think Catalog bower.json needs to be updated  to "angular-patternfly": ">=4.18.0 <5.0.0",
2) Assuming a Console PR next to update Console bower.json to:  "angular-patternfly": "4.18.0" from  currently "angular-patternfly": "4.14.9"

jeff-phillips-18: Added bower.json.
Yes, when console is upgraded to a version including this change it will also need to upgrade to angular-patternfly 4.18.0 or higher.

PR #680, opened 2018-05-15
Update components/update-service to update secret parameters in place

benjaminapetersen:
Fixes [bugzilla 1564979](https://bugzilla.redhat.com/show_bug.cgi?id=1564979)
- Updates `components/update-service` to update a secret in place rather than creating a new secret.

Attempted to break this massive function down into something a bit easier to maintain. Typescript is interesting, not sure I'm sold yet (but working on it), struggled with a type `string[][]` that didn't work (see `note`).  Made some other minor changes like using `let` instead of `var`, etc. 

Attempted to update types, but `object` only works on a very simple object, TypeScript seems to think a multidimensional object should have a custom type def.  That would be a good bit of work to update throughout the codebase I think.

Compiler complains about this a lot:

screen shot 2018-05-15 at 11 40 02 am

For some reason I prefer the first:

```JavaScript 
// :)
let foo:object = { };
// :/
let foo: object = {};
```
But I can update this && add the space to make it happy.

### DIFF
--- a/src/components/update-service/update-service.controller.ts
+++ b/src/components/update-service/update-service.controller.ts
@@ -213,74 +213,28 @@ export class UpdateServiceController implements angular.IController {
     this.ctrl.orderComplete = false;
     this.ctrl.error = null;
 
-    var parameterData: any = this.getParameters(this.ctrl.parameterData);
+    let parameterData: any = this.getParameters(this.ctrl.parameterData);
 
     // Update the instance parameters
-    var originalParameters: any = _.get(this.ctrl.serviceInstance, 'spec.parameters');
-    var originalKeys: any = _.map(originalParameters, (value: any, key: string) => {
-      // Make sure properties with a '.' in the name aren't treated as paths.
+    let originalParameters: object = _.get(this.ctrl.serviceInstance, 'spec.parameters');
+    // NOTE: should be string[][], but compiler doesn't handle this
+    let originalKeys: any[] = _.map(originalParameters, (value: any, key: string) => {
       return [key];
     });
-    var updateParameters: any = _.pick(parameterData, originalKeys);
-    var newParameters: any = _.omit(parameterData, originalKeys);
-    var updateInstance: any = angular.copy(this.ctrl.serviceInstance);
+    let updateParameters: object = _.pick(parameterData, originalKeys);
+    let newParameters: object = _.omit(parameterData, originalKeys);
+    let updateInstance: object = angular.copy(this.ctrl.serviceInstance);
 
-    if (_.get(updateInstance, 'spec.clusterServicePlanExternalName') !== _.get(this.ctrl.selectedPlan, 'spec.externalName')) {
-      _.unset(updateInstance, 'spec.clusterServicePlanRef');
-      _.set(updateInstance, 'spec.clusterServicePlanExternalName', _.get(this.ctrl.selectedPlan, 'spec.externalName'));
-    }
+    this.updateServicePlanName(updateInstance);
+    this.updateSpecParameters(updateInstance, updateParameters, originalParameters);
 
-    if (!angular.equals(updateParameters, originalParameters)) {
-      _.set(updateInstance, 'spec.parameters', updateParameters);
-    }
+    let updatedSecretParameters: object = this.updateSecretParameters(updateInstance, newParameters);
 
-    // Update the secret parameters
-    var updatedSecretParameters: any = {};
-
-    _.each(this.secrets, (secret: any) => {
-      var originalSecretParameters: any = JSON.parse(this.SecretsService.decodeSecretData(secret.data).parameters);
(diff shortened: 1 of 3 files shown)


benjaminapetersen:
Cleaned up the typescript lint warnings.

src/components/update-service/update-service.controller.ts
```diff
@@ -213,74 +213,28 @@ export class UpdateServiceController implements angular.IController {
     this.ctrl.orderComplete = false;
     this.ctrl.error = null;
 
-    var parameterData: any = this.getParameters(this.ctrl.parameterData);
+    let parameterData: any = this.getParameters(this.ctrl.parameterData);
```
spadgett: If we're going to switch, we should use `const` here and below :)

src/components/update-service/update-service.controller.ts
```diff
@@ -213,74 +213,28 @@ export class UpdateServiceController implements angular.IController {
...
 
     // Update the instance parameters
-    var originalParameters: any = _.get(this.ctrl.serviceInstance, 'spec.parameters');
-    var originalKeys: any = _.map(originalParameters, (value: any, key: string) => {
-      // Make sure properties with a '.' in the name aren't treated as paths.
+    let originalParameters: object = _.get(this.ctrl.serviceInstance, 'spec.parameters');
+    // NOTE: should be string[][], but compiler doesn't handle this
+    let originalKeys: any[] = _.map(originalParameters, (value: any, key: string) => {
```
spadgett: Does this work?

```ts
const originalKeys: string[][] = _.map(originalParameters, (value: any, key: string): string[] => {
```

I'm not sure why we need a multidimensional array here, though.

src/components/update-service/update-service.controller.ts
```diff
@@ -213,74 +213,28 @@ export class UpdateServiceController implements angular.IController {
...
       return [key];
     });
-    var updateParameters: any = _.pick(parameterData, originalKeys);
-    var newParameters: any = _.omit(parameterData, originalKeys);
-    var updateInstance: any = angular.copy(this.ctrl.serviceInstance);
+    let updateParameters: object = _.pick(parameterData, originalKeys);
+    let newParameters: object = _.omit(parameterData, originalKeys);
+    let updateInstance: object = angular.copy(this.ctrl.serviceInstance);
```
spadgett: Not sure object vs any buys us much... :/

I'd probably just use

```ts
const updateParameters = _.pick(parameterData, originalKeys);
const newParameters = _.omit(parameterData, originalKeys);
const updateInstance = angular.copy(this.ctrl.serviceInstance);
```

src/components/update-service/update-service.controller.ts
```diff
@@ -299,6 +253,37 @@ export class UpdateServiceController implements angular.IController {
...
+  };
+
+  private updateSpecParameters(instance: object, updateParameters: object, originalParameters: object) {
+    if (!angular.equals(updateParameters, originalParameters)) {
+      _.set(instance, 'spec.parameters', updateParameters);
+    }
+  };
+
+  private updateSecretParameters(instance: object, newParameters: object) {
+    let secretParams: any = {};
```
spadgett: A lot of these `any` declarations are not needed. It will infer the type based on assignment.

tslint.json
```diff
@@ -65,7 +65,8 @@
     "typedef-whitespace": [true,
       ["callSignature", "noSpace"],
       ["catchClause", "noSpace"],
-      ["indexSignature", "space"]
+      ["indexSignature", "space"],
+      [ "variable-declaration", "nospace"]
```
spadgett: I'd rather just use `{}` since we use that everywhere else in the code.

spadgett (commented):
Thanks @benjaminapetersen. I think this change is risky enough to defer to 3.11 at this point since it's not a blocker bug.

src/components/update-service/update-service.controller.ts
benjaminapetersen: Not much, but is aiming in the right direction I think :/

tslint.json
benjaminapetersen: I can just remove the rule & leave it as it was.

benjaminapetersen: (agree, code already looks a certain way, so roll with it.)

src/components/update-service/update-service.controller.ts
benjaminapetersen: It's a copy-paste move, the code was originally a multidimensional array.  Not sure why, really.  I was aiming to type it anyway, though.

src/components/update-service/update-service.controller.ts
benjaminapetersen: agree, i can change this.

dtaylor113 (commented):
obsolete PR?

